PR #7, opened 2019-02-18
Add getrandom wrapper func and documentation

dhardy:
Add a wrapper function for docs and to ensure all platforms have the same API.

Add docs (based on `rand_os`).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,7 +42,8 @@
 //! The bare WASM target `wasm32-unknown-unknown` tries to call the javascript
 //! methods directly, using either `stdweb` or `wasm-bindgen` depending on what
 //! features are activated for this crate. Note that if both features are
-//! enabled `wasm-bindgen` will be used.
+//! enabled `wasm-bindgen` will be used. If neither feature is enabled,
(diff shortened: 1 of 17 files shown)


src/lib.rs
```diff
@@ -5,6 +5,91 @@
...
+//! ## Support for WebAssembly and ams.js
+//!
+//! The three Emscripten targets `asmjs-unknown-emscripten`,
+//! `wasm32-unknown-emscripten` and `wasm32-experimental-emscripten` use
+//! Emscripten's emulation of `/dev/random` on web browsers and Node.js.
+//!
+//! The bare WASM target `wasm32-unknown-unknown` tries to call the javascript
+//! methods directly, using either `stdweb` or `wasm-bindgen` depending on what
+//! features are activated for this crate. Note that if both features are
+//! enabled `wasm-bindgen` will be used.
```
newpavlov: It's probably worth to add that if none of the features is enabled, then dummy function which always returns "unavailable" error will be used.

dhardy:
We also have `syscall_getrandom` so perhaps `inner` makes more sense than `os` :+1:

PR #9, opened 2019-02-23
8 Update Solaris getrandom

jasonbking:
One note -- the original unmodified code fails to build on 1.32.0 or later (at least).  The for loops returned `()` instead of `Result<>` from `getrandom_inner()` which then conflicted with the `getrandom_inner()` return type.

With this version, it builds and all the test pass successfully, though I'm not sure if my solution is the preferred one or not.


dhardy:
We haven't set up CI on this repo yet. I also found non-compilable code recently.

PR #10, opened 2019-02-25
Add CI config and implement on WASM

dhardy:
Some failures are expected for now:

- WASM support needs implementing
- WASM test needs to be copied from Rand
- doc links need fixing

Some may fail:

- Android support was having problems on Rand
- Fuchsia tests were disabled following problems

See also https://github.com/rust-random/rand/issues/743

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 /target
 **/*.rs.bk
 Cargo.lock
+*.ts
+*.js
+*.wasm
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,162 @@
+language: rust
+sudo: false
+
+matrix:
+  include:
+    - rust: 1.28.0
+      env: DESCRIPTION="Linux, 1.28.0"
+      os: linux
+
+    - rust: 1.28.0
+      env: DESCRIPTION="OSX, 1.22.0"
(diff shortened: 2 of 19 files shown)


dhardy:
@newpavlov review please?

There are a few errors I didn't know what to do with.

newpavlov:
Hm, I am not that knowledgeable about CI setups and WASM, so not sure if I will be able to help. At a cursory look I couldn't find anything out of place. I think that after fixing the obvious failures it will be better to link this PR on users forum and reddit and ask for reviews.

dhardy:
Lets see how many of the failures the last patch fixes...

I'm not too concerned about reviewing the Travis config; I have a reasonable understanding of what it does even if I dislike poking it. I've also tested the WASM stuff locally, including with intentionally-failing tests.

But @newpavlov I would like your opinion on the code changes to the lib itself.

BTW I brought the `cloudabi` lib back as a dependency; I prefer to make use of external API crates where they exist (`libc`, `winabi`). The fact that all builds pull this in is a Cargo bug I hope they fix soon (apparently quite substantial changes are planned for this year).

newpavlov:
>But @newpavlov I would like your opinion on the code changes to the lib itself.

Looks good to me!

PR #11, opened 2019-03-01
Revise error types and codes

dhardy:
This makes a few changes:

-   revises error codes to have a common prefix
-   allows custom codes and messages per target
-   makes the `error` module public

It feels quite clunky having to add an empty `fn error_msg_inner` to all modules, but I don't see a better approach.

@newpavlov review? I think this concludes all my to-do list for the first release.

BTW the `dummy` module was broken; we don't appear to have a test for it (I hacked `lib.rs` locally).

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -12,20 +12,25 @@ use core::fmt;
 #[cfg(not(target_env = "sgx"))]
 use std::{io, error};
 
+// A randomly-chosen 16-bit prefix for our codes
+pub(crate) const CODE_PREFIX: u32 = 0x57f40000;
+const CODE_UNKNOWN: u32 = CODE_PREFIX | 0;
+const CODE_UNAVAILABLE: u32 = CODE_PREFIX | 1;
+
 /// An unknown error.
-pub const UNKNOWN_ERROR: Error = Error(unsafe {
-    NonZeroU32::new_unchecked(0x756e6b6e) // "unkn"
+pub const UNKNOWN: Error = Error(unsafe {
+    NonZeroU32::new_unchecked(CODE_UNKNOWN)
 });
 
 /// No generator is available.
-pub const UNAVAILABLE_ERROR: Error = Error(unsafe {
-    NonZeroU32::new_unchecked(0x4e416e61) // "NAna"
+pub const UNAVAILABLE: Error = Error(unsafe {
+    NonZeroU32::new_unchecked(CODE_UNAVAILABLE)
 });
 
 /// The error type.
 /// 
 /// This type is small and no-std compatible.
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[derive(Copy, Clone, Eq, PartialEq)]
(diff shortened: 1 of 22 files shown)


src/error.rs
```diff
@@ -12,20 +12,25 @@ use core::fmt;
 #[cfg(not(target_env = "sgx"))]
 use std::{io, error};
 
+// A randomly-chosen 16-bit prefix for our codes
+pub(crate) const CODE_PREFIX: u32 = 0x57f40000;
+const CODE_UNKNOWN: u32 = CODE_PREFIX | 0;
+const CODE_UNAVAILABLE: u32 = CODE_PREFIX | 1;
```
newpavlov: Maybe we should make `CODE_UNKNOWN` and `CODE_UNAVAILABLE` associated constants of `Error` type?

dhardy:
We also don't *need* the `std::error` module, but I don't think it gets in the way? At least in this case the other error types can have better names.

newpavlov:
I think a better analogy will be `std::io::error`, which as you know does not exist. Think in terms of potential inclusion into `std`.

dhardy:
And yet `std::io::prelude` exists, while there are several error-related things in `std::io`. But okay, I guess we can do away with the `error` module.

newpavlov:
BTW maybe we should add `UNKNOWN` and `UNAVAILABLE` error code values to docstrings (in both hex and dec forms), so it will be possible to search for them via search engines?

src/error.rs
```diff
@@ -38,14 +43,34 @@ impl Error {
...
+            }
+        }
+    }
+}
+
+impl fmt::Debug for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        match self.msg() {
+            Some(msg) => write!(f, "Error(\"{}\")", msg),
+            None => write!(f, "Error({})", self.0.get()),
```
newpavlov: Maybe it's better to use `"Error({:08X})"` formatting here? Same for `Display` implementation. Also `nonZeroU32` already implements `Display`, so no need for `get()` here.

newpavlov:
Sorry, I meant `"Error(0x{:08X})"` in the comment earlier. Also what do you think about associated constants?

dhardy:
Okay. I added the mentioned logging. What do you think about this? Many of the messages are not hugely useful, but this feature is optional and it's still not much overhead.

newpavlov:
No objections, though I am also not sure about usefulness.

dhardy:
Agreed. Still, I don't wish to spend the time on trivial tweaks now so this will do (once I fix tests).

dhardy:
FYI, libc version bumped [to solve this error](https://github.com/rust-lang/libc/issues/1303). I lack motivation to find the minimal version required.

newpavlov:
I can prepare a PR with some minor tweaks after you'll merge this one.

dhardy:
Please do.

PR #12, opened 2019-03-15
Minor updates

newpavlov:
I haven't added `RETRYABLE` constant and `is_retryable` method, as they can be added later in a backwards compatible way if need for them will arise.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -14,46 +14,36 @@ use std::{io, error};
 
 // A randomly-chosen 24-bit prefix for our codes
 pub(crate) const CODE_PREFIX: u32 = 0x57f4c500;
-const CODE_UNKNOWN: u32 = CODE_PREFIX | 0;
-const CODE_UNAVAILABLE: u32 = CODE_PREFIX | 1;
+const CODE_UNKNOWN: u32 = CODE_PREFIX | 0x00;
+const CODE_UNAVAILABLE: u32 = CODE_PREFIX | 0x01;
 
 /// The error type.
-/// 
+///
 /// This type is small and no-std compatible.
 #[derive(Copy, Clone, Eq, PartialEq)]
 pub struct Error(NonZeroU32);
 
 impl Error {
     /// An unknown error.
-    /// 
-    /// This is the following constant: 57f4c500 (hex) / 1475659008 (decimal).
-    ///
-    /// The constant value is provided for reference only. It may change in
-    /// future and it will not be considered a breaking change.
     pub const UNKNOWN: Error = Error(unsafe {
         NonZeroU32::new_unchecked(CODE_UNKNOWN)
     });
 
     /// No generator is available.
-    /// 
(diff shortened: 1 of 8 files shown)


src/error.rs
```diff
@@ -13,31 +13,37 @@ use core::fmt;
 use std::{io, error};
 
 // A randomly-chosen 16-bit prefix for our codes
-pub(crate) const CODE_PREFIX: u32 = 0x57f40000;
+pub(crate) const CODE_PREFIX: u32 = 0x57f4c500;
```
dhardy: Apparently this is now a 24-bit prefix.

I was thinking of allowing an 8-bit section per platform for platform-specific constants, hence leaving 16-bits zero, but I guess it's not too important.

src/error.rs
```diff
@@ -13,31 +13,37 @@ use core::fmt;
...
 #[derive(Copy, Clone, Eq, PartialEq)]
 pub struct Error(NonZeroU32);
 
 impl Error {
+    /// An unknown error.
+    /// 
+    /// This is the following constant: 57f4c500 (hex) / 1475659008 (decimal).
+    ///
+    /// The constant value is provided for reference only. It may change in
+    /// future and it will not be considered a breaking change.
```
dhardy: I think post-release we should try not to change this though?

src/lib.rs
```diff
@@ -67,12 +67,12 @@
...
 //! We always choose failure over returning insecure "random" bytes. In general,
-//! on supported platforms, failure is unlikely, though not impossible. If an
-//! error does occur, then it is likely that it will occur on every call to
+//! on supported platforms, failure is highly unlikely, though not impossible.
+//! If an error does occur, then it is likely that it will occur on every call to
 //! `getrandom`, hence after the first successful call one can be reasonably
 //! confident that no errors will occur.
 //! 
-//! On unsupported platforms, `getrandom` always fails.
+//! On unsupported platforms, `getrandom` always fails with [`Error::UNKNOWN`].
```
dhardy: wrong error

src/error.rs
newpavlov: Yes, but I think we should specify somehow, that code should not match on the given error code and use constant instead. I guess we can change wording somehow, but I don't have good ideas right now.

src/error.rs
newpavlov: Hm, platform specific codes are defined by platform and not by us, so I am not sure how it would work.

newpavlov:
>BTW I think that would be spelt retriable.

Looks like both variants can be used: https://english.stackexchange.com/questions/305273

src/error.rs
dhardy: Not existing platform-specific codes, but at least the wasm-stdweb impl seems to want its own codes.

newpavlov: Hm, how about simply listing those platform-specific error codes in documentation?

Issue #13, opened 2019-03-15
Removing per-thread file descriptors


newpavlov:
I think we can replace TLS for several targets by using a mutable static `RawFd` initialized via `std::sync::Once`. I _think_ it should be safe to read from `/dev/urandom` using a single file descriptor from several threads without any synchronization, but it's better to check it. A small [experiment](https://gist.github.com/newpavlov/ca9442fd5c1e75e90232db209f7174d8) shows that reading a huge amount of random data in one thread does block other threads.

dhardy:
Interesting idea. I guess [why not](https://stackoverflow.com/questions/5256599/what-are-file-descriptors-explained-in-simple-terms)?

I wouldn't expect the kernel to care about high-performance simultaneous reading of `/dev/urandom` from multiple threads anyway, so I don't think performance or blocking would be (much) different either way, but as you say it does avoid the need for TLS on these platforms.

`RawFd` is type `c_int` so it should even be possible to use an atomic int — otherwise, I believe `Cell` should be preferred over mutable statics.

newpavlov:
Yes, I thought about using atomic, but:
1) `c_int` _may_ be equal to 16 bits in some rare cases
2) Atomic i32 will be stable only from Rust 1.34
3) IIUC zero is a valid descriptor, usually it's used for stdin, but in some circumstances opening file could result in a zero fd.

>I believe Cell should be preferred over mutable statics.

Can you elaborate? Either way we need a static, so why not make it mutable directly without `Cell`? Mutable static initialization even used as an example for `Once::call_once`.

One thing I am not sure about is to how process error from `File::open("/dev/urandom")`. Due to the limitations of `Once` I was thinking about the following options, adding to enum stored in the static `Errored` state, i.e. if attempt to open `/dev/urandom` failed first time, `getrandom` will always return errors. We could've used `call_once_force`, but unfortunately it's currently unstable. The most common error case will be probably exhausted file descriptors, so it may happen in practice.

dhardy:
Firstly, `AtomicUsize` is already stable (though of course larger than necessary). The 0 state isn't a problem since you need to initialise with `Once` either way (to avoid opening multiple times); of course this means it's not necessary to use atomics.

[...]

PR #14, opened 2019-03-17
Remove TLS

newpavlov:
Closes: #13

### DIFF
--- a/src/linux_android.rs
+++ b/src/linux_android.rs
@@ -10,75 +10,102 @@
 extern crate std;
 
 use crate::Error;
-use crate::utils::use_init;
-use std::{thread_local, io::{self, Read}, fs::File};
-use core::cell::RefCell;
-use core::num::NonZeroU32;
-use core::sync::atomic::{AtomicBool, Ordering};
+use std::fs::File;
+use std::io::Read;
+use std::num::NonZeroU32;
+use std::sync::atomic::{AtomicIsize, AtomicUsize, Ordering};
+use std::os::unix::io::{AsRawFd, FromRawFd};
+use std::{io, mem, thread, time};
 
-static RNG_INIT: AtomicBool = AtomicBool::new(false);
+// replace with AtomicU8 on stabilization and MSRV bump
+static RNG_STATE: AtomicUsize = AtomicUsize::new(0);
+// replace with AtomicI32 on stabilization and MSRV bump
+static RNG_FD: AtomicIsize = AtomicIsize::new(-1);
 
-enum RngSource {
-    GetRandom,
-    Device(File),
+const STATE_INIT_ONGOING: usize = 1 << 0;
+const STATE_USE_SYSCALL: usize = 1 << 1;
+const STATE_USE_FD: usize = 1 << 2;
+
+pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
+    let state = RNG_STATE.load(Ordering::Acquire);
+    if state & STATE_USE_SYSCALL != 0 {
+        use_syscall(dest)
+    } else if state & STATE_USE_FD != 0 {
+        use_fd(dest)
+    } else {
+        init_loop(dest)
+    }
 }
 
-thread_local!(
-    static RNG_SOURCE: RefCell<Option<RngSource>> = RefCell::new(None);
-);
+fn init_loop(dest: &mut [u8]) -> Result<(), Error> {
+    loop {
+        let state = RNG_STATE.fetch_or(STATE_INIT_ONGOING, Ordering::AcqRel);
 
-fn syscall_getrandom(dest: &mut [u8]) -> Result<(), io::Error> {
+        if state & STATE_INIT_ONGOING != 0 {
+            // initialization is not finished, so wait
+            thread::sleep(time::Duration::from_nanos(1));
+            continue;
+        }
+        return if state & STATE_USE_SYSCALL != 0 {
+            use_syscall(dest)
+        } else if state & STATE_USE_FD != 0 {
+            use_fd(dest)
+        } else {
+            init(dest)
+        };
+    }
+}
+
+fn init(dest: &mut [u8]) -> Result<(), Error> {
+    match use_syscall(&mut []) {
+        Ok(()) => {
+            RNG_STATE.store(STATE_USE_SYSCALL, Ordering::Release);
+            use_syscall(dest)
+        },
+        Err(err) if err.code().get() as i32 == libc::ENOSYS => {
+            match init_fd() {
+                Ok(fd) => {
+                    RNG_FD.store(fd as isize, Ordering::SeqCst);
+                    RNG_STATE.store(STATE_USE_FD, Ordering::SeqCst);
+                    use_fd(dest)
+                },
+                Err(err) => {
+                    RNG_STATE.store(0, Ordering::Release);
+                    Err(err.into())
+                }
+            }
+        },
+        Err(err) => Err(err),
+    }
+}
+
+fn init_fd() -> io::Result<i32> {
+    // read one byte from "/dev/random" to ensure that OS RNG has initialized
+    File::open("/dev/random")?.read_exact(&mut [0u8; 1])?;
+    let f = File::open("/dev/urandom")?;
+    let fd = f.as_raw_fd();
+    mem::forget(f);
+    Ok(fd)
+}
+
+fn use_syscall(dest: &mut [u8]) -> Result<(), Error> {
     let ret = unsafe {
         libc::syscall(libc::SYS_getrandom, dest.as_mut_ptr(), dest.len(), 0)
     };
     if ret < 0 || (ret as usize) != dest.len() {
-        error!("Linux getrandom syscall failed with return value {}", ret);
-        return Err(io::Error::last_os_error());
+        return Err(io::Error::last_os_error().into());
     }
     Ok(())
 }
 
-pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
-    RNG_SOURCE.with(|f| {
-        use_init(f,
-        || {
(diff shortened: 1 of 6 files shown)


src/linux_android.rs
```diff
@@ -10,77 +10,90 @@
...
-fn syscall_getrandom(dest: &mut [u8]) -> Result<(), io::Error> {
+fn use_syscall(dest: &mut [u8]) -> Result<(), Error> {
     let ret = unsafe {
         libc::syscall(libc::SYS_getrandom, dest.as_mut_ptr(), dest.len(), 0)
     };
     if ret < 0 || (ret as usize) != dest.len() {
-        error!("Linux getrandom syscall failed with return value {}", ret);
-        return Err(io::Error::last_os_error());
+        // replace with errno?
+        return Err(io::Error::last_os_error().into());
```
dhardy: I don't think `errno` is exposed in either `std::os` or `libc`? Anyway, this method of capturing it is sufficient.

According to the `errno` man page, the value must be saved immediately after the syscall, so the old

dhardy:
Any plans to extend this to the other platforms (besides WASM) using TLS?

newpavlov:
Yes, but I am not sure if Haiku is considered a part of UNIX family by Rust. Also current implementation has a bug, which I'll fix later.

src/linux_android.rs
newpavlov: No, it's not, but it's possible to copy a bit of code from [here](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/unix/os.rs#L29). The problem with `io::Error::last_os_error().into()` is that IIRC it does not get optimized into pure `errno` call, so there is a small overhead.

newpavlov:
I think I've finished what I wanted to do with non-WASM targets. Result is a bit more complex than that I initially had in mind, but oh well. :) There are certain code duplication, but due to subtle differences I don't see how they can be abstracted nicely. I think orderings are correct, but it's better to double check.

I guess we could remove TLS for WASM targets as well and replace them with "atomics" (which in future may become true atomics) using a similar approach.

newpavlov:
BTW maybe we can remove `rust_122_compat` macro from `__wbg_shims` and simplify relevant code a bit?

dhardy:
Good call, but lets keep that in a separate PR. You can test the WASM stuff by following the CI script.

src/dragonfly_haiku_emscripten.rs
```diff
@@ -0,0 +1,87 @@
...
+use Error;
+use std::fs::File;
+use std::io::Read;
+use std::num::NonZeroU32;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::os::unix::io::{RawFd, AsRawFd, FromRawFd};
+use std::{thread, mem};
+
+// replace with AtomicI32 on stabilization and MSRV bump
+static RNG_FD: AtomicUsize = AtomicUsize::new(-1i32 as usize);
```
dhardy: Why use an unsigned type to store a signed value? `AtomicIsize` is also stable.

src/dragonfly_haiku_emscripten.rs
```diff
@@ -0,0 +1,87 @@
...
+        let state = RNG_STATE.fetch_or(STATE_INIT_ONGOING, Ordering::AcqRel);
+        if state & STATE_INIT_DONE != 0 {
+            // initialization is complete, use fd from atomic
+            break RNG_FD.load(Ordering::Acquire) as RawFd;
+        } else if state & STATE_INIT_ONGOING == 0 {
+            // start initialization and return resulting fd
+            match init_fd() {
+                Ok(fd) => {
+                    RNG_STATE.store(STATE_INIT_DONE, Ordering::Release);
+                    RNG_FD.store(fd as usize, Ordering::Release);
```
dhardy: I believe we need to store the FD *before* the STATE (it is `Ordering::Release` on the last write which matters).

src/dragonfly_haiku_emscripten.rs
```diff
@@ -0,0 +1,87 @@
...
+                    break fd
+                },
+                Err(err) => {
+                    RNG_STATE.store(0, Ordering::Release);
+                    return Err(err);
+                }
+            }
+        }
+        // initialization is not finished, so wait
+        thread::yield_now();
```
dhardy: As you mentioned, it's preferable to use `thread::sleep(Duration::from_millis(1))` here? We can *only* get here if another thread has already started initialisation, so it makes sense to wait.

src/dragonfly_haiku_emscripten.rs
newpavlov: Hm, you are right, but maybe we should even change ordering to `SeqCst` here? Either way I would like someone to specifically review use of atomics in this PR.

src/dragonfly_haiku_emscripten.rs
newpavlov: I've misinterpreted your comment about "wasted CPU cycles" then. But I think here we better to use `thread::sleep(Duration::from_nanos(1))`, either way it will not be smaller than the granularity of underlying clock.

dhardy:
@newpavlov, I would say you've done a good job on this, but I'm still not keen to accept it. The reason is that making heavy use of low-level tools (atomics, bit-ops) makes the code considerably less readable, which makes the code:

- harder to review, thus less trustworthy
- harder to modify (esp. without breaking it)

Also, if we are going to take this approach, it would be good to have someone who really understands atomic synchronisation primitives to review this code.

I'm not saying a definite no to this, but I would like us to look at alternatives first (e.g. an `AtomicBool` is-ready flag + `Once` init + `Cell`/`RefCell` storage, or a higher-level wrapper over atomics).

(I guess I also owe you an apology for the slow review and initially-positive comments; I've had a busy week.)

newpavlov:
While I agree that this code should be carefully reviewed, I think that increasing reader-facing complexity a bit, but getting in return the most efficient implementation is a not bad trade-off for foundational low-level library.

Regarding `Once`,as I've already mentioned earlier, the problem is that initialization function may fail, and with `Once` we lose an ability to retry calls. And `RefCell` approach is less memory efficient, because we have to use `None` as an initialization value.

dhardy:
It may be best then to leave this PR unmerged for now (there's no reason optimisations cannot land in a point release). Would you mind making a separate PR for the module changes + wasm-bindgen clean-up, then rebasing this?

newpavlov:
If you don't want to delay `getrandom` release, then yes, we can postpone this PR, though I hope it will not stay open for too long (i.e. it will be merged or a better approach will be found).

What do you mean by module changes? Also how about bumping MSRV to 1.31 and migrating to 2018 edition? I think 3 release difference is not a big deal, especially considering that upcoming Debian stable will package Rust 1.32.

dhardy:
You merged four modules into one in this PR; I think this is easy to do without the other changes here? Also you renamed 'solarish'; I don't really care about the name but it would be nicer if this PR didn't change it.

If it makes the code significantly nicer, go ahead and use 1.31. There seems to be very little interest in support for old compilers.

dhardy:
I think this could be tidied up a little if you still want to do so, e.g. avoid renaming `solaris_illumos` and rebase.

newpavlov:
Done! Looks like CI failure is due to rust-lang/rust#59731.

To be safe it would be nice to ask someone to review atomics usage.

newpavlov:
@HadrienG2 
Thank you a lot for review! So if I understand correctly the current code is too conservative and orderings can be relaxed, but otherwise it's correct? In that case I think we can go ahead without any changes and open a separate issue/PR for relaxing orderings later.

>it might be a good idea to somehow centralize all the tricky synchronization in a single place that is shared by all implementations

I've tried to think about how it can be done, bit I couldn't find good ideas, each case has a slightly different requirements (some platforms got merged into `use_file.rs`, but that's it), so I don't see an elegant way to abstract it.

@dhardy 
What do you think?

dhardy:
I think it still needs a very careful review, and it would be worth trying to address @HadrienG2's points before this.

Alternatively, we could still avoid TLS by using regular mutexes. Not as efficient perhaps, but I don't think that is so important here, and it's much easier to ensure correctness. (In this case I'm not sure if it's better to use `std::sync` or `parking_lot`; although there was discussion of using the latter to implement `std::sync`, `parking_lot` still has the better API.)

newpavlov:
>it would be worth trying to address @HadrienG2's points before this

I think it will be better to do it in a separate issue/PR, in other words let's introduce a conservative version first and then work on relaxing it if needed. Though I think we better to return to `yield_now()` as was suggested [here](https://users.rust-lang.org/t/28233/4).

src/solaris_illumos.rs
```diff
@@ -57,8 +57,7 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
         let state = RNG_STATE.fetch_or(STATE_INIT_ONGOING, Ordering::AcqRel);
         if state & STATE_INIT_DONE != 0 { break; }
         if state & STATE_INIT_ONGOING != 0 {
-            // initialization is not finished, so wait
-            thread::sleep(time::Duration::from_nanos(1));
+            thread::yield_now()
```
vks: There is a semicolon missing here.

src/solaris_illumos.rs
```diff
@@ -57,8 +57,7 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
         let state = RNG_STATE.fetch_or(STATE_INIT_ONGOING, Ordering::AcqRel);
         if state & STATE_INIT_DONE != 0 { break; }
         if state & STATE_INIT_ONGOING != 0 {
-            // initialization is not finished, so wait
-            thread::sleep(time::Duration::from_nanos(1));
```
vks: The import of the `std::time` is no longer required.

Issue #4, opened 2019-01-19
embedded source compatibility


dhardy:
This crate requires `std::io` code on many platforms to read from special devices, therefore will normally require `std`. Ideally we should provide a smooth route to `no_std` compatibility however, possibly by allowing users to define `extern` functions. (We've discussed using "lang items" but these are restricted to Rust's core libraries; we may or may not integrate this code into `std` in the future and use that option, but ideally should have a `no_std` option now.)

dhardy:
One solution here would be a special feature enabling an alternative mode:
```rust
#[cfg(feature = "getrandom_custom")]
extern "C" {
    fn getrandom_custom(dest: *mut u8, len: usize) -> u32;
}

pub fn getrandom(dest: &mut [u8]) -> Result<(), Error> {
    #[cfg(feature = "getrandom_custom")] {
        use core::num::NonZeroU32;
        let ret = unsafe { getrandom_custom(dest.as_mut_ptr(), dest.len()) };
        return if let Some(err) = NonZeroU32::new(ret) {
            Err(Error::from(err))
        } else {
            Ok(())
        };
    }
    getrandom_inner(dest)
}
```
This would allow *any* single crate in the build to replace the entropy source, which has both advantages (easier testing with dummy entropy, easy of usage of other randomness source libs) and disadvantages (relatively easy to accidentally or possibly even maliciously mess this up; no single place to check whether the alternative is enabled).

---

Another solution (which doesn't need any support here) would be to make a replacement crate with the same API (copy error types, provide `fn getrandom`), then use a patch section in the top `Cargo.toml` to replace this.
```toml
[patch.crates-io]
getrandom = { path = "my/replacement/getrandom" }
```

dhardy:
There is another issue with `no_std` support: the error type should implement `std::error::Error` iff using `std`. Currently the lib does so on all platforms except SGX, i.e. this is automatic and requires `std` on all platforms which *should* support `std`. IIRC we did once have a reason to disable support on Mac OS.

In contrast, `rand_core` makes std-support a crate feature, and causes confusing errors if you get this wrong in the build.

dhardy:
Note: there is a long discussion on this topic in rand#579 with several sub-topics:

- [a hook to set an entropy source](https://github.com/rust-random/rand/pull/579#issuecomment-410464293)
- [synchronisation/shared memory](https://github.com/rust-random/rand/pull/579#issuecomment-417260440)
[...]

PR #15, opened 2019-03-22
MSRV bump, module changes

newpavlov:
Note that with this PR `getrandom` by default will not use anything from `std` anymore, so it will be possible to implement #4.


.travis.yml
```diff
@@ -3,12 +3,12 @@ sudo: false
...
-    - rust: 1.28.0
-      env: DESCRIPTION="Linux, 1.28.0"
+    - rust: 1.32.0
+      env: DESCRIPTION="Linux, 1.32.0"
       os: linux
 
-    - rust: 1.28.0
-      env: DESCRIPTION="OSX, 1.22.0"
+    - rust: 1.32.0
+      env: DESCRIPTION="OSX, 1.32.0"
```
dhardy: Is there any reason not to use 1.31 as the MSRV? Also you need to update the note in the README.

dhardy:
```
error[E0468]: an `extern crate` loading macros must be at the crate root
  --> src/wasm32_stdweb.rs:11:14
   |
11 | #[macro_use] extern crate stdweb;
   |              ^^^^^^^^^^^^^^^^^^^^
```
Can we use Edition 2018 style macro imports?

.travis.yml
newpavlov: `extern crate std;` and `use std::<..>;` in modules does not work on 1.31.0. It fails with "error[E0658]: imports can only refer to extern crate names passed with `--extern` on stable channel (see issue #53130)".

newpavlov:
I looks like `stdweb` does not work nicely with 2018 imports. I think I've worked around it, but ideally I would like to remove `stdweb` support in favour of `wasm-bindgen`. IIUC the necessary work on bindgen side is already [done](https://github.com/rustwasm/wasm-bindgen/pull/1295).

.travis.yml
dhardy: Fair enough. 1.32 is barely over two months old, but there doesn't appear to be much demand for back compatibility. Perhaps we should not do the same for the main Rand project yet though just in case we need to revert this.

dhardy:
Seems like your workaround does the job for now. It may require fixing again later, but both `stdweb` and `wasm_bindgen` are being actively maintained and with not so different usage numbers (both somewhat low actually), so IMO we shouldn't abandon either yet.

This is ready for merging then?

newpavlov:
Yes, it's ready to be merged.

Regarding WASM, I think we should understand how future changes will be handled. Say we publish v0.1 and then `stdweb` becomes compatible with `wasm-bidgen`. Or new web/nodejs specific targets will be introduced. The most conservative way will be to publish v0.2 and propagate version bump across other rand crates dependent on `getrandom`, but I think it will be too much churn. Another approach is to state that both WASM features are experimental and not covered by backward compatibility promise. And the last approach is to move both WASM back-ends to separate crates and tell WASM users to use solution from #4. All three approaches have their own pros and cons, so maybe we should open a separate issue for discussing it.

dhardy:
I think if both `stdweb` and `wasm-bindgen` features are enabled, then only `wasm-bindgen` will actually be used? So if the two libs become compatible, I don't think that's a problem for us.

As for our implementation, if we choose to deprecate one, we can simply add a `stdweb` feature depending on the `wasm-bindgen` crate and implementation, so no problem there either that I can see.

PR #16, opened 2019-03-23
Prepare 0.1.0 release

dhardy:
@newpavlov I think we are ready for the initial release?

This would mark the lib as ready for testing, though the real testing happens when Rand 0.7 reaches pre-release status.


newpavlov:
Yes, I think we can go ahead with the initial release. I still have some concerns regarding WASM targets and ideally would prefer to remove TLS, but those issues are not important enough to delay release.

Maybe though it would make sense to do a pre-release first and do v0.1 release synchronously with rand v0.7 release a bit later?

dhardy:
I do not see any advantage to making a pre-release when there is no existing code to break.

Version "0.1.0" already says clearly that this is an early version. Pre-release versions get almost no testing from past experience.

PR #18, opened 2019-04-05
Prepare 0.1.1 release

dhardy:


newpavlov:
I wouldn't say it's a "fix", I would write something like "enable std functionality for CloudABI by default".

Also how about adding zero buffer test in this PR as well?

dhardy:
Oh, sure. Shall I close this and let you do that?

newpavlov:
I think you can simply add the test as part of this PR.

dhardy:
Okay, I will do it.

PR #20, opened 2019-04-06
Added support for wasm32-unknown-wasi

nipzu:
This update adds support for the `wasm32-unknown-wasi` compilation target. Random data is generated using the `__wasi_random_get` function.


nipzu:
I hope that works.

dhardy:
Looks good enough. If you care to bump the version to 0.1.2 and add a note in the changelog I can make a new release.

PR #22, opened 2019-05-15
Update the WASI support.

sunfishcode:
- WASI is now categorized as an OS rather than an environment.
   (https://github.com/rust-lang/rust/pull/60117)
 - The WASI triple is renamed from wasm32-unknown-wasi to wasm32-wasi.
   (https://github.com/rust-lang/rust/pull/60585)


sunfishcode:
Bumped and added!

dhardy:
Perfect!

PR #23, opened 2019-06-10
Support Rustfmt

josephlr:
We will actually run `rustfmt` in a later PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,13 @@ matrix:
     - rust: beta
       env: DESCRIPTION="Linux, beta"
 
+    - rust: stable
+      env: DESCRIPTION="Rustfmt"
+      install:
+        - rustup component add rustfmt
+      script:
+        - cargo fmt --all -- src/*.rs --check
+
     - rust: nightly
       os: linux
       env: DESCRIPTION="Linux, nightly, docs"
@@ -136,6 +143,10 @@ matrix:
         - source ~/.cargo/env || true
       script:
         - bash utils/ci/script.sh
+
+  allow_failures:
+    # Formatting errors should appear in Travis, but not break the build.
+    - env: DESCRIPTION="Rustfmt"
 
 before_install:
   - set -e


dhardy:
Does this actually help anyone? I find it needless fiddling, debatable whether the result is an improvement, and potentially clashing with other PRs.

josephlr:
@dhardy ideally there would be a way to just make sure that new PRs were properly formatted. That way the git blame wouldn't get messed up by switching over to this. However, `rustfmt` doesn't support such a thing IIRC. So the options are either:
   - Not having any formatting tools for PRs
   - This change (or something equivalent)

josephlr:
As for clashing with other PRs, there's really only https://github.com/rust-random/getrandom/pull/14 and it's simple enough to just run Rustfmt on that PR.

dhardy:
> make sure that new PRs were properly formatted

In my book, "properly" means simply *reasonably neat and consistent*. Actually, I may be biased because of bad experiences with `rustfmt` in the past (frequently finding I have the wrong version, and it being impossible to work with old commits without bumping into formatting issues).

> That way the git blame wouldn't get messed up

I don't count this very important (it's very easy to modify a line in insignificant ways).

> As for clashing with other PRs

Sorry, this isn't an issue here.

I'll let @newpavlov decide since he's mostly responsible for this repo, but in my book having to do extra work to get PRs merged for very little gain is just not worth it.

newpavlov:
I also had a bad experience with `rustfmt`, but I think consistent formatting is not a bad thing to have. (though personally I dislike some defaults) But I believe `rustfmt` failure should be allowed in CI, as otherwise it's quite irritating for contributors. We can run it manually from time to time.

I think it will be better to revert formatting changes in this PR and postpone them until other PRs get sorted out. In other words let's limit it to CI changes.

josephlr:
> I think it will be better to revert formatting changes in this PR and postpone them until other PRs get sorted out. In other words let's limit it to CI changes.

Done, the check now prints the diff to the Travis CI log, but will not fail the build. I'll do a separate CL to actually run rustfmt once all the other PR are merged/closed.

newpavlov:
I think it will be better use [`allow_failures`](https://docs.travis-ci.com/user/customizing-the-build#rows-that-are-allowed-to-fail) for `rustfmt` test (it will make it stand out more on CI page) and remove `cargo test`s (they are simply redundant).

josephlr:
> I think it will be better use [`allow_failures`](https://docs.travis-ci.com/user/customizing-the-build#rows-that-are-allowed-to-fail) for `rustfmt` test (it will make it stand out more on CI page) and remove `cargo test`s (they are simply redundant).

Are the `cargo test`s redundant? I thought they were needed to make sure the tests pass on stable w/ Linux.

josephlr:
This should be ready to merge. The `rustfmt` job fails, but the build as a whole succeeds.

PR #24, opened 2019-06-10
Improve RDRAND implementation

josephlr:
This change makes a few improvements to be closer to Intel's recommendations

1. The value of `RETRY_LIMIT` is the one recommended by Intel.
2. Use `chunks_exact_mut` to elide most calls to `memcpy`. See [generated assembly](https://rust.godbolt.org/z/JgbjFj).
3. Remove all `unsafe` code except that used to invoke the RDRAND intrinsic.
4. Adds Travis build check for `x86_64-fortanix-unknown-sgx`

Stylistically, the code is now much easier to read. The file has also been renamed to `rdrand.rs` to better reflect what it does. [This diff](https://github.com/rust-random/getrandom/pull/24/commits/91db682cbee6a5cf4c382f8ccc89c7970ed707e3) makes it easier to see what has actually changed.


src/rdrand.rs
```diff
@@ -0,0 +1,54 @@
...
+    if slice.len() > 0 {
+        let mut data = unsafe { mem::uninitialized() };
+        get_rand_u64(&mut data)?;
+        slice.copy_from_slice(&data.to_ne_bytes()[..slice.len()])
+    }
+    Ok(())
+}
+
+pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
+    let (prefix, aligned, suffix) = unsafe { dest.align_to_mut() };
```
newpavlov: Citing `align_to_mut` docs:

>The method does a best effort to make the middle slice the greatest length possible for a given type and input slice, but only your algorithm's performance should depend on that, not its correctness.

In other words strictly speaking it does not guarantee that `prefix` and `suffix` are smaller or equal to 8 bytes and you rely on it in `fill_small_slice`. Looking at generated assembly we can see that `slice_index_len_fail` is still present, so compiler can't prove that their length is smaller or equal to 8 bytes.

One option is to explicitly construct slices without relying on `std` method.

src/rdrand.rs
```diff
@@ -0,0 +1,54 @@
...
+        };
+    }
+    Err(Error::UNKNOWN)
+}
+
+fn fill_small_slice(slice: &mut [u8]) -> Result<(), Error> {
+    if slice.len() > 0 {
+        let mut data = unsafe { mem::uninitialized() };
+        get_rand_u64(&mut data)?;
+        slice.copy_from_slice(&data.to_ne_bytes()[..slice.len()])
```
newpavlov: `copy_from_slice` results in `memcpy` call instead of a simple move. Maybe keep the old `ptr::write_unaligned`?

newpavlov: Never mind, we can't use `write_unaligned` for the second `fill_small_slice`, as we will overwrite data outside of the provided slice.

newpavlov: BTW this line results in "cannot borrow `*slice` as immutable because it is also borrowed as mutable" compilation error on Rust 1.34.

src/rdrand.rs
josephlr: I decided just to change `get_rand_unaligned` to support slices longer than 8 bytes, so we are now properly relying on `align_to_mut` for performance rather than correctness. Furthermore, the [generated assembly](https://rust.godbolt.org/z/wcFE62) now:
  - Contains no `slice_index_len_fail` or panic hooks, so the compiler can elide all of the slice checks.
  - Inlines all the functions (as expected).

src/rdrand.rs
josephlr: Ya it looks like I was implicitly relying on NLL, should be fixed now. There's no way to implement this without a call to `memcpy`, as the provided slice might be smaller than 8 bytes.

josephlr:
@newpavlov this should be ready to merge, provided you're OK with an implementation that needs up to two calls to `memcpy` at the beginning and end of `dest`.

I have an [alternative implementation](https://rust.godbolt.org/z/QzdYjL) that only calls `memcpy` on small slices. On large slices, it calls `write_unaligned` at the beginning and end of `dest` (writing to some of the memory twice). While the alternative implementation also gives a smaller binary, it has to use much more unsafe code to stop the compiler from emitting `slice_index_len_fail` or panic hooks.

newpavlov:
Personally I prefer the alternative implementation. We even could drop `0 => {}` branch to make binary even smaller, as I guess calls of`getrandom` with zero-length slices will be really rare.

If we'll decide to keep the current implementation, then  you can make `get_rand_unaligned` slightly more idiomatic, while keeping the efficiency:
```rust
pub fn get_rand_unaligned(dest: &mut [u8]) -> Result<(), Error> {
    const U64_SIZE: usize = mem::size_of<u64>();
    for chunk in dest.chunks_mut(U64_SIZE) {
        let data = get_rand_u64()?;
        let n = cmp::min(chunk.len(), U64_SIZE);
        chunk.copy_from_slice(&data.to_ne_bytes()[..n]);
    }
    Ok(())
}
```

@dhardy 
What do you think?

josephlr:
> If we'll decide to keep the current implementation, then you can make `get_rand_unaligned` slightly more idiomatic, while keeping the efficiency:

Oh that's much nicer, I just updated this PR to use essentially that code. The [generated assembly](https://rust.godbolt.org/z/DR2vDB) still gives what we expect.

> Personally I prefer the alternative implementation. We even could drop `0 => {}` branch to make binary even smaller, as I guess calls of `getrandom` with zero-length slices will be really rare.

I'm still trying to figure out if there is a way to get the alternative implementation working without as much unsafe code. I'll let you know my progress.

josephlr:
So after looking at this some more, it’s probably not worth it to hyper-optimize this to maximize the number of aligned accesses. On x86 the same code gets emitted for aligned/unaligned accesses anyway (unaligned ones just happen to be slower).

I’ll rewrite this to just use chunks_mut_exact, should make everything fairly short.

Note: as most of the time this is being used to fill a freshly allocated  128/256 but key, everything will usually be properly aligned regardless.

josephlr:
@dhardy @newpavlov this should be ready for final review/merging, see the updated PR description.

src/rdrand.rs
```diff
@@ -0,0 +1,55 @@
...
+use core::num::NonZeroU32;
+
+#[cfg(not(target_feature = "rdrand"))]
+compile_error!("enable rdrand target feature!");
+
+// Recommendation from "Intel® Digital Random Number Generator (DRNG) Software
+// Implementation Guide" - Section 5.2.1 and "Intel® 64 and IA-32 Architectures
+// Software Developer’s Manual" - Volume 1 - Section 7.3.17.1.
+const RETRY_LIMIT: usize = 10;
+const WORD_SIZE: usize = mem::size_of::<usize>();
```
newpavlov: Why do you use word size instead of `u64` size? `_rdrand64_step` explicitly works with `u64` and not with `usize`. Yes, `x86_64-fortanix-unknown-sgx` is a 64-bit target, but I think it will be better to use `u64` here.

josephlr: Whoops, I accidentally committed some of my changes that will make this work on 32-bit x86, should be fixed now.

PR #26, opened 2019-06-10
Don't block in getrandom() detection

josephlr:
On Linux, we detect if `getrandom` is present by calling `getrandom` with
an empty buffer and seeing if `ENOSYS` is returned. However, even with
an empty buffer, this call will block unless we explicitly pass a flag.

This can be seen in the source for `getrandom`:
  https://elixir.bootlin.com/linux/v5.1.8/source/drivers/char/random.c#L2043

This change adds a boolean parameter to `syscall_getrandom` to control
the blocking behavior. We now don't block in initialization, but do
block when actually reading random data.

### DIFF
--- a/src/linux_android.rs
+++ b/src/linux_android.rs
@@ -16,6 +16,8 @@ use core::cell::RefCell;
 use core::num::NonZeroU32;
 use core::sync::atomic::{AtomicBool, Ordering};
 
+// This flag tells getrandom() to return EAGAIN instead of blocking.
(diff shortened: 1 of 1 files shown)


newpavlov:
I also think that blocking is not a big deal. If kernel does not have `getrandom` syscall we immediately get `ENOSYS`, and if call blocks, the next `getrandom` call will block either way. But I guess with this PR we will get a slightly better behavior in multi-threaded programs on a system launch, i.e. instead of blocking on user-land synchronization, threads will be blocked on syscall.

PR #28, opened 2019-06-12
macos: Cleanup SecRandom type

josephlr:
Right now `SecRandom` is an uninhabited type. However, opaque pointers
are not supposed to be pointers to uninhabited types or ZSTs. This
issue was solved in `core` by introducing `core::ffi:c_void` that just
does all the right things.

This PR:

  - Changes `SecRandom` to be a `transparent` wrapper around `c_void`.
  - Links `kSecRandomDefault` instead of defining it ourself.
  - Should not change any behavior on macos.

### DIFF
--- a/src/macos.rs
+++ b/src/macos.rs
@@ -11,7 +11,6 @@ extern crate std;
 
 use crate::Error;
 use std::io;
-use core::ffi::c_void;
 use core::num::NonZeroU32;
 
 // TODO: Make extern once extern_types feature is stabilized. See:
--- a/src/macos.rs
+++ b/src/macos.rs
@@ -11,29 +11,23 @@ extern crate std;
 
 use crate::Error;
 use std::io;
+use core::ffi::c_void;
 use core::num::NonZeroU32;
 
-enum SecRandom {}
-
-/// Essentially a null pointer (type `SecRandomRef`)
-#[allow(non_upper_case_globals)]
-const kSecRandomDefault: *const SecRandom = 0 as *const SecRandom;
+// TODO: Make extern once extern_types feature is stabilized. See:
+//   https://github.com/rust-lang/rust/issues/43467
+#[repr(C)]
+struct SecRandom([u8; 0]);
 
 #[link(name = "Security", kind = "framework")]
 extern {
-    fn SecRandomCopyBytes(
-        rnd: *const SecRandom, count: libc::size_t, bytes: *mut u8,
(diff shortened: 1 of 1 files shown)


newpavlov:
IIUC this is one of the use-cases for extern types (currently unstable) and uninhabited types is a common workaround, which I've seen in several places. I am not sure if this PR is correct, citing the `c_void` docs:
>To model pointers to opaque types in FFI, until extern type is stabilized, it is recommended to use a newtype wrapper around an empty byte array.

src/macos.rs
```diff
@@ -11,29 +11,23 @@ extern crate std;
 
 use crate::Error;
 use std::io;
+use core::ffi::c_void;
```
newpavlov: You don't need this import anymore.

josephlr:
> I wasn't aware of this. I found the section on [Handling Zero-Sized Types](https://doc.rust-lang.org/nomicon/vec-zsts.html?highlight=pointer#handling-zero-sized-types) in the nomicon, but it only talks about issues with allocation and iterators — we don't appear to do either here.
> 
> What our version of `OpaquePointer` _is_ missing is alignment info, though given that we only ever construct the address `0` it shouldn't matter.

> IIUC this is one of the use-cases for extern types (currently unstable) and uninhabited types is a common workaround, which I've seen in several places. I am not sure if this PR is correct, citing the `c_void` docs:

Whoops, missed that. Fixed to do what the nomicon says (use a private ZST field) and link to the [Github issue on `extern_type`](https://github.com/rust-lang/rust/issues/43467).

> Also, I guess usize is always the same as libc::size_t but can we actually rely on that?

So this assumption is built deep into how Rust interfaces with C libraries. If a platform is supportted by `libc`, [`size_t` is a typedef for `usize`](https://github.com/rust-lang/libc/blob/e0ff1e68b9e34173e9c4c3217d1b0fc81a7d352d/src/unix/mod.rs#L19). This is assumption is also [built into things like `memcpy`](https://docs.rs/rlibc/1.0.0/rlibc/fn.memcpy.html). So I'd say it is about as safe a relying on [`u8, u16, u32, u64` being the same as `libc::{uint8_t, uint16_t, uint32_t, uint64_t}`](https://github.com/rust-lang/libc/blob/master/src/fixed_width_ints.rs)

src/macos.rs
josephlr: Fixed

PR #30, opened 2019-06-12
Support x86_64-unknown-uefi

### LINKED ISSUE #27: rdrand implementation should enable the feature itself
Current implementation of rdrand code fails to compile if the user does not enable the `rdrand` feature during compilation. Instead, the code should enable the feature for the relevant portions of the code itself with `#[target_feature(enable="rdrand")]`.

josephlr:
Closes #27 and adds support for the [`x86_64-unknown-uefi` target](https://github.com/rust-lang/rust/pull/56769). For our purposes, this target is quite similar to SGX in that the only source of RNG is RDRAND.

This PR:
  - Uses [`cargo xbuild`](https://github.com/rust-osdev/cargo-xbuild) to make sure we can build on this target.
  - Adds in CPUID detection (only run if `cfg!(not(target_feature = "rdrand"))`)

As a consequence, SGX targets that forgot to set the `+rdrand` feature will now work with `getrandom`.


josephlr:
Note that because this removed the compiler error when `cfg(not(target_feature = "rdrand"))` all the rdrand methods are now unsafe (as it is UD behavior to call them if rdrand is not supported). We also need to keep these functions `unsafe` as they need the `#[target_feature(enable = "rdrand")]` annotation to enable rdrand to be inlined (which is critical for performance in this case). See the difference:
  - [No annotations, and therefore no inlining and terrible output](https://rust.godbolt.org/z/zs52DY)
  - [With annotations and cpuid-based detection](https://rust.godbolt.org/z/ZwxEQ6)
  - [With annotations and `target-feature=+rdrnd`](https://rust.godbolt.org/z/fjqzTQ), CPUID checks are properly elided

Also, @dhardy @newpavlov to use `std_detect` (outside of `std`) requires nightly. Is this acceptable? If not I can just copy the rdrand detection code from there (it's only 5 lines).

dhardy:
I'm still not convinced whether we should omit the CPU detection just to avoid a CPUID test. @nagisa?

I think there's no other reason we can't support SGX with stable compilers? If so then copying the code sounds best.

josephlr:
> I'm still not convinced whether we should omit the CPU detection just to avoid a CPUID test.

To be clear, the CPUID test is only omitted if the compiler already knows it can support RDRAND. This is crucial for SGX, as the CPUID instruction is illegal in an enclave.

> If so then copying the code sounds best.

Will do

nagisa:
You should just use the `is_x86_target_feature_detected` macro. It does the
sane thing - caches results (globally) and does the detection in a target
specific way which is more likely to work compared to whatever you might
want to maintain yourselves.

nagisa:
Caveat: the macro is not supported on non-std targets. In that case you
still want to maintain something similar.

josephlr:
> Caveat: the macro is not supported on non-std targets. In that case you
still want to maintain something similar.

Ya this is the issue, because most of the targets we would want to use RDRAND on are `no_std` (like UEFI, or SGX targets that aren't `x86_64-fortanix-unknown-sgx`). I added a comment linking to the `stdsimd` issue.

josephlr:
@dhardy @nagisa does the `is_rdrand_supported()` look correct?

nagisa:
https://github.com/nagisa/rust_rdrand/blob/fa7b3b13f393a1eaf345fb8d207631ca2a8a78bc/src/lib.rs#L143 is what I have in the rdrand library, your’s seems equivalent.

src/rdrand.rs
```diff
@@ -9,33 +9,57 @@
...
+        const FLAG: u32 = 1 << 30;
+        lazy_static! {
+            static ref HAS_RDRAND: bool = unsafe { __cpuid(1).ecx & FLAG != 0 };
+        }
+        *HAS_RDRAND
+    }
+}
+
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
+    if is_rdrand_supported() {
```
dhardy: Missing a "not" sign here?

josephlr: *face-palm*

josephlr: Fixed

src/rdrand.rs
```diff
@@ -9,33 +9,57 @@
...
     Err(Error::UNKNOWN)
 }
 
+// TODO use is_x86_feature_detected!("rdrand") when that works in core. See:
+//   https://github.com/rust-lang-nursery/stdsimd/issues/464
+fn is_rdrand_supported() -> bool {
+    if cfg!(target_feature = "rdrand") {
+        true
+    } else if cfg!(target_env = "sgx") {
+        false // No CPUID in SGX enclaves
```
newpavlov: Shouldn't we issue a compilation error here?

josephlr: I'm just copying what's in `std_detect`, and I don't think we should deviate from their implementation.

josephlr: Also, if we were going to deviate from `std_detect`, it would probably be to just always assume RDRAND is present on SGX, see https://github.com/nagisa/rust_rdrand/issues/9. But I don't think that is a good idea either.

josephlr:
> Is there a way to detect if EFI_RNG_PROTOCOL is implemented and use RDRAND as a fallback?

Not without access to the EFI system table. Using `EFI_RNG_PROTOCOL` would be possible in theory if solving #4 gave us a way to externally provide randomness, but in practice `EFI_RNG_PROTOCOL` is almost always either unimplemented or aliased to RDRAND.

src/rdrand.rs
newpavlov: I mean why not just use:
```rust
#[cfg(not(and(target_feature = "rdrand", target_env = "sgx")))]
compile_error!("..");
```
and drop the SGX branch. If somehow code which targets SGX gets compile without `rdrand` feature (considering that it's enabled by default such event should be quite unlikely), developer will get a compilation error instead of always failing application.

josephlr: Added a check for SGX && !RDRAND

newpavlov: Ups, you should use `all` instead of `and`.

dhardy:
Agreed, that sounds like a needless complication.

```
$ cargo build --target=x86_64-fortanix-unknown-sgx --all-features
   Compiling getrandom v0.1.3 (/home/travis/build/rust-random/getrandom)
error[E0537]: invalid predicate `and`
  --> src/rdrand.rs:34:7
   |
34 | #[cfg(and(target_env = "sgx", not(target_feature = "rdrand")))]
   |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
I think you want `all`.

src/lib.rs
```diff
@@ -190,6 +190,7 @@ mod_use!(cfg(target_os = "redox"), use_file);
 mod_use!(cfg(target_os = "solaris"), solaris_illumos);
 mod_use!(cfg(windows), windows);
 mod_use!(cfg(target_env = "sgx"), rdrand);
+mod_use!(cfg(target_os = "uefi"), rdrand);
```
newpavlov: I think we should use `all(target_os = "uefi", target_arch="x86_64")`.

josephlr:
> I think you want all.

Fixed

src/lib.rs
josephlr: Fixed here and below

src/rdrand.rs
```diff
@@ -31,13 +31,14 @@ unsafe fn rdrand() -> Result<[u8; WORD_SIZE], Error> {
     Err(Error::UNKNOWN)
 }
 
+#[cfg(and(target_env = "sgx", not(target_feature = "rdrand")))]
+compile_error!("SGX targets must enable RDRAND to get randomness");
```
dhardy: What does "enable RDRAND" mean? We know, but I expect it could confuse some users. Better to say "... must enable the `rdrand` feature of `getrandom` crate"?

newpavlov: No, it should be "enable `rdrand` target feature for your build" or something similar.

josephlr: Fixed to be `Target feature +rdrnd must be enabled for SGX targets.` Note that `+rdrand` is not a typo, see: https://github.com/rust-lang/rust/blob/0dc9e9c10ca6dc78cba8b9f9b15038c977b10a77/src/librustc_target/spec/x86_64_fortanix_unknown_sgx.rs#L56

newpavlov: I think you should use `rdrand`, `rdrnd` is a LLVM name, while `rdrand` is a "vendor" name. And it was decided for Rust compiler to follow "vendor" feature names. IIRC Rust compiler translates `rdrand` to `rdrnd` before moving to LLVM, so compiler users should use the former.

Cargo.toml
```diff
@@ -39,5 +39,8 @@ stdweb = { version = "0.4.9", optional = true }
 [target.wasm32-wasi.dependencies]
 libc = "0.2.54"
 
+[target.'cfg(any(target_env = "sgx", target_os = "uefi"))'.dependencies]
+lazy_static = { version = "1.3.0", features = ["spin_no_std"] }
```
newpavlov: Note that `spin_no_std` feature depends on unstable language features, so SGX and UEFI targets will not work on stable compiler.

josephlr: Does it? I was able to `cargo build --target x86_64-fortanix-unknown-sgx` for getrandom without any issues on current stable Rust. And the `lazy_static` and `spin` crates don't seem to use unstable features.

src/rdrand.rs
josephlr: The problem is that the target feature provided to rustc *must* be `+rdrnd`. If you provide `+rdrand` it does not work. See [this example](https://rust.godbolt.org/z/TXrbCy), with `+rdrand` the checks are not removed, but with `+rdrnd` they are.

dhardy: Okay, using a target feature for configuration makes sense if does what we need, but why then are you using `#[target_feature(enable = "rdrand")]` in the code (not `rdrnd`)?

josephlr: Because when doing detection in the code you have to use "rdrand" (I know, it's inconsistent). If you use `rdrnd` in the rust code, you get:

```
error: the feature named `rdrnd` is not valid for this target
```

newpavlov: I believe it's a `rustc` bug. I've created an issue: rust-lang/rust#61895

Cargo.toml
newpavlov: Hm, I was following `lazy_static` docs: https://docs.rs/lazy_static/1.3.0/lazy_static/#cargo-features Maybe it's out of date?

src/rdrand.rs
newpavlov: For the time being we could write something like this:
>`rdrand` target feature must be enabled for SGX targets (e.g. by using `-C target-feature=+rdrnd`, see: rust-lang/rust#61895)

And we will change this message depending on how the issue will be handled .

Cargo.toml
josephlr: Yup, was fixed in https://github.com/rust-lang-nursery/lazy-static.rs/pull/130 (for 1.2.0), but the docs didn't get updated until https://github.com/rust-lang-nursery/lazy-static.rs/pull/144 (after the current release).

src/rdrand.rs
josephlr: I added something like the above in and added a comment explaining the `rdrand` vs `+rdrnd` disparity.

newpavlov: You have a typo in the comment:
>requires "+rdrand" flag

josephlr: Fixed, I really hope https://github.com/rust-lang/rust/issues/49653 gets resolved

newpavlov:
@dhardy 
I think we can merge it?

src/rdrand.rs
```diff
@@ -9,33 +9,61 @@
...
+unsafe fn rdrand() -> Result<[u8; WORD_SIZE], Error> {
     for _ in 0..RETRY_LIMIT {
-        unsafe {
-            // SAFETY: we've checked RDRAND support, and u64 can have any value.
-            let mut el = mem::uninitialized();
-            if _rdrand64_step(&mut el) == 1 {
-                return Ok(el.to_ne_bytes());
-            }
-        };
+        let mut el = mem::uninitialized();
```
tarcieri: A bit late to the party here, but I was curious why this uses `mem::uninitialized()` instead of `mem::zeroed()`.

Zeroing seems relatively lightweight compared to RDRAND itself, would make RNG failures more obvious, and eliminates the potential risk of using attacker-controlled values which appear plausibly random at first glance as e.g. cryptographic keys.

newpavlov: Yeah, `zeroed` could be a better choice. Although in practice it does not matter, as both `zeroed` and `uninitialized` get [optimized out](https://rust.godbolt.org/z/tBks1s).

newpavlov: Fixed in the latest [commit](https://github.com/rust-random/getrandom/commit/1705b40655e99a3b3f974013ef31de9cd880864a).

josephlr: @tarcieri good catch! It also gets rid of a deprecation warning on the latest nightly.

Issue #35, opened 2019-06-15
Use getrandom on FreeBSD


myfreeweb:
[`getentropy` in libc](https://github.com/freebsd/freebsd/blob/master/lib/libc/gen/getentropy.c) uses `sysctl` only as a fallback and uses the `getrandom` syscall if available.

This crate should check for `getrandom` via `dlopen` and use it if available.

josephlr:
Is there a good reason to add this? If we have to keep the sysctl approach around regardless, it just seems like more maintenance to maintain two equivalent implementations for FreeBSD.

myfreeweb:
The sysctl is slower and does not block by itself (as in, it would occasionally return less bytes and your current code would return an error, it [must be used in a loop](https://github.com/freebsd/freebsd/blob/d946d8f14c81df5c94524f0c759db84880bbcae8/lib/libc/gen/getentropy.c#L64-L71))

josephlr:
> The sysctl is slower

So it looks like the only reason this is true is because the sysctl only allows up to 256 bytes per call, while the getrandom syscall allows very large reads per call. This only seems relevant if we are bottlenecked by the ring transition overhead. Are there any metrics on this?

> and does not block by itself (as in, it would occasionally return less bytes and your current code would return an error, it must be used in a loop)

Good catch, fixed in: #36 

Looking at [FreeBSD's `getrandom(2)` man page](https://www.freebsd.org/cgi/man.cgi?query=getrandom&apropos=0&sektion=0&manpath=FreeBSD+12.0-stable&arch=default&format=html#end), it would seem that we would have to do this loop regardless, as it does not guarantee the completion of large buffer requests.

@myfreeweb to do this we need to add libc related bindings (this was done for Linux in https://github.com/rust-lang/libc/pull/1399). Is there a better way to detect if the syscall is supported? On Linux, this is indicated by the syscall returning ENOSYS, does FreeBSD do something similar?

myfreeweb:
Do not call the syscall directly, FreeBSD's public ABI is the libc.

Just do what you're doing on Illumos:

https://github.com/rust-random/getrandom/blob/0b87a0974b75357d55d6ec2c0358a81371294716/src/solaris_illumos.rs#L93

newpavlov:
@myfreeweb 
Can you prepare PR with suggested changes?

PR #36, opened 2019-06-17
freebsd: Increase consistancy with libc implementation

josephlr:
In #35 @myfreeweb pointed out that the [FreeBSD's libc implemenation](https://github.com/freebsd/freebsd/blob/d946d8f14c81df5c94524f0c759db84880bbcae8/lib/libc/gen/getentropy.c#L64-L71) calls `kern_arnd` in a loop, and simply retries if the syscall returns a short buffer.

There doesn't seem like a good reason to differ from their implementation, so this PR changes the implementation to be more like the one for Linux `getrandom`.

### DIFF
--- a/src/freebsd.rs
+++ b/src/freebsd.rs
@@ -14,21 +14,30 @@ use std::io;
 use core::ptr;
 use core::num::NonZeroU32;
 
+fn kern_arnd(buf: &mut [u8]) -> Result<usize, Error> {
+    static MIB: [libc::c_int; 2] = [libc::CTL_KERN, libc::KERN_ARND];
+    let mut len = buf.len();
+    let ret = unsafe {
+        libc::sysctl(
+            MIB.as_ptr(),
+            MIB.len() as libc::c_uint,
+            buf.as_mut_ptr() as *mut _,
+            &mut len,
+            ptr::null(),
+            0,
+        )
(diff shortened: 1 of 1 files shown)


dhardy:
And yet that implementation still has a [limit of 256 bytes](https://github.com/freebsd/freebsd/blob/d946d8f14c81df5c94524f0c759db84880bbcae8/lib/libc/gen/getentropy.c#L115), so unless someone wants to check up on the kernel source code we should keep the chunking I think.

newpavlov:
IIUC if we provide buffer larger than 256 bytes, kernel [will fill](https://github.com/freebsd/freebsd/blob/48905db5ab901055e23e890e49a487120536b8e9/sys/kern/kern_mib.c#L157) only first 256 bytes and will return 256. Thus we don't need explicit chunking.

josephlr:
> IIUC if we provide buffer larger than 256 bytes, kernel will fill only first 256 bytes and will return 256. Thus we don't need explicit chunking.

This is correct, I also think the the 256 bytes is an implementation detail; I couldn't find documentation for the behavior. In that case, this implementation no longer makes assumptions about how KERN ARND works.

dhardy:
In that case, :+1: . Over to you @newpavlov

PR #39, opened 2019-06-26
Run Rustfmt on entire repo

josephlr:
This PR keeps `allow_failures` in the Travis presubmit, we should just run rustfmt from time to time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ matrix:
       install:
         - rustup component add rustfmt
       script:
-        - cargo fmt --all -- src/*.rs --check
+        - cargo fmt --all -- */*.rs --check
 
     - rust: nightly
       os: linux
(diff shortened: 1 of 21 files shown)


newpavlov:
Personally I am not sure about removing `allow_failures`, in my experience it results in a more annoyance than it's worth.

josephlr:
> Personally I am not sure about removing allow_failures, in my experience it results in a more annoyance than it's worth.

I mean all it does is force PRs to be properly formatted when they are submitted. Or is it more a concern about `rustfmt`'s stability?

newpavlov:
Both, I had experienced CI breakage on my projects due to `rustfmt` changes, and it will be annoying for contributors who don't use `rustfmt`.

josephlr:
@newpavlov that makes sense, fixed

PR #40, opened 2019-06-26
Remove cloudabi, winapi, and fuchsia-cprng dependancies

josephlr:
Depends on #39 and helps address issues raised in https://github.com/rust-lang/rust/pull/62082#issuecomment-505756669

We don't really need these external crates as they just provide declarations of system functions that we can just `extern` declare ourselves. 

This PR also updates the docs and cleans up `Cargo.toml` to explain why we need `lazy_static` on certain targets.

This also fixes the fuchsia CI issues.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,19 +23,18 @@ log = { version = "0.4", optional = true }
 [target.'cfg(any(unix, target_os = "wasi"))'.dependencies]
 libc = "0.2.54"
 
-[target.'cfg(unix)'.dependencies]
+# For holding file descriptors
+[target.'cfg(any(unix, target_os = "redox"))'.dependencies]
 lazy_static = "1.3.0"
 
-[target.'cfg(target_os = "redox")'.dependencies]
-lazy_static = "1.3.0"
+# For caching result of CPUID check for RDRAND
+[target.'cfg(any(target_env = "sgx", target_os = "uefi"))'.dependencies]
+lazy_static = { version = "1.3.0", features = ["spin_no_std"] }
 
 [target.wasm32-unknown-unknown.dependencies]
 wasm-bindgen = { version = "0.2.29", optional = true }
 stdweb = { version = "0.4.9", optional = true }
 lazy_static = "1.3.0"
-
-[target.'cfg(any(target_env = "sgx", target_os = "uefi"))'.dependencies]
-lazy_static = { version = "1.3.0", features = ["spin_no_std"] }
 
 [features]
 std = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,9 +26,6 @@ libc = "0.2.54"
 [target.'cfg(unix)'.dependencies]
 lazy_static = "1.3.0"
 
-[target.'cfg(windows)'.dependencies]
-winapi = { version = "0.3.6", features = ["minwindef", "ntsecapi", "winnt"] }
-
 [target.'cfg(fuchsia)'.dependencies]
 fuchsia-cprng = "0.1"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,8 +20,10 @@ members = ["tests/wasm_bindgen"]
 [dependencies]
 log = { version = "0.4", optional = true }
 
+[target.'cfg(any(unix, target_os = "wasi"))'.dependencies]
+libc = "0.2.54"
+
 [target.'cfg(unix)'.dependencies]
-libc = "0.2.34"
 lazy_static = "1.3.0"
 
 [target.'cfg(windows)'.dependencies]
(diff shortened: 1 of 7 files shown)


newpavlov:
I also prefer to keep those `extern`'s inside `getrandom` instead of depending on platform crates, but @dhardy had a different opinion at the time.

dhardy:
I'm not too fussed; seems sensible enough to avoid the dependencies in this case.

newpavlov:
Can you fix the CI failure and rebase (otherwise right now PR contains formatting changes)?

josephlr:
> Can you fix the CI failure and rebase (otherwise right now PR contains formatting changes)?

Done

Cargo.toml
```diff
@@ -19,33 +19,20 @@ members = ["tests/wasm_bindgen"]
 
 [dependencies]
 log = { version = "0.4", optional = true }
+libc = "0.2.54"
```
newpavlov: I don't think we want to depend on `libc` for all targets. Maybe instead construct one big `cfg` expression with enumeration of targets which use `libc`?

src/windows.rs
```diff
@@ -12,14 +12,16 @@ extern crate std;
...
+    fn RtlGenRandom(RandomBuffer: *mut u8, RandomBufferLength: libc::c_ulong) -> u8;
+}
 
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
-    // Prevent overflow of ULONG
-    for chunk in dest.chunks_mut(ULONG::max_value() as usize) {
-        let ret = unsafe { RtlGenRandom(chunk.as_mut_ptr() as PVOID, chunk.len() as ULONG) };
+    // Prevent overflow of libc::c_ulong
+    for chunk in dest.chunks_mut(libc::c_ulong::max_value() as usize) {
+        let ret = unsafe { RtlGenRandom(chunk.as_mut_ptr(), chunk.len() as libc::c_ulong) };
```
newpavlov: Isn't `c_ulong` on Windows always equals to `u32`? I think we don't need the whole `libc` just for that.

Cargo.toml
josephlr: Done, I just used `unix || wasi`. Although that's not perfect, as we don't use libc on all unix targets (only 7 of them), but writing all that out makes the `Cargo.toml` way harder to read.

src/windows.rs
josephlr: That's [correct](https://docs.microsoft.com/en-us/cpp/cpp/data-type-ranges). Similarly on macOS/iOS, `SecRandomCopyBytes` is actually [defined to return `i32`](https://developer.apple.com/documentation/security/1399291-secrandomcopybytes), so I also removed that `libc` dependency on macOS/iOS.

Cargo.toml
newpavlov: We also could remove `libc` dependency for WASI as well, `libc` [code](https://github.com/rust-lang/libc/blob/0e702c1b4e2e56788e8b67d3efd9c242807c3d4b/src/wasi.rs#L1269) is quite straightforward. Not sure if it's worth though.

dhardy:
Can I leave this to you @newpavlov and @josephlr?

newpavlov:
Yes, I will do the merge after we will decide if we want to depend on `libc` for WASI target.

Cargo.toml
josephlr: We totally could (Strictly speaking we could do it for any libc declaration), but I think it’s best to not duplicate declarations from the libc, if possible.

josephlr: If we wanted we could not depend on the libc at all for any target, by copying over the 6 or 7 libc declarations needed to make everything link. But I think that would make breakage/instability more likely, not less.

newpavlov: It's just a bit strange for me to depend on `libc` crate for a target which, well, does not have `libc`.

josephlr: So I think they added it because [`wasi-libc` defines that symbol](https://github.com/CraneStation/wasi-libc/blob/320054e84f8f2440def3b1c8700cedb8fd697bf8/libc-bottom-half/headers/public/wasi/core.h#L708). However, it's just a macro that [links to the `wasi_unstable` module](https://github.com/CraneStation/wasi-libc/blob/94d34a13ad223930dce354ac33d3dfc73e9b5122/libc-bottom-half/headers/public/wasi/core.h#L451-L452). From their Github issues, it seems like the `wasi_unstable` module [is an implementation detail and may change in the future](https://github.com/CraneStation/wasmtime/issues/74#issuecomment-477779927). 

Given that, linking to the libc seems like the most stable approach, so we are not depending on an unstable components of the WASI API.

newpavlov: In my understanding we shouldn't care about `wasi-libc` at all. WASI defines [Core API](https://github.com/CraneStation/wasmtime/blob/master/docs/WASI-api.md) and [`__wasi_random_get`](https://github.com/CraneStation/wasmtime/blob/master/docs/WASI-api.md#random_get) is part of it. So we should be able to use it directly without any issues.

cc @alexcrichton @gnzlbg

PR #43, opened 2019-06-28
Check for AMD RDRAND bug

josephlr:
Pre-Ryzen AMD CPUs have a very serious bug where they will sometimes give `0xFFFFFFFF` as the output of RDRAND without setting the carry flag. This issue usually occurs when those CPU's awake from sleep: https://github.com/systemd/systemd/issues/11810#issuecomment-489727505

This CL simply implement's systemd's fix, where they [check for the bogus value and issue an error](https://github.com/systemd/systemd/blob/master/src/basic/random-util.c#L129-L138).

This PR also cleans up the `target_feature` checks. As we already fail to compile on `sgx` when `rdrand` is not enabled, the `is_rdrand_supported` function checks can just be `target_feature = "rdrand"`.

### DIFF
--- a/src/rdrand.rs
+++ b/src/rdrand.rs
@@ -26,7 +26,7 @@ unsafe fn rdrand() -> Result<[u8; WORD_SIZE], Error> {
             // AMD CPUs from families 14h to 16h (pre Ryzen) will sometimes give
             // bogus random data. Discard these values and warn the user.
             // See https://github.com/systemd/systemd/issues/11810#issuecomment-489727505
-            if el == 0 || el == !0 {
+            if cfg!(not(target_env = "sgx")) && (el == 0 || el == !0) {
                 error!("RDRAND returned suspicious value {}, CPU RNG is broken", el);
                 return Err(Error::UNKNOWN)
             }
--- a/src/rdrand.rs
+++ b/src/rdrand.rs
@@ -23,6 +23,13 @@ unsafe fn rdrand() -> Result<[u8; WORD_SIZE], Error> {
     for _ in 0..RETRY_LIMIT {
         let mut el = mem::uninitialized();
         if _rdrand64_step(&mut el) == 1 {
+            // AMD CPUs from families 14h to 16h (pre Ryzen) will sometimes give
+            // bogus random data. Discard these values and warn the user.
+            // See https://github.com/systemd/systemd/issues/11810#issuecomment-489727505
+            if el == 0 || el == !0 {
+                error!("RDRAND returned suspicious value {}, CPU RNG is broken", el);
+                return Err(Error::UNKNOWN)
+            }
             return Ok(el.to_ne_bytes());
         }
     }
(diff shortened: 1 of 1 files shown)


src/rdrand.rs
```diff
@@ -23,6 +23,13 @@ unsafe fn rdrand() -> Result<[u8; WORD_SIZE], Error> {
     for _ in 0..RETRY_LIMIT {
         let mut el = mem::uninitialized();
         if _rdrand64_step(&mut el) == 1 {
+            // AMD CPUs from families 14h to 16h (pre Ryzen) will sometimes give
+            // bogus random data. Discard these values and warn the user.
+            // See https://github.com/systemd/systemd/issues/11810#issuecomment-489727505
+            if el == 0 || el == !0 {
```
newpavlov: Here we essentially lose 1 bit of randomness out of 64 and introduce a potential point of failure even on bot affected CPUs (I know, chance is extremely small, but nevertheless). A better approach may be to [detect](https://github.com/systemd/systemd/issues/11810#issuecomment-490275562) faulty CPU using CPUID and store check result in an atomic flag, which will be used in this condition as `is_cpu_faulty && (el == 0 || el == !0)`. Also IIUC right now SGX is an Intel-only thing, so we can remove this check for it.

src/rdrand.rs
```diff
@@ -36,14 +43,14 @@ compile_error!(
     "SGX targets require 'rdrand' target feature. Enable by using -C target-feature=+rdrnd."
 );
 
-#[cfg(any(target_env = "sgx", target_feature = "rdrand"))]
+#[cfg(target_feature = "rdrand")]
```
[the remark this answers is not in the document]
josephlr: That's not what we're doing here.

On UEFI:
  - If `target_feature = "rdrand"` - No CPUID check
  - If `not(target_feature = "rdrand")` - CPUID check

On SGX:
  - If `target_feature = "rdrand"` - No CPUID check
  - If `not(target_feature = "rdrand")` - Compiler error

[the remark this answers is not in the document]
newpavlov: Ah, yes. Both variants are equivalent to each other. I've understood your `cfg`s incorrectly.

src/rdrand.rs
josephlr: So this doesn't make use lose 1 bit of randomness (that would be if we masked the high bit for example). It causes us to lose only 2 possibilities out of the 2^64 possibilities. So the `p(failure) = )(1/2)^63`, meaning that you would have to read 8 exabytes to hit this normally.

If you're concerned about this failure prob, we could just move the check inside the RDRAND loop, then this would never be hit on any sort of system ever.

newpavlov: Yes, you are right.

The check is already inside the loop and moving it outside will not change anything since we return from the loop using return result of `_rdrand64_step`.

So I think we just need modify this condition to ` cfg!(target_env = "sgx") && (el == 0 || el == !0)` and we are good to merge.

josephlr: Done

PR #45, opened 2019-06-28
Use shared File instead of shared FD.

### LINKED ISSUE #44: Shared file descriptor is closed prematurely
[Here](https://github.com/rust-random/getrandom/blob/master/src/use_file.rs#L38), we create a `File` from a raw fd. However, as this object is created on the stack, at the endo the function, the file will be dropped, and thus closed.

This means that on systems using a shared FD, the second call to `getrandom` will fail. This can be fixed by just storing a `File` instead of a `RawFd` in static memory.

josephlr:
Fixes #44 and removes use of `unsafe` code.

This can be implemented without `unsafe` because [`&'_ File` implements `Read`](https://doc.rust-lang.org/std/fs/struct.File.html#implementations), indicating that files can be read by multiple readers at the same time.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -74,6 +74,12 @@ impl From<NonZeroU32> for Error {
     }
 }
 
+impl From<&Error> for Error {
+    fn from(error: &Error) -> Self {
+        *error
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::Error;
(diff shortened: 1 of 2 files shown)


josephlr:
Tested on Linux with and without `getrandom` syscall enabled.

newpavlov:
TIL about `&File` implementing `Read`. :)

PR #46, opened 2019-06-29
macOS: Try getentropy() then fallback to /dev/random

### LINKED ISSUE #38: Use /dev/urandom on macOS
See this [comment](https://github.com/rust-lang/rust/pull/62082#issuecomment-505363176) by @RalfJung. It also should help with rust-random/rand#733.

Open question: do we need to read one byte from `/dev/random` to ensure that entropy pool has initialized?

cc @ebarnard

josephlr:
Fixes #38. As we moved files around `macos.rs -> ios.rs` the diff for this PR is weird. However, the diffs for each individual commit are fine, so review will be easier if you look at those.

As `/dev/random` and `/dev/urandom` are the same on macOS. We just use `/dev/random` to keep our `use_file` implementation simple.

Tested on macOS with and without `getentropy(2)`


newpavlov:
Can you also modify `lib.rs` docs?

josephlr:
> Can you also update `lib.rs` docs?

Done, I had to link to [unix.com's man pages](https://www.unix.com/man-page-mojave-repository.php) as there doesn't seem to be a good way to view macOS manpages on Apple's website.

newpavlov:
Great! Thank you!

PR #47, opened 2019-06-29
Prepare release v0.1.5

newpavlov:


josephlr:
Looks good to me, @newpavlov just needs to decide if we should detect the AMD RDRAND bug using the current implementation, or the implementation in #48.

We should release this before considering #49, as we want users to get all of the bug fixes without breaking backwards compatibility.

newpavlov:
Since FD bug is quite serious in my opinion I think I'll release v0.1.5 as-is and we can change RDRAND handling in v0.1.6.

PR #48, opened 2019-06-29
Alternative way to detect AMD bug

josephlr:
This is an alternative way to do the AMD bug detection "fixed" in #43.

@newpavlov feel free to immediately close this if you think this way is overly complex.

Before we checked the return value of RDRAND to detect the bug. However, this will occasionally return a false positive (every (1/2)^63 invocations), so we use the existing retry loop to try again. I also added comments explaining that:
  
  - The issue here is AMD not setting the CF flag properly (as RDRAND is allowed to fail)
  - Why we perform this check on all targets.

### DIFF
--- a/src/rdrand.rs
+++ b/src/rdrand.rs
@@ -27,7 +27,7 @@ unsafe fn rdrand() -> Result<[u8; WORD_SIZE], Error> {
             // set CF on bogus random data, so we check these values explictly.
             // See https://github.com/systemd/systemd/issues/11810#issuecomment-489727505
             // We perform this check regardless of target to guard against
-            // future implementations that incorrectly set CF.
+            // any implementation that incorrectly fails to set CF.
             if el != 0 && el != !0 {
                 return Ok(el.to_ne_bytes());
             }
--- a/src/rdrand.rs
+++ b/src/rdrand.rs
@@ -23,14 +23,16 @@ unsafe fn rdrand() -> Result<[u8; WORD_SIZE], Error> {
     for _ in 0..RETRY_LIMIT {
         let mut el = mem::uninitialized();
         if _rdrand64_step(&mut el) == 1 {
-            // AMD CPUs from families 14h to 16h (pre Ryzen) will sometimes give
-            // bogus random data. Discard these values and warn the user.
+            // AMD CPUs from families 14h to 16h (pre Ryzen) sometimes fail to
+            // set CF on bogus random data, so we check these values explictly.
             // See https://github.com/systemd/systemd/issues/11810#issuecomment-489727505
-            if cfg!(not(target_env = "sgx")) && (el == 0 || el == !0) {
-                error!("RDRAND returned suspicious value {}, CPU RNG is broken", el);
-                return Err(Error::UNKNOWN);
+            // We perform this check regardless of target to guard against
+            // future implementations that incorrectly set CF.
+            if el != 0 && el != !0 {
+                return Ok(el.to_ne_bytes());
             }
-            return Ok(el.to_ne_bytes());
+            error!("RDRAND returned {:X}, CPU RNG may be broken", el);
+            // Keep looping in case this was a false positive.
         }
     }
     error!("RDRAND failed, CPU issue likely");


dhardy:
So in practical terms, this change means that anyone with an affected CPU cannot use the RDRAND implementation at all. Is the presence of a sporadic bug (on resume from standby) sufficient justification to not trust RDRAND on those CPUs at all, while still trusting it on other CPUs? @tarcieri thoughts? I haven't read up on the issue, but this seems unnecessary.

Question: is there a reason we can't discard "obviously bad" values and try again? There are three aspects to this question:

-   Can the CPU "recover" and return other values?
-   In this case, could the occurrence of the bug reduce security of the result?
-   What is the expected number of false-positives of the guard removed by this PR? (2^-63 times the number of uses of RDRAND via this lib — so unless servers actually make millions of requests the number of false positives is likely to be 0 or *very* close.)

There is also the question: is it likely other (possibly future) CPUs might exhibit a similar bug?

---

My view is that the most robust solution would be to trap the `!0` value, but retry a couple of times (we're already in a loop, so `continue` is sufficient).

src/rdrand.rs
```diff
@@ -43,25 +36,53 @@ compile_error!(
...
     *HAS_RDRAND
 }
 
+#[cfg(not(target_env = "sgx"))]
+pub fn has_rdrand() -> bool {
+    use core::arch::x86_64::__cpuid;
+    // SAFETY: All x86_64 CPUs support CPUID leaf 0 and 1
+    let leaf_0 = unsafe { __cpuid(0) };
+    let vendor_id: [u8; 12] = unsafe { mem::transmute([leaf_0.ebx, leaf_0.edx, leaf_0.ecx]) };
+    let is_amd = &vendor_id == b"AuthenticAMD";
```
newpavlov: You can write this part slightly more efficiently:
```rust
let leaf0 = unsafe { __cpuid(0) };
let amd: &[u8; 12] = b"AuthenticAMD";
let r: [u32; 3] = unsafe { mem::transmute_copy(amd) };
leaf0.ebx == r[0] && leaf0.edx == r[1] && leaf0.ecx == r[2]
```

josephlr: I just got rid of this vendor ID check entirely.

src/rdrand.rs
```diff
@@ -23,14 +23,16 @@ unsafe fn rdrand() -> Result<[u8; WORD_SIZE], Error> {
...
-            // bogus random data. Discard these values and warn the user.
+            // AMD CPUs from families 14h to 16h (pre Ryzen) sometimes fail to
+            // set CF on bogus random data, so we check these values explictly.
             // See https://github.com/systemd/systemd/issues/11810#issuecomment-489727505
-            if cfg!(not(target_env = "sgx")) && (el == 0 || el == !0) {
-                error!("RDRAND returned suspicious value {}, CPU RNG is broken", el);
-                return Err(Error::UNKNOWN);
+            // We perform this check regardless of target to guard against
+            // future implementations that incorrectly set CF.
+            if el != 0 && el != !0 {
```
newpavlov: You've removed `cfg!(not(target_env = "sgx"))`. Was it intentional?

josephlr:
I updated this PR and the description to incorporate the above suggestions

> I think disabling RDRAND completely on affected CPUs is too extreme of a measure. Especially considering that failure can be reliably detected.

I agree. Checking the output value and trying again seems avoid false positives and being overly aggressive.

> Personally I like approach of detecting CPU family and doing value check only for affected CPUs. I think we can write it like this (note the changed rdrand signature). But I understand the complexity concern, so @dhardy's proposal is a viable alternative.

The currently version has the check in place for all targets. This bug could be present in the future and performing the check is very cheep (just a [`lea` and `cmp` for both checks](https://rust.godbolt.org/z/lr4rr_))

> There is also the question: is it likely other (possibly future) CPUs might exhibit a similar bug?

This is a good point. [Intel's own documentation](https://software.intel.com/en-us/blogs/2014/01/13/rdrand-do-i-need-to-check-the-carry-flag-or-can-i-just-check-for-zero) says that 0x00000000 is returned when RDRAND fails. Now they set the CF flag correctly. However, failing to set a CF flag is a relatively easy failure mode, so I think keeping this check for all implementations is reasonable.

src/rdrand.rs
josephlr: Yes, see my comment below

PR #51, opened 2019-07-01
Add Structures for lock-free initialization

### LINKED ISSUE #50: Don't use spin_no_std feature
See this issue: rust-lang-nursery/lazy-static.rs#150

I think we can relatively easily replace code [here](https://github.com/rust-random/getrandom/blob/master/src/rdrand.rs#L61) with something like this:
```rust

fn is_rdrand_supported() -> bool {
    use core::arch::x86_64::__cpuid;
    use std::sync::atomic::{AtomicUsize, Ordering};
    
    static RDRAND: AtomicUsize = AtomicUsize::new(0);
    const RDRAND_FLAG: u32 = 1 << 30;
    const STATE_FLAG: usize = 1 << 1;
    const RESULT_FLAG: usize = 1 << 0;

    let flags = RDRAND.load(Ordering::Relaxed);
    if (flags & STATE_FLAG) == 0 {
        let leaf1_ecx = unsafe { __cpuid(1).ecx };
        if leaf1_ecx & RDRAND_FLAG != 0 {
            // rdrand is supported
            RDRAND.store(RESULT_FLAG | STATE_FLAG, Ordering::Release);
            true
        } else {
            // rdrand is not supported
            RDRAND.store(STATE_FLAG, Ordering::Release);
            false
        }
    } else {
        (flags & RESULT_FLAG) != 0
    }
}
```
In the worst-case scenario we may call CPUID several times instead of just one, but I don't think it's a big deal. Also it will remove `spin` crate from our dependency tree, which will help a bit with `std` inclusion.

cc @josephlr

josephlr:
Fixes #50 

This is the first step to improving the synchronization concerns in `getrandom`. Currently we need synchronization to:
  
  1. Hold a global file descriptor for `/dev/urandom` or `/dev/random`
  2. "Weak linkage" of a libc symbol (`getrandom` on solaris or `getentropy` on macos).
  3. Detecting RDRAND
  4. Detecting `getrandom(2)` on Linux.
  5. Determining the environment on `wasm32-unknown-unknown`.

Right now we use `lazy_static` for all of these, but that isn't ideal if we want `getrandom` to be used to implement `libstd`. To address some of these concerns, we add a `LazyUsize` type that just wraps an `AtomicUsize`, and provides methods for lazy initialization, where it is OK for the initialization to be run multiple times. We then build two structs on top of `LazyUsize`:

  1. `LazyBool` - This is just like `LazyUsize` except with `bool` types in the function signatures. It is used to replace `lazy_static` for (3) and (4).
  2. `Weak` - Based on the [`libstd` struct of the same name](https://github.com/rust-lang/rust/blob/9ebf47851a357faa4cd97f4b1dc7835f6376e639/src/libstd/sys/unix/weak.rs), this allows us to "link" to system `libc` functions that may or may not be present at runtime. This replaces `lazy_static` for (2). It also simplifies some duplicated logic, avoids locking, and makes implementing #35 easier. 

The remaining uses of `lazy_static` can be eliminated by just using a combination of `std::sync::Once` and `static mut` in a follow up PR (see #52).


josephlr:
This is based on @newpavlov's idea here: https://github.com/rust-random/getrandom/issues/50#issue-462416233, it just uses `AtomicBool` instead of `AtomicUsize`,  and encapsulates the complexity in a struct in a separate file.

src/util.rs
```diff
@@ -0,0 +1,24 @@
...
+        LazyBool {
+            value: AtomicBool::new(false),
+            needs_init: AtomicBool::new(true),
+        }
+    }
+
+    pub fn unsync_init(&self, init: impl FnOnce() -> bool) -> bool {
+        if self.needs_init.load(Ordering::SeqCst) {
+            self.value.store(init(), Ordering::SeqCst);
+            self.needs_init.store(false, Ordering::SeqCst);
```
dhardy: This isn't a "lazy bool" the way one may expect from `lazy_static`: it may run `init()` multiple times. Do we care about this, e.g. if someone spins up a lot of threads which each use `getrandom` simultaneously? See https://github.com/rust-random/rand/issues/804

josephlr: The description mentions this a little, but here we shouldn't care. The call to check `getrandom(2)` support is non-blocking and requests no randomness. So it would be about as quick as any other linux syscall that doesn't really do anything. And probably quicker than waiting on a kernel lock. 

Should the name be changed?

newpavlov:
Why do we need to bother with the `LazyBool` abstraction? I think it will be fine to simply use code in the issue, it's more efficient and not that complex in comparison. (ideally it would use `AtomicU8`, but it will require bumping MSRV for UEFI and SGX targets)

josephlr:
> Why do we need to bother with the LazyBool abstraction?

Having the abstraction is crucial if we want to use this logic in multiple places (like this PR does).

src/util.rs
```diff
@@ -0,0 +1,24 @@
+use core::sync::atomic::{AtomicBool, Ordering};
+
+#[derive(Debug)]
+pub struct LazyBool {
+    value: AtomicBool,
+    needs_init: AtomicBool,
```
newpavlov: I think using one atomic and bit flags will be more efficient than two `AtomicBool`s. Plus you will be able to relax orderings in `usync_init`.

src/util.rs
josephlr: I added some comments to better explain this.

src/util.rs
josephlr: Done, now both `LazyBool` and `Weak` use `LazyUsize`. `LazyUsize` itself just uses an `AtomicUsize` with `Ordering::Relaxed`.

josephlr:
I've update the description for this. Now there are two lock-free helpers `LazyBool` and `Weak`, both using an `AtomicUsize` to implement the lazy initialization.

src/util.rs
```diff
@@ -0,0 +1,45 @@
...
+        Self(AtomicUsize::new(usize::max_value()))
+    }
+
+    // Runs the init() function at least once, returning the value of some run
+    // of init(). Unlike std::sync::Once, the init() function may be run
+    // multiple times. If init() returns usize::max_value(), the init() function
+    // will always be retried on a future call to unsync_init(). This makes it
+    // ideal for representing failure.
+    pub fn unsync_init(&self, init: impl FnOnce() -> usize) -> usize {
+        // Relaxed ordering is fine, as init() is allowed to run mulitple times.
```
dhardy: Wrong explanation. Probably you know, but the reason is that we only have a single shared variable.

dhardy: https://llvm.org/docs/Atomics.html#atomic-orderings

josephlr: Oh duh, I was overthinking my explanation. Fixed. 

I also renamed the `lazy` module to be `util_libc` as I'm anticipating we will need more `libc` based utility functions.

src/util.rs
```diff
@@ -0,0 +1,45 @@
...
+    // of init(). Unlike std::sync::Once, the init() function may be run
+    // multiple times. If init() returns usize::max_value(), the init() function
+    // will always be retried on a future call to unsync_init(). This makes it
+    // ideal for representing failure.
+    pub fn unsync_init(&self, init: impl FnOnce() -> usize) -> usize {
+        // Relaxed ordering is fine, as we only have a single atomic variable.
+        if self.0.load(Ordering::Relaxed) == usize::max_value() {
+            self.0.store(init(), Ordering::Relaxed)
+        }
+        self.0.load(Ordering::Relaxed)
```
newpavlov: Why do you use two loads here? The same ca be written as:
```rust
let mut val = self.0.load(Ordering::Relaxed);
if val == usize::max_value() {
    val = init();
    self.0.store(val, Ordering::Relaxed);
}
val
```

src/util_libc.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    // Returns the function pointer if it is present at runtime. Otherwise,
+    // return None, indicating the function does not exist.
+    pub fn func(&self) -> Option<F> {
+        assert_eq!(mem::size_of::<Option<F>>(), mem::size_of::<usize>());
+
+        let addr = self.addr.unsync_init(|| unsafe {
+            libc::dlsym(libc::RTLD_DEFAULT, self.name.as_ptr() as *const _) as usize
+        });
+        unsafe { mem::transmute_copy(&addr) }
```
newpavlov: Strictly speaking this code is unsafe since we do not check anything except size, so it's possible to create an invalid value. I think a better approach will be to drop `F` type parameter and return `*mut c_void` from here and handle transformation to function pointer in user code.

src/util.rs
josephlr: I was just about to submit this, good catch. Fixed.

src/util_libc.rs
josephlr: Good idea.

I opted for `Option<NonNull<libc::c_void>>` to convey the success/failure as well as the fact that we're dealing with a pointer. Although the code isn't any different than using `Option<NonZeroUsize>` if you think that's cleaner.

PR #52, opened 2019-07-01
Remove lazy_static dependancy

josephlr:
EDIT1: We make the following changes:
 - Replace `wasm32-stdweb`'s use of `lazy_static` with `static mut` and `std::sync::Once`
 - Add a `sync_init` method to `LazyUsize` complementing `unsync_init`.
 - Add a `LazyFd` abstraction around `LazyUsize` to work with `Option<i32>` values, were we either have `None` or `Some(x)` with x non-negative.
 - Use the `LazyFd` in `use_file`, eliminating our last use of `lazy_static`.

EDIT2: Note that this does not try to remove all `std` dependencies from `use_file.rs`. That is for a followup CL.

### DIFF
--- a/src/use_file.rs
+++ b/src/use_file.rs
@@ -14,7 +14,10 @@ use crate::Error;
 use core::mem::ManuallyDrop;
 use core::num::NonZeroU32;
 use std::os::unix::io::{FromRawFd, IntoRawFd, RawFd};
-use std::{fs::File, io::Read};
+use std::{
+    fs::File,
+    io::{self, Read},
+};
 
 #[cfg(target_os = "redox")]
 const FILE_PATH: &str = "rand:";
@@ -32,7 +35,7 @@ const FILE_PATH: &str = "/dev/random";
 
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
     static FD: LazyFd = LazyFd::new();
-    let fd = FD.init(init_file).ok_or(Error::UNKNOWN)?;
+    let fd = FD.init(init_file).ok_or(io::Error::last_os_error())?;
     let file = ManuallyDrop::new(unsafe { File::from_raw_fd(fd) });
     let mut file_ref: &File = &file;
 
--- a/src/use_file.rs
+++ b/src/use_file.rs
@@ -9,9 +9,11 @@
 //! Implementations that just need to read from a file
 extern crate std;
 
+use crate::util_libc::LazyFd;
 use crate::Error;
+use core::mem::ManuallyDrop;
 use core::num::NonZeroU32;
-use lazy_static::lazy_static;
+use std::os::unix::io::{FromRawFd, IntoRawFd, RawFd};
 use std::{fs::File, io::Read};
 
 #[cfg(target_os = "redox")]
@@ -29,28 +31,31 @@ const FILE_PATH: &str = "/dev/urandom";
 const FILE_PATH: &str = "/dev/random";
 
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
-    lazy_static! {
-        static ref FILE: Result<File, Error> = init_file();
-    }
-    let mut f = FILE.as_ref()?;
+    static FD: LazyFd = LazyFd::new();
+    let fd = FD.init(init_file).ok_or(Error::UNKNOWN)?;
+    let file = ManuallyDrop::new(unsafe { File::from_raw_fd(fd) });
+    let mut file_ref: &File = &file;
 
     if cfg!(target_os = "emscripten") {
         // `Crypto.getRandomValues` documents `dest` should be at most 65536 bytes.
         for chunk in dest.chunks_mut(65536) {
-            f.read_exact(chunk)?;
+            file_ref.read_exact(chunk)?;
         }
     } else {
-        f.read_exact(dest)?;
+        file_ref.read_exact(dest)?;
     }
     Ok(())
 }
 
-fn init_file() -> Result<File, Error> {
+fn init_file() -> Option<RawFd> {
     if FILE_PATH == "/dev/urandom" {
         // read one byte from "/dev/random" to ensure that OS RNG has initialized
-        File::open("/dev/random")?.read_exact(&mut [0u8; 1])?;
+        File::open("/dev/random")
+            .ok()?
+            .read_exact(&mut [0u8; 1])
+            .ok()?;
     }
-    Ok(File::open(FILE_PATH)?)
+    Some(File::open(FILE_PATH).ok()?.into_raw_fd())
 }
 
 #[inline(always)]
--- a/src/util.rs
+++ b/src/util.rs
@@ -6,10 +6,26 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use core::sync::atomic::{AtomicUsize, Ordering};
+use core::sync::atomic::{AtomicUsize, Ordering::Relaxed};
 
 // This structure represents a laziliy initialized static usize value. Useful
 // when it is perferable to just rerun initialization instead of locking.
+// Both unsync_init and sync_init will invoke an init() function until it
+// succeeds, then return the cached value for future calls.
+//
+// Both methods support init() "failing". If the init() method returns UNINIT,
+// that value will be returned as normal, but will not be cached.
+//
+// Users should only depend on the _value_ returned by init() functions.
+// Specifically, for the following init() function:
+//      fn init() -> usize {
+//          a();
+//          let v = b();
+//          c();
+//          v
+//      }
+// the effects of c() or writes to shared memory will not necessarily be
+// observed and additional syncronization methods with be needed.
 pub struct LazyUsize(AtomicUsize);
 
 impl LazyUsize {
@@ -19,19 +35,43 @@ impl LazyUsize {
 
     // The initialization is not completed.
     pub const UNINIT: usize = usize::max_value();
+    // The initialization is currently running.
+    pub const ACTIVE: usize = usize::max_value() - 1;
 
     // Runs the init() function at least once, returning the value of some run
-    // of init(). Unlike std::sync::Once, the init() function may be run
-    // multiple times. If init() returns UNINIT, future calls to unsync_init()
-    // will always retry. This makes UNINIT ideal for representing failure.
+    // of init(). Multiple callers can run their init() functions in parallel.
+    // init() should always return the same value, if it succeeds.
     pub fn unsync_init(&self, init: impl FnOnce() -> usize) -> usize {
         // Relaxed ordering is fine, as we only have a single atomic variable.
-        let mut val = self.0.load(Ordering::Relaxed);
+        let mut val = self.0.load(Relaxed);
         if val == Self::UNINIT {
             val = init();
-            self.0.store(val, Ordering::Relaxed);
+            self.0.store(val, Relaxed);
         }
         val
+    }
+
+    // Synchronously runs the init() function. Only one caller will have their
+    // init() function running at a time, and exactly one successful call will
+    // be run. The init() function should never return LazyUsize::ACTIVE.
+    pub fn sync_init(&self, init: impl FnOnce() -> usize, mut wait: impl FnMut()) -> usize {
+        // Common and fast path with no contention. Don't wast time on CAS.
+        match self.0.load(Relaxed) {
+            Self::UNINIT | Self::ACTIVE => {}
+            val => return val,
+        }
+        // Relaxed ordering is fine, as we only have a single atomic variable.
+        loop {
+            match self.0.compare_and_swap(Self::UNINIT, Self::ACTIVE, Relaxed) {
+                Self::UNINIT => {
+                    let val = init();
+                    self.0.store(val, Relaxed);
+                    return val;
+                }
+                Self::ACTIVE => wait(),
+                val => return val,
+            }
+        }
     }
 }
 
--- a/src/util_libc.rs
+++ b/src/util_libc.rs
@@ -36,3 +36,29 @@ impl Weak {
         NonNull::new(addr as *mut _)
     }
 }
+
+pub struct LazyFd(LazyUsize);
+
+impl LazyFd {
+    pub const fn new() -> Self {
+        Self(LazyUsize::new())
+    }
+
+    // If init() returns Some(x), x should be nonnegative.
+    pub fn init(&self, init: impl FnOnce() -> Option<libc::c_int>) -> Option<libc::c_int> {
+        let fd = self.0.sync_init(
+            || match init() {
+                // OK as val >= 0 and val <= c_int::MAX < usize::MAX
+                Some(val) => val as usize,
+                None => LazyUsize::UNINIT,
+            },
+            || unsafe {
+                libc::usleep(1000);
+            },
+        );
+        match fd {
+            LazyUsize::UNINIT => None,
+            val => Some(val as libc::c_int),
+        }
+    }
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,13 +102,14 @@ matrix:
 
     - rust: nightly
       env: DESCRIPTION="cross-platform build only"
+      # Redox: wait for https://github.com/rust-lang/rust/issues/60139
       install:
         - rustup target add x86_64-sun-solaris
         - rustup target add x86_64-unknown-cloudabi
         - rustup target add x86_64-unknown-freebsd
         - rustup target add x86_64-fuchsia
         - rustup target add x86_64-unknown-netbsd
-        - rustup target add x86_64-unknown-redox
+        # - rustup target add x86_64-unknown-redox
         - rustup target add x86_64-fortanix-unknown-sgx
         # For no_std targets
         - rustup component add rust-src
@@ -119,7 +120,7 @@ matrix:
         - cargo build --target=x86_64-unknown-freebsd --all-features
         - cargo build --target=x86_64-fuchsia --all-features
         - cargo build --target=x86_64-unknown-netbsd --all-features
-        - cargo build --target=x86_64-unknown-redox --all-features
+        # - cargo build --target=x86_64-unknown-redox --all-features
         - cargo build --target=x86_64-fortanix-unknown-sgx --all-features
         - cargo xbuild --target=x86_64-unknown-uefi
         # also test minimum dependency versions are usable
@@ -129,7 +130,7 @@ matrix:
         - cargo build --target=x86_64-unknown-freebsd --all-features
         - cargo build --target=x86_64-fuchsia --all-features
         - cargo build --target=x86_64-unknown-netbsd --all-features
-        - cargo build --target=x86_64-unknown-redox --all-features
+        # - cargo build --target=x86_64-unknown-redox --all-features
         - cargo build --target=x86_64-fortanix-unknown-sgx --all-features
         - cargo xbuild --target=x86_64-unknown-uefi
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,6 @@ lazy_static = "1.3.0"
 [target.wasm32-unknown-unknown.dependencies]
 wasm-bindgen = { version = "0.2.29", optional = true }
 stdweb = { version = "0.4.9", optional = true }
-lazy_static = "1.3.0"
 
 [features]
 std = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,10 +23,6 @@ log = { version = "0.4", optional = true }
 [target.'cfg(any(unix, target_os = "redox", target_os = "wasi"))'.dependencies]
 libc = "0.2.54"
 
-# For holding file descriptors
-[target.'cfg(any(unix, target_os = "redox"))'.dependencies]
-lazy_static = "1.3.0"
-
 [target.wasm32-unknown-unknown.dependencies]
 wasm-bindgen = { version = "0.2.29", optional = true }
 stdweb = { version = "0.4.9", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ members = ["tests/wasm_bindgen"]
 [dependencies]
 log = { version = "0.4", optional = true }
 
-[target.'cfg(any(unix, target_os = "wasi"))'.dependencies]
+[target.'cfg(any(unix, target_os = "redox", target_os = "wasi"))'.dependencies]
 libc = "0.2.54"
 
 # For holding file descriptors
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -142,11 +142,11 @@ extern crate std;
 
 mod error;
 pub use crate::error::Error;
+
 #[allow(dead_code)]
 mod util;
-
-// These targets need weak linkage to libc randomness functions.
-#[cfg(any(target_os = "macos", target_os = "solaris", target_os = "illumos"))]
+#[cfg(any(unix, target_os = "redox"))]
+#[allow(dead_code)]
 mod util_libc;
 
 // System-specific implementations.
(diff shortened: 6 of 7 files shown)


Centril:
cc @RalfJung, @eddyb -- can you please have a look at this in terms of `static mut`? (https://github.com/rust-lang/rust/pull/62082 wants to use `getrandom` in libstd)

Centril:
also cc @alexcrichton

src/use_file.rs
```diff
@@ -29,10 +28,12 @@ const FILE_PATH: &str = "/dev/urandom";
 const FILE_PATH: &str = "/dev/random";
 
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
-    lazy_static! {
-        static ref FILE: Result<File, Error> = init_file();
-    }
-    let mut f = FILE.as_ref()?;
+    static ONCE: Once = Once::new();
+    static mut FILE: Option<Result<File, Error>> = None;
```
newpavlov: Why not return to storing file descriptor in `AtomicIsize` like I was doing in #14? AFAIK negative values are invalid for all supported targets, so we could use the highest bit as a flag "initialization is not finished" and the lowest bit will be "initialization is ongoing".

josephlr: Given that we shouldn't be depending on `std` at all. You're correct, we would need to do something like that. Let me take another look.

RalfJung: Or you just use an `AtomicBool` for the synchronization, then you don't have to do bit fiddling. ;)

dhardy: However in that case you have multiple shared vars and can't use relaxed loads, so is it worth it?

RalfJung: Also, how do you plan to open a file without depending on `std`?
(Seems like a separate issue though from not using `std` for synchronization.)

RalfJung: > However in that case you have multiple shared vars and can't use relaxed loads, so is it worth it?

Not sure what you mean? `AtomicBool` doesn't work, I'm sorry, you need three states (unless your initialziation is idempotent, which it probably is not when file descriptiors are involved).

Then you only need atomic accesses (release/acquire) for the atomic flag, everything else can be non-atomic.

RalfJung: So, roughly, the algorithm is something like
```rust
const UNINIT: u8 = 0;
const PENDING: u8 = 1;
const DONE: u8 = 2;
static FLAG: AtomicU8 = AtomicU8::new(UNINIT);
static mut DATA: T := <whatever>;

fn init_data() -> T {
  // the lazy init code.
}

fn get_data() -> T {
  loop {
    let flag = FLAG.load(Ordering::Acquire);
    if flag == DONE {
      // Safe because: We know the data is initialized and we did an acquire so
      // we synchronized with the initialization.
      return unsafe { DATA }
    }
    // Let's try to initialize.
    // This could likely be "Relaxed", but let's play safe and use AcqRel.
    if flag == UNINIT && FLAG.compare_and_swap(UNINIT, PENDING, Ordering::AcqRel) == UNINIT {
      // We won the race to initialize!
      let data = init_data();
      // Safe because: we "grabbed the lock" by successfully CASing from UNINIT to PENDING.
      unsafe { DATA = data; }
      // Now signal everyone else that the DATA is ready. Must be "release";
      // this synchronizes with the "acquire" load above.
      FLAG.store(DONE, Ordering::Release);
      return data;
    }
    // Just try again, spinning until we see DONE.
  }
}
```

newpavlov: Yes, we have 3 states, so single `AtomicBool` will not work. Also the idea was to store initialization state and file descriptor in a single atomic, so code will use just a single memory load.

BTW if we are aiming to drop `std` we can not use `Once`, so we have to return to custom spinlock from #14. To prevent wastefully burning CPU we can use `libc::nanosleep`.

RalfJung: Alternatively, you could look into using https://docs.rs/spin/0.5.0/spin/struct.Once.html directly? You know what code you are going to put into that initializer, so you know if it's fine to spin-wait on that. Seems better than open-coding the same thing. ;)

However, you are putting file system code into there, right? To open `/dev/random` or so? (a) how does that make any sense with nostd and (b) that's really not something you want to do with a spin lock. You might be waiting for spinning metal here. That takes *ages*.

newpavlov: Nothing prevents us from using system capabilities in a `no_std` crate via `libc`.

RalfJung: > Also the idea was to store initialization state and file descriptor in a single atomic, so code will use just a single memory load.

Does this actually help with performance, or does it just make the code more complicated to think about by mixing up (synchronization) control and data?

> BTW if we are aiming to drop std we can not use Once, so we have to return to custom spinlock from #14. To prevent wastefull burning of CPU we can use libc::nanosleep.

Another one of these cyclic dependencies with libstd. :/ `parking_lot` has that problem, too.

RalfJung: > Nothing prevents us from using system capabilities in a no_std crate via libc.

It's super annoying though to not be able to use `File`. :/

newpavlov: On a second thought keeping all state in a single atomic will not work, `fetch_or` in the initialization loop may corrupt file descriptor value. So we return to this [code](https://github.com/newpavlov/getrandom/blob/6ab631c27646b815beb709c6f7b196563094e5cc/src/use_file.rs) (but we can relax some orderings, see comments in the PR). But I think it's worth to keep `-1` as an initialization value for file descriptor atomic to prevent wasteful memory reads.

>It's super annoying though to not be able to use File

Yeah, but luckily we only need `open` and `read`, though duplicating read loop will be indeed annoying.

dhardy: The example @RalfJung posted above uses more precise synchronisation ([this article](https://preshing.com/20120913/acquire-and-release-semantics/) gives a better description of the locks required). Though as noted we will want some type of `thread::yield` or sleep when waiting.

josephlr: > The example @RalfJung posted above uses more precise synchronisation ([this article](https://preshing.com/20120913/acquire-and-release-semantics/) gives a better description of the locks required). Though as noted we will want some type of `thread::yield` or sleep when waiting.

[`libc::sched_yield`](https://docs.rs/libc/0.2.58/libc/fn.sched_yield.html) is what we want here.

> It's super annoying though to not be able to use File. :/

Agree 100%. However, we only need to do this complex stuff (both opening files, and non-trivial syncronization) on `unix` targets, so we can just use `libc::open`, `libc::read`, and `libc::close`. It will be like writing C code (not exactly a good thing, but it will work).

> On a second thought keeping all state in a single atomic will not work

So I think I have an idea that will allow us to keep using a single value, and not make the code too terrible, I'll update this CL with my proposal.

newpavlov: >libc::sched_yield is what we want here.

Don't forget about rust-lang/rust#46774. We have discussed it in your initial lazy_static PR, I think sleeping short interval will be a better option in our case.

RalfJung: AFAIK `parking_lot` has some fancy back-off logic that would, in principle, be useful here. @Amanieu is that available as a separate crate or so?

However, this is also not very performance sensitive, as it can only happen a few times and will never happen once initialization is complete, so `sleep_ms(1)` is likely good enough.

josephlr: > Don't forget about [rust-lang/rust#46774](https://github.com/rust-lang/rust/issues/46774).

Whoops I had forgotten, and sleeping for 1 ms sounds like the best bet. Ideally (at least in my opinion) `parking_lot` would just be the crate that implements all the synchronization primitives for  the stdlib, making this all a lot easier. But until then, I think we will have to roll our own less good versions.

RalfJung: It [will be](https://github.com/rust-lang/rust/pull/56410). But even then it will not be possible for a `no_std` crate to use it as there is a cyclic dependency between libstd and parking_lot.

I suggest we don't add `getrandom` to that cycle. Things are complicated enough already.^^

josephlr:
Here's my basic idea for adding a one-time synchronization method to `LazyUsize`.

```rust
// Runs the init() function exactly once.
pub fn sync_init(&self, init: impl FnOnce() -> usize, mut wait: impl FnMut()) -> usize {
    loop {
        match self.0.compare_and_swap(Self::UNINIT, Self::ACTIVE, Relaxed) {
            Self::UNINIT => {
                let val = init();
                self.0.store(val, Relaxed);
                return val;
            }
            Self::ACTIVE => wait(),
            val => return val,
        }
    }
}
```
Where `wait` would be `sleep_ms(1)` or something, and:
```rust
// The initialization is not completed.
pub const UNINIT: usize = usize::max_value();
// The initialization is currently running.
pub const ACTIVE: usize = usize::max_value() - 1;
```

Then we can just use the fact that non-zero [RawFd](https://doc.rust-lang.org/std/os/unix/io/type.RawFd.html)'s will fit in the range `[0, usize::max_value() - 2]`.

newpavlov:
Looks good! Small nitpicks:
- I think we can use `impl Fn()` for `wait` and `init` functions.
- I am not sure about `Relaxed` ordering. Especially considering that you will use it together with `LazyBool`/`LazyUsize` on some targets. @HadrienG2 what do you think?
- Zero is a valid value for file descriptor. It does not matter for the provided snippet, but you've used "non-zero RawFd's" in your comment.

HadrienG2:
@newpavlov Relaxed is correct for synchronizing accesses to a single memory location, which I understand is what you are doing here.

If you want to synchronize accesses to multiple memory locations (e.g. writing data to a static mut and using a shared flag to tell other threads when it's been committed), then you need stronger memory orderings like Acquire/Release pairs.

josephlr:
> * I think we can use `impl Fn()` for `wait` and `init` functions.

Isn’t `FnMut` a weaker trait bound than `Fn`?

> * I am not sure about `Relaxed` ordering. Especially considering that you will use it together with `LazyBool`/`LazyUsize` on some targets. @HadrienG2 what do you think?

Does this matter if we’re only using a single atomic value? 

> * Zero is a valid value for file descriptor. It does not matter for the provided snippet, but you've used "non-zero RawFd's" in your comment.

Typo, now fixed.

newpavlov:
>Isn’t FnMut a weaker trait bound than Fn?

Yes, but our callbacks will not contain any state, so stronger bound will work as well. It's just my preference, you can leave it as-is.

>Does this matter if we’re only using a single atomic value?

On some targets you use it together with `LazyBool`/`LazyUsize` which also use relaxed load/store under the hood. But I guess since use of the second atomic resides behind branch execution of which is explicitly dependent on a certain non-default value of the first atomic it should not create any problems.

dhardy:
Good point. I think on Linux we could just use a single `AtomicUsize` if we allow *another* special value to imply "use getrandom syscall", but on Solaris I believe we will need either multiple atomic vars or one "gate" atomic plus one or more data statics.

HadrienG2:
> > Does this matter if we’re only using a single atomic value?
> 
> On some targets you use it together with `LazyBool`/`LazyUsize` which also use relaxed load/store under the hood. But I guess since use of the second atomic resides behind branch execution of which is explicitly dependent on a certain non-default value of the first atomic it should not create any problems.

This can cause problems :

```
if a.load(Ordering::Relaxed) == SOMETHING {
    return b.load(Ordering::Relaxed);
}
```

Because the compiler and CPU can turn it into this:

```
let tmp = b.load(Ordering::Relaxed);
if a.load(Ordering::Relaxed) == SOMETHING {
    return tmp;
}
```

In that case, you need Acquire when loading from `a` on the consumer side and Release when writing to it on the producer side.
```

**Comment by dhardy** [comment]
Thanks @HadrienG2 — I believe [this example](https://github.com/rust-random/getrandom/pull/52#discussion_r299363108) posted earlier is correct however.

**Comment by HadrienG2** [comment]
@dhardy Yup, it uses stronger Acquire/Release memory orderings to avoid the aforementioned reordering effects.

**Comment by newpavlov** [comment]
@HadrienG2 
Our code can be represented as:
```rust
// `a` is initialized with C1, and `b` with `C2`
let val = match a.load(Relaxed) {
    C1 => {
        let val = init1();
        a.store(val, Relaxed);
        val
    }
    val => val,
};

if val == 1 {
    do_stuff1();
} else {
    loop {
        let val = match b.compare_and_swap(C2, C3, Relaxed) {
            C2 => {
                let val = init2();
                b.store(val, Relaxed);
                val
            }
            C3 => continue,
            val => val,
        };
        do_stuff2(val);
    }
}
```
At the first glance it shouldn't have any problems, since `compare_and_swap` is a side-effect operation and can not be executed before checking value of the first atomic.

newpavlov:
BTW `compare_and_swap` compiles down to a `lock` instruction, while relaxed load compiles to a simple `mov`. Maybe it's worth to add code like this before the loop?
```rust
let val = self.0.load(Relaxed);
if val != Self::UNINIT && val != SELF::ACTIVE {
    return val;
}
```

dhardy:
@newpavlov IIUC x86 is strongly ordered, implying most aquire/release boundaries are unnecessary anyway. You should therefore check code for another target like ARM.

In your example, I *think* it doesn't actually matter if `b` is loaded before `a` is stored anyway — they are independent. (But if not, the first read to `a` must be *aquire* and the store to `a` must be `release`, I believe.)

HadrienG2:
@dhardy I would advise caution against using "what the hardware does" as a reference for atomic ordering correctness, because the compiler can introduce additional reorderings on top of those that hardware performs, in a manner that can change in future releases even on a given architecture.

@newpavlov Here's my understanding/analysis of the code summary that you posted:

First, a number of threads will read `C1` from `a`, concurrently execute `init1()`, and race to write the output of that function in `a`. From my understanding, the correctness conditions at this point are:

- It must be correct for multiple executions of `init1()` to occur in parallel
- It must be correct for only one output of these executions to be remembered and the others silently discarded (but still upheld by the threads that produced them).
- From my understanding of the code's logic, I _think_ `init1()` must probably also be prevented from outputting the value `C1`.

After this first initialization stage, any thread which reads `1` from `a`, or produces that value and writes it to `a`, will then execute `do_stuff1()` without synchronizing with the initialization of `a`. This is my understanding of the correctness conditions here:

- Either `do_stuff1()` must not depend on any side-effect/write to shared memory from `init1()`, or the two functions must implement their own synchronization protocol to synchronize with each other. If both of these sentences are incorrect, then you must provide your own synchronization by using `a.load(Ordering::Acquire)` and `a.store(Ordering::Release)`.
- It must be correct for `do_stuff1()` and `init1()` to run in parallel.
- If multiple calls to `init1()` can return different values, then it must be correct for `do_stuff1()` and the entire loop from the `else` branch to run in parallel.

If `a` does not end up holding the value `1` but another value (which again I assume to be different from `C1`), then the threads move to a more classic lazy initialization logic where they race to replace the initial value of `b`, called `C2`, with another special value called `C3`. The winner of the race will get the privilege of running `init2()` and writing its output to b, while the other threads will be waiting in a spin loop. Assuming that the output of `init2()` is neither `C2` nor `C3`, every thread involved will then proceed to call `do_stuff2()` with that output of that parameter. Thus, my understanding is that...

- It must be correct for `init1()` and either of `init2()` or `do_stuff2()` to run in parallel.
- If `init1()` is allowed to sometimes return 1 and sometimes return something else, then it must additionally be correct for `do_stuff1()` and either of `init2()` or `do_stuff2()` to run in parallel.
- Either `do_stuff2()` must not depend on any side_effect of `init2()` or these side effects must be taken care of by the two functions. If that's not the case, then you must also switch to an `Acquire`/`Release` synchronization protocol here (`Acquire` on the CAS, `Release` on the store).
- It is almost certainly incorrect for `init2()` to return either `C2` or `C3`.

Hope this attempt at summarizing the underlying assumptions helps.

----

On an unrelated note, I agree that you should probably check the value of `b` before doing a compare and swap. From my personal microbenchmarks, atomic read-modify-write instructions like CAS are ~10x more expensive than loads and stores on x86 in isolation, and here you're additionally running them in a loop on multiple threads which will additionally cause cache invalidation and contention...

josephlr:
Hey everyone, I've updated this PR and the description to incorporate your recommendations, and I split up the commits to make review easier.

@HadrienG2 thanks so much for the rundown. I've tried to address all of your points by explictly putting your concerns in the comments. BTW https://github.com/rust-random/getrandom/pull/52/commits/8566c39846417e9d1f50e4812af31be41d6e767a contains the changes to `LazyStatic` and `sync_init()` which are the tricky parts of this PR. The other commits are mostly straightforward wrappers.

> * It must be correct for multiple executions of `init1()` to occur in parallel
> * It must be correct for `do_stuff1()` and `init1()` to run in parallel.
> * It must be correct for `init1()` and either of `init2()` or `do_stuff2()` to run in parallel.
> * If multiple calls to `init1()` can return different values, then it must be correct for `do_stuff1()` and the entire loop from the `else` branch to run in parallel.

For our implementation, these parallelism assumptions are correct. Also, in the documentation on `unsync_init`, I note that the `init()` function could run in parallel with other callers.

> * If `init1()` is allowed to sometimes return 1 and sometimes return something else, then it must additionally be correct for `do_stuff1()` and either of `init2()` or `do_stuff2()` to run in parallel.

In the documentation on `unsync_init`, I note that the `init()` function should always return the same value, if it succeeds. This is the case for all of our current use cases.

> * It must be correct for only one output of these executions to be remembered and the others silently discarded (but still upheld by the threads that produced them).
> * Either `do_stuff1()` must not depend on any side-effect/write to shared memory from `init1()`, or the two functions must implement their own synchronization protocol to synchronize with each other. If both of these sentences are incorrect, then you must provide your own synchronization by using `a.load(Ordering::Acquire)` and `a.store(Ordering::Release)` in the previous code snippet.
> * Either `do_stuff2()` must not depend on any side_effect of `init2()` or these side effects must be taken care of by the two functions. If that's not the case, then you must also switch to an `Acquire`/`Release` synchronization protocol here (`Acquire` on the CAS, `Release` on the store).

I note in the description of `LazyUsize` that we should only depend on the value returned by `init()` (along with including a code example). This is the case for all of our current use cases. We only depend on the value of `init()`, and code in `init()` that runs before we generate the value.

> * From my understanding of the code's logic, `init1()` must probably also be prevented from outputting the value `C1`, and I will assume that this is true in the following.
> * It is almost certainly incorrect for `init2()` to return either `C2` or `C3`.

Agreed on `init2()` returning C2, which I now mention in the comments for `sync_init`. However, I think it should be fine for `init()` to return `UNINIT`. This will just return `UNINIT` to the caller and future calls to `sync_init` and `unsync_init` will just run again.

My comments on `LazyUsize` explain this "retrying" behavior, and the assumptions necessary to make it work. PTAL

> On an unrelated note, I agree that you should probably check the value of `b` before doing a compare and swap.

Done, and good idea.

HadrienG2:
LGTM :)

src/use_file.rs
```diff
@@ -29,28 +31,31 @@ const FILE_PATH: &str = "/dev/urandom";
 const FILE_PATH: &str = "/dev/random";
 
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
-    lazy_static! {
-        static ref FILE: Result<File, Error> = init_file();
-    }
-    let mut f = FILE.as_ref()?;
+    static FD: LazyFd = LazyFd::new();
+    let fd = FD.init(init_file).ok_or(Error::UNKNOWN)?;
```
newpavlov: I think you can use `last_os_error` and convert it into a code. In future full migration from `std` it can be replaced with `errno`.

josephlr: Done

newpavlov:
Regarding CI tests for Redox we can bring them back after rust-lang/libc#1420 gets fixed.

src/util.rs
```diff
@@ -19,20 +35,44 @@ impl LazyUsize {
...
+    pub const ACTIVE: usize = usize::max_value() - 1;
 
     // Runs the init() function at least once, returning the value of some run
-    // of init(). Unlike std::sync::Once, the init() function may be run
-    // multiple times. If init() returns UNINIT, future calls to unsync_init()
-    // will always retry. This makes UNINIT ideal for representing failure.
+    // of init(). Multiple callers can run their init() functions in parallel.
+    // init() should always return the same value, if it succeeds.
     pub fn unsync_init(&self, init: impl FnOnce() -> usize) -> usize {
         // Relaxed ordering is fine, as we only have a single atomic variable.
```
RalfJung: This assumes that a file descriptor consists *only* of that integer and no other user-space data. I suppose that is accurate on Unixes?

src/util.rs
```diff
@@ -19,20 +35,44 @@ impl LazyUsize {
...
+        match self.0.load(Relaxed) {
+            Self::UNINIT | Self::ACTIVE => {}
+            val => return val,
+        }
+        // Relaxed ordering is fine, as we only have a single atomic variable.
+        loop {
+            match self.0.compare_and_swap(Self::UNINIT, Self::ACTIVE, Relaxed) {
+                Self::UNINIT => {
+                    let val = init();
+                    self.0.store(val, Relaxed);
```
RalfJung: IMO it would be a good idea to `assert!` here that `init` did not return `UNINIT` or `ACTIVE`.  This is not hot code, so always asserting seems fine.

src/util_libc.rs
```diff
@@ -36,3 +36,29 @@ impl Weak {
...
+    // If init() returns Some(x), x should be nonnegative.
+    pub fn init(&self, init: impl FnOnce() -> Option<libc::c_int>) -> Option<libc::c_int> {
+        let fd = self.0.sync_init(
+            || match init() {
+                // OK as val >= 0 and val <= c_int::MAX < usize::MAX
+                Some(val) => val as usize,
+                None => LazyUsize::UNINIT,
+            },
+            || unsafe {
+                libc::usleep(1000);
```
RalfJung: Wow that's a long sleep, isn't it? A comment seems always warranted with such a magic number.

src/util.rs
RalfJung: Oh, seems like the function is actually allowed to return `UNINIT` to indicate "just try again"? I guess this is implied by the doc comment *not* ruling out that case, but might be worth calling out explicitly.

`ACTIVE` can still be asserted.

src/util.rs
josephlr: Yup this is correct, [`libc::open` always returns a positive int](http://man7.org/linux/man-pages/man2/open.2.html) (link is to the Linux docs, but the function is a POSIX standard).

So a file descriptor will always fit in a positive int on unix

src/util_libc.rs
josephlr: Ya this sleep should probably be shorter, any recommendations? My guess is that we would want it to be longer than the normal latency for an `open(2)` syscall.

src/util.rs
josephlr: > Oh, seems like the function is actually allowed to return `UNINIT` to indicate "just try again"? I guess this is implied by the doc comment _not_ ruling out that case, but might be worth calling out explicitly.

This is called out, just in the docs for `LazyUsize`.

```rust
// Both methods support init() "failing". If the init() method returns UNINIT,
// that value will be returned as normal, but will not be cached.
```

> `ACTIVE` can still be asserted.

Good idea. Thanks @RalfJung for taking a look at this stuff.

newpavlov: I think it may be better to write something like this:
```
let mut val = init();
if val == Self::ACTIVE { val = Self::UNINIT }
self.0.store(val, Relaxed);
```
And describe this behavior in docs.

src/util_libc.rs
newpavlov: On my laptop `libc::open` takes ~7µs and on playground ~8µs, so I think 10µs will be a good value.

RalfJung: I'm sorry this is outside my range of experience.^^ 1ms just seemed like a long time. `/dev/random` doesn't even do "real" I/O, so it shouldn't have *huge* latency.

newpavlov: It can be a huge latency if system entropy pool has not initialized, IIRC up to several minutes.

RalfJung: Sure, but that's not the common case.

Ideally `wait` would do something like double the wait time each time around the loop, starting at e.g. 1µs and maxing out at 1s or so. But that seems like overkill.^^

josephlr:
@newpavlov @RalfJung this is a good discussion, but might be better had on https://github.com/rust-random/getrandom/pull/60 instead of this closed PR.

PR #54, opened 2019-07-05
Improve Error handling

### LINKED ISSUE #49: Stop using dummy implementation
Inspired by @dhardy's comment https://github.com/rust-lang/rust/pull/62082#issuecomment-505822196

This would simply remove the dummy implementation, causing a compiler error on unsupported platforms. This makes it much easier to detect platform support issues (as a compiler error is permeable to a run time error). This would also make it clear that when porting `libstd` or `rand` to a different platform, an implementation in `getrandom` is needed.

For example, @newpavlov's code in https://github.com/rust-lang/rust/pull/62082 would be changed to:
```rust
let mut buf = [0u8; 16];

// Using a constant value is acceptable on whitelisted targets.
#[cfg(not(target = "wasm32-unknown-unknown"))]
if let Err(err) = getrandom::getrandom(&mut buf) {
    panic!("getrandom failure: {:?}", err)
}           

let n = u128::from_ne_bytes(buf);
Cell::new([n as u64, (n >> 64) as u64])
```

Note that this would be a breaking change, so it would require `0.2`

josephlr:
Edit: Significantly revised to account for earlier feedback. Depends on #57 (for `fill_exact`)

The PR allows for error handling to work without depending on `libstd`, it also cleans up the `Error` type. Specifically:

- The `Error` type maintains its `NonZeroU32` representation. But now instead of holding a single code, it holds one of two values:
  - An OS Error: represented by a positive `i32` value
  - A custom Error: represented by a `u32` value larger than `i32::MAX`.
- We improve the `Debug` and `Display` implementations for OS Errors (on UNIX) by using `libc::strerror_r`.
- We improve the `Debug` and `Display` implementations for Custom Errors (on some platforms) by defining additional error constants (instead of just returning `Error::UNKNOWN`)
- `util_libc::last_os_error()` replaces `std::io::Error::last_os_error()`.

Now the only things that depend on `libstd` are:
  - Conversions enabled by the `std` feature
  - File-based implementations (to be fixed in a followup PR).

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -13,32 +13,21 @@ use core::num::NonZeroU32;
 /// The [`Error::raw_os_error()`] will indicate if the error is from the OS, and
 /// if so, which error code the OS gave the application. If such an error is
 /// encountered, please consult with your system documentation.
+///
+/// Internally this type is a NonZeroU32, with certain values reserved for
+/// certain purposes, see [`Error::INTERNAL_START`] and [`Error::CUSTOM_START`].
 #[derive(Copy, Clone, Eq, PartialEq)]
 pub struct Error(NonZeroU32);
 
-// This NonZeroU32 in Error has enough room for two types of errors:
-//   - OS Errors:       in range [1, 1 << 31) (i.e. positive i32 values)
-//   - Custom Errors:   in range [1 << 31, 1 << 32) (in blocks of 1 << 16)
-const CUSTOM_START: u32 = 1 << 31;
-const BLOCK_SIZE: u32 = 1 << 16;
-
 impl Error {
-    /// Create a new error from a raw OS error number (errno).
-    #[inline]
-    pub fn from_os_error(errno: i32) -> Self {
-        assert!(errno > 0);
-        Self(NonZeroU32::new(errno as u32).unwrap())
-    }
+    /// Codes below this point represent OS Errors (i.e. positive i32 values).
+    /// Codes at or above this point, but below [`Error::CUSTOM_START`] are
+    /// reserved for use by the `rand` and `getrandom` crates.
+    pub const INTERNAL_START: u32 = 1 << 31;
 
-    /// Crate a custom error in the provided block (group of 2^16 error codes).
-    /// The provided block must not be negative, and block 0 is reserved for
-    /// custom errors in the `getrandom` crate.
-    #[inline]
-    pub fn custom_error(block: i16, code: u16) -> Self {
-        assert!(block >= 0);
-        let n = CUSTOM_START + (block as u16 as u32) * BLOCK_SIZE + (code as u32);
-        Self(NonZeroU32::new(n).unwrap())
-    }
+    /// Codes at or above this point can be used by users to define their own
+    /// custom errors.
+    pub const CUSTOM_START: u32 = (1 << 31) + (1 << 30);
 
     /// Extract the raw OS error code (if this error came from the OS)
     ///
@@ -47,32 +36,20 @@ impl Error {
     /// error value can still be formatted via the `Diplay` implementation.
     #[inline]
     pub fn raw_os_error(&self) -> Option<i32> {
-        self.try_os_error().ok()
+        if self.0.get() < Self::INTERNAL_START {
+            Some(self.0.get() as i32)
+        } else {
+            None
+        }
     }
 
     /// Extract the bare error code.
     ///
     /// This code can either come from the underlying OS, or be a custom error.
-    /// Use [`raw_os_error()`] to disambiguate.
+    /// Use [`Error::raw_os_error()`] to disambiguate.
     #[inline]
     pub fn code(&self) -> NonZeroU32 {
         self.0
-    }
-
-    /// Helper method for creating internal errors
-    #[allow(dead_code)]
-    pub(crate) fn internal(code: u16) -> Self {
-        Self::custom_error(0, code)
-    }
-
-    /// Returns either the OS error or a (block, code) pair
-    fn try_os_error(&self) -> Result<i32, (i16, u16)> {
-        if self.0.get() < CUSTOM_START {
-            Ok(self.0.get() as i32)
-        } else {
-            let offset = self.0.get() - CUSTOM_START;
-            Err(((offset / BLOCK_SIZE) as i16, (offset % BLOCK_SIZE) as u16))
-        }
     }
 }
 
@@ -96,24 +73,17 @@ fn os_err_desc(_errno: i32, _buf: &mut [u8]) -> Option<&str> {
 impl fmt::Debug for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let mut dbg = f.debug_struct("Error");
-        match self.try_os_error() {
-            Ok(errno) => {
-                dbg.field("os_error", &errno);
-                let mut buf = [0u8; 128];
-                if let Some(desc) = os_err_desc(errno, &mut buf) {
-                    dbg.field("description", &desc);
-                }
+        if let Some(errno) = self.raw_os_error() {
+            dbg.field("os_error", &errno);
+            let mut buf = [0u8; 128];
+            if let Some(desc) = os_err_desc(errno, &mut buf) {
+                dbg.field("description", &desc);
             }
-            Err((0, code)) => {
-                dbg.field("internal_code", &code);
-                if let Some(desc) = internal_desc(code) {
-                    dbg.field("description", &desc);
-                }
-            }
-            Err((block, code)) => {
-                dbg.field("block", &block);
-                dbg.field("custom_code", &code);
-            }
+        } else if let Some(desc) = internal_desc(*self) {
+            dbg.field("internal_code", &self.0.get());
+            dbg.field("description", &desc);
+        } else {
+            dbg.field("unknown_code", &self.0.get());
         }
         dbg.finish()
     }
@@ -121,19 +91,16 @@ impl fmt::Debug for Error {
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match self.try_os_error() {
-            Ok(errno) => {
-                let mut buf = [0u8; 128];
-                match os_err_desc(errno, &mut buf) {
-                    Some(desc) => f.write_str(desc),
-                    None => write!(f, "OS Error: {}", errno),
-                }
-            }
-            Err((0, code)) => match internal_desc(code) {
+        if let Some(errno) = self.raw_os_error() {
+            let mut buf = [0u8; 128];
+            match os_err_desc(errno, &mut buf) {
                 Some(desc) => f.write_str(desc),
-                None => write!(f, "Internal Error: {}", code),
-            },
-            Err((block, code)) => write!(f, "Custom Error: block={}, code={}", block, code),
+                None => write!(f, "OS Error: {}", errno),
+            }
+        } else if let Some(desc) = internal_desc(*self) {
+            f.write_str(desc)
+        } else {
+            write!(f, "Unknown Error: {}", self.0.get())
         }
     }
 }
@@ -145,19 +112,23 @@ impl From<NonZeroU32> for Error {
 }
 
 /// Internal Error constants
-pub(crate) const UNSUPPORTED: u16 = 0;
-pub(crate) const UNKNOWN_IO_ERROR: u16 = 1;
-pub(crate) const SEC_RANDOM_FAILED: u16 = 2;
-pub(crate) const RTL_GEN_RANDOM_FAILED: u16 = 3;
-pub(crate) const FAILED_RDRAND: u16 = 4;
-pub(crate) const NO_RDRAND: u16 = 5;
-pub(crate) const BINDGEN_CRYPTO_UNDEF: u16 = 6;
-pub(crate) const BINDGEN_GRV_UNDEF: u16 = 7;
-pub(crate) const STDWEB_NO_RNG: u16 = 8;
-pub(crate) const STDWEB_RNG_FAILED: u16 = 9;
+pub(crate) const UNSUPPORTED: Error = internal_error(0);
+pub(crate) const UNKNOWN_IO_ERROR: Error = internal_error(1);
+pub(crate) const SEC_RANDOM_FAILED: Error = internal_error(2);
+pub(crate) const RTL_GEN_RANDOM_FAILED: Error = internal_error(3);
+pub(crate) const FAILED_RDRAND: Error = internal_error(4);
+pub(crate) const NO_RDRAND: Error = internal_error(5);
+pub(crate) const BINDGEN_CRYPTO_UNDEF: Error = internal_error(6);
+pub(crate) const BINDGEN_GRV_UNDEF: Error = internal_error(7);
+pub(crate) const STDWEB_NO_RNG: Error = internal_error(8);
+pub(crate) const STDWEB_RNG_FAILED: Error = internal_error(9);
 
-fn internal_desc(code: u16) -> Option<&'static str> {
-    match code {
+const fn internal_error(n: u16) -> Error {
+    Error(unsafe { NonZeroU32::new_unchecked(Error::INTERNAL_START + n as u32) })
+}
+
+fn internal_desc(error: Error) -> Option<&'static str> {
+    match error {
         UNSUPPORTED => Some("getrandom: this target is not supported"),
         UNKNOWN_IO_ERROR => Some("Unknown std::io::Error"),
         SEC_RANDOM_FAILED => Some("SecRandomCopyBytes: call failed"),
--- a/src/error.rs
+++ b/src/error.rs
@@ -20,6 +20,11 @@ use core::num::NonZeroU32;
 pub struct Error(NonZeroU32);
 
 impl Error {
+    #[deprecated(since = "0.1.7")]
+    pub const UNKNOWN: Error = UNSUPPORTED;
+    #[deprecated(since = "0.1.7")]
+    pub const UNAVAILABLE: Error = UNSUPPORTED;
+
     /// Codes below this point represent OS Errors (i.e. positive i32 values).
     /// Codes at or above this point, but below [`Error::CUSTOM_START`] are
     /// reserved for use by the `rand` and `getrandom` crates.
@@ -111,25 +116,31 @@ impl From<NonZeroU32> for Error {
     }
 }
 
-/// Internal Error constants
-pub(crate) const UNSUPPORTED: Error = internal_error(0);
-pub(crate) const UNKNOWN_IO_ERROR: Error = internal_error(1);
-pub(crate) const SEC_RANDOM_FAILED: Error = internal_error(2);
-pub(crate) const RTL_GEN_RANDOM_FAILED: Error = internal_error(3);
-pub(crate) const FAILED_RDRAND: Error = internal_error(4);
-pub(crate) const NO_RDRAND: Error = internal_error(5);
-pub(crate) const BINDGEN_CRYPTO_UNDEF: Error = internal_error(6);
-pub(crate) const BINDGEN_GRV_UNDEF: Error = internal_error(7);
-pub(crate) const STDWEB_NO_RNG: Error = internal_error(8);
-pub(crate) const STDWEB_RNG_FAILED: Error = internal_error(9);
-
-const fn internal_error(n: u16) -> Error {
-    Error(unsafe { NonZeroU32::new_unchecked(Error::INTERNAL_START + n as u32) })
+// TODO: Convert to a function when min_version >= 1.33
+macro_rules! internal_error {
+    ($code:expr) => {{
+        let n: u16 = $code;
+        Error(unsafe { NonZeroU32::new_unchecked(Error::INTERNAL_START + n as u32) })
+    }};
 }
+
+/// Internal Error constants
+pub(crate) const UNSUPPORTED: Error = internal_error!(0);
+pub(crate) const ERRNO_NOT_POSITIVE: Error = internal_error!(1);
+pub(crate) const UNKNOWN_IO_ERROR: Error = internal_error!(2);
+pub(crate) const SEC_RANDOM_FAILED: Error = internal_error!(3);
+pub(crate) const RTL_GEN_RANDOM_FAILED: Error = internal_error!(4);
+pub(crate) const FAILED_RDRAND: Error = internal_error!(5);
+pub(crate) const NO_RDRAND: Error = internal_error!(6);
+pub(crate) const BINDGEN_CRYPTO_UNDEF: Error = internal_error!(7);
+pub(crate) const BINDGEN_GRV_UNDEF: Error = internal_error!(8);
+pub(crate) const STDWEB_NO_RNG: Error = internal_error!(9);
+pub(crate) const STDWEB_RNG_FAILED: Error = internal_error!(10);
 
 fn internal_desc(error: Error) -> Option<&'static str> {
     match error {
         UNSUPPORTED => Some("getrandom: this target is not supported"),
+        ERRNO_NOT_POSITIVE => Some("errno: did not return a positive value"),
         UNKNOWN_IO_ERROR => Some("Unknown std::io::Error"),
         SEC_RANDOM_FAILED => Some("SecRandomCopyBytes: call failed"),
         RTL_GEN_RANDOM_FAILED => Some("RtlGenRandom: call failed"),
--- a/src/error.rs
+++ b/src/error.rs
@@ -5,78 +5,166 @@
 // <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
-use core::convert::From;
 use core::fmt;
 use core::num::NonZeroU32;
 
-// A randomly-chosen 24-bit prefix for our codes
-pub(crate) const CODE_PREFIX: u32 = 0x57f4c500;
-const CODE_UNKNOWN: u32 = CODE_PREFIX | 0x00;
-const CODE_UNAVAILABLE: u32 = CODE_PREFIX | 0x01;
-
-/// The error type.
+/// A small and `no_std` compatible error type.
 ///
-/// This type is small and no-std compatible.
+/// The [`Error::raw_os_error()`] will indicate if the error is from the OS, and
+/// if so, which error code the OS gave the application. If such an error is
+/// encountered, please consult with your system documentation.
 #[derive(Copy, Clone, Eq, PartialEq)]
-pub struct Error(pub(crate) NonZeroU32);
+pub struct Error(NonZeroU32);
+
+// This NonZeroU32 in Error has enough room for two types of errors:
+//   - OS Errors:       in range [1, 1 << 31) (i.e. positive i32 values)
+//   - Custom Errors:   in range [1 << 31, 1 << 32) (in blocks of 1 << 16)
+const CUSTOM_START: u32 = 1 << 31;
+const BLOCK_SIZE: u32 = 1 << 16;
 
 impl Error {
-    /// An unknown error.
-    pub const UNKNOWN: Error = Error(unsafe { NonZeroU32::new_unchecked(CODE_UNKNOWN) });
+    /// Create a new error from a raw OS error number (errno).
+    pub fn from_os_error(errno: i32) -> Self {
+        assert!(errno > 0);
+        Self(NonZeroU32::new(errno as u32).unwrap())
+    }
 
-    /// No generator is available.
-    pub const UNAVAILABLE: Error = Error(unsafe { NonZeroU32::new_unchecked(CODE_UNAVAILABLE) });
+    /// Crate a custom error in the provided block (group of 2^16 error codes).
+    /// The provided block must not be negative, and block 0 is reserved for
+    /// custom errors in the `getrandom` crate.
+    pub fn custom_error(block: i16, code: u16) -> Self {
+        assert!(block >= 0);
+        let n = CUSTOM_START + (block as u32) * BLOCK_SIZE + (code as u32);
+        Self(NonZeroU32::new(n).unwrap())
+    }
 
-    /// Extract the error code.
+    /// Extract the raw OS error code (if this error came from the OS)
     ///
-    /// This may equal one of the codes defined in this library or may be a
-    /// system error code.
+    /// This method is identical to `std::io::Error::raw_os_error()`, except
+    /// that it works in `no_std` contexts. If this method returns `None`, the
+    /// error value can still be formatted via the `Diplay` implementation.
+    pub fn raw_os_error(&self) -> Option<i32> {
+        self.try_os_error().ok()
+    }
+
+    /// Extract the bare error code.
     ///
-    /// One may attempt to format this error via the `Display` implementation.
+    /// This code can either come from the underlying OS, or be a custom error.
+    /// Use [`raw_os_error()`] to disambiguate.
     pub fn code(&self) -> NonZeroU32 {
         self.0
     }
 
-    pub(crate) fn msg(&self) -> Option<&'static str> {
-        if let Some(msg) = crate::imp::error_msg_inner(self.0) {
-            Some(msg)
+    /// Helper method for creating internal errors
+    #[allow(dead_code)]
+    pub(crate) fn internal(code: u16) -> Self {
+        Self::custom_error(0, code)
+    }
+
+    /// Returns either the OS error or a (block, code) pair
+    fn try_os_error(&self) -> Result<i32, (i16, u16)> {
+        if self.0.get() < CUSTOM_START {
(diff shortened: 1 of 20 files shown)


dhardy:
This change to the error type looks like it would [interfere with `rand_core`](https://github.com/rust-random/rand/blob/master/rand_core/src/error.rs#L96). Specifically we still need a `NonZeroU32` code, unless you want to chip in [here](https://github.com/rust-random/rand/issues/837) with another idea.

newpavlov:
>We issue a compiler_error! on unsupported platforms

I think we should not rush this decision and possibly postpone it. (ideally we would make this change together with dropping `wasm32-unknown-unknown` support)

>The public Error type is now an opaque struct, wrapping a implementation-specific data structure. 

What is rationale behind this change?

josephlr:
> I think we should not rush this decision and possibly postpone it. (ideally we would make this change together with dropping `wasm32-unknown-unknown` support)

Agreed, I put this change in this PR, but it really should be in a separate PR. My initial thought was to put all of the breaking changes together, but I think it makes sense for them to be split up.

> What is rationale behind this change?

My basic reasoning was that currently `getrandom::Error` initially looks like it has values can be `match`ed on. However, when I was trying to use it, I quickly realized that the only useful thing I can do with these types is print them out with `Display` or `Debug`. The main reason for this problem is that (depending on the target) the reasons for failure can be wildly different. This change just makes this implicit behavior explicit.

> This change to the error type looks like it would [interfere with `rand_core`](https://github.com/rust-random/rand/blob/master/rand_core/src/error.rs#L96). Specifically we still need a `NonZeroU32` code, unless you want to chip in [here](https://github.com/rust-random/rand/issues/837) with another idea.

I guess I'm surprised that `rand_core::Error` even exists. I has just assumed that the `RngCore` trait would just have a `type Error` parameter, and each generator would chose its own type. is there a reason things are not done that way?

@newpavlov I saw that you were against such a change in https://github.com/rust-random/rand/issues/837 (adding an associated type). But if `getrandom` implemented this change, and `rand_core` implemented the associated types change, we would get `no_std` printable errors, as each generator would just specify it's own type.

newpavlov:
>My basic reasoning was that currently getrandom::Error initially looks like it has values can be matched on. However, when I was trying to use it, I quickly realized that the only useful thing I can do with these types is print them out with Display or Debug. 

But you can retrieve `NonZeroU32` error code via `code` method and match on it. This will allow you to trap some platform-specific error codes if needed.

I will answer regarding `rand_core::Error` in the dedicated thread.

josephlr:
> This will allow you to trap some platform-specific error codes if needed

While this is _possible_, there's not a good way to actually go about doing it (as it's not clear what a user would even match on, or what types of values can be returned). This PR tries to address this by implementing `From<Error> for std::io::Error` which seems better suited for matching, printing, etc... This requires the `std` feature, but even without it, you still get good messages.

There's also the bigger issue that certain implementations _can't_ return a meaningful error code.
 Windows and iOS really shouldn't use `std::io::Error::last_os_error()` or `errno`; their  implementations don't specify how those functions interact with `errno`. But by ditching error codes for an opaque type, we actually get _better_ error messages and a more efficient implementation (as the Error types can be smaller). It also lets certain platforms (like Fuchsia) have an Error that's Infallible.

newpavlov:
> This PR tries to address this by implementing From<Error> for std::io::Error

Nothing forbids as from adding conversion to `io::Error` for current design as well, so i don't think it's a valid argument in favor of making error type opaque.

>There's also the bigger issue that certain implementations can't return a meaningful error code.
Windows and iOS really shouldn't use std::io::Error::last_os_error() or errno

While it's indeed may be a valid critique of the current implementation (though it also applies to the current std [implementation](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/windows/rand.rs) as well). But it's not an argument in favor of your error type design.

>But by ditching error codes for an opaque type, we actually get better error messages and a more efficient implementation (as the Error types can be smaller).

But for most platforms you use `io::Error` which is 16 bytes on 64-bit platforms instead of the current 4. I don't see how it's more efficient. While an ability to indicate infallibility is indeed nice, I don't think it will matter performance wise. The current `Result<(), Error>` fits into a single register, so in the most cases it will not even hit memory! (well, assuming we use 32 or 64 bit CPU) So 1 byte vs 4 will not make any difference here, thus efficiency argument is also weak in my opinion.

josephlr:
I agree with most of the above, as the changes I made here could be made with the old implementation. I've also updated this PR to make a few of those changes (specifically fixing Windows and iOS). Most the the efficiency arguments don't matter as long as `Result<(), Error>` fits in a `usize` (which should be possible for all platforms).

If a user needs raw OS error codes, the current implementation allows that by converting to `std::io::Error` and calling `raw_os_error()`. However, this doesn't work on `no_std`.

So @newpavlov what about a mix of what we've been talking about? Each target can still define the type of error that makes sense, essentially one of:
  - `io::Error` (soon to be just an `errno` implementation inside of `getrandom`)
  - `Infallible`
  - `StaticError` (if all you can indicate is just a static `str`)
  - `WebError` (for `stdweb`)

This `Error` type implements `Debug`, `Display`, and can be converted into a `io::Error`. However, it also implements a `raw_os_error()` method returning an `Option<i32>` that gives the raw OS error code. This would seem to give us the best of everything:
  - We don't have to deal with the weird artificial error codes that currently exist for the Error type
  - We get better error messages with simpler code.
  - Users now know if they hit an OS error or not.
  - `no_std` users can use `raw_os_error()`.
  - `std` users can just use `io::Error` (useful for composing `getrandom` with other IO operations).

newpavlov:
I am certainly against using `io::Error` instead of a simple error code. Not only it ties us to `std` which goes contrary to the `std` integration work, but also simply feels wasteful. Swapping `io::Error` with error code without enabled `std` feature looks like a needless complication. If you want reuse formatting functionality you can do it simply inside `cfg`'d `Debug` and `Display` impls. And I don't see much uses for matching on `io::ErrorKind`, especially considering that you propose using `io::Error` only for some platforms.

Swapping error type based on the target does not look like a simplification to me, even worse I have a hunch that it will only cause weird issues in future. Also it looks like you assume that `rand_core` will use associated error type, as you can see I am not sure if it's a better design compared to the current one.

Overall, motivation of this rework looks a bit weak to me. Looks like the main driving factor is desire to improve `Debug` and `Display` impls, but you already can improve them via `error_msg_inner`.

newpavlov:
Meanwhile can you please submit `cfg-if` and Windows/iOS changes as a separate PR?

josephlr:
> I am certainly against using `io::Error` instead of a simple error code.

So there are two issues here:

  -  Using `io::Error` as the error type on Unix targets.
    - This is just a temporary artifact until my next PR removes the `std::io` reqirement for all targets (by just using `libc` errno). Then it will be an `i32` or something.
  - Having `io::Error: From<getrandom::Error>`. This is very valuable as it lets you call `getrandom` and then other io functions, and easily use `?` to have your parent function return a `io::Result`.

The switching on `io::Error` type is not ideal I agree. This is why I proposed having `raw_os_error()` that allows users on all platforms to see if/what the underlying OS error was.

> Also it looks like you assume that `rand_core` will use associated error type, as you can see I am not sure if it's a better design compared to the current one.

So this is actually why I think having a `raw_os_error()` method directly on `getrandom::Error` for all platforms is a good idea. It allows `getrandom` to be updated without resolving the issue of associated types in `rng_core`. The `rng_core` crate can just use

dhardy:
I agree with most of what @newpavlov says; from the API perspective this isn't a simplification, and you haven't attempted to explain how this works with `rand_core` (which currently supports `getrandom` in `no_std` mode for all platforms due to the weird way feature flags work).

> > This will allow you to trap some platform-specific error codes if needed
> 
> While this is possible, there's not a good way to actually go about doing it (as it's not clear what a user would even match on, or what types of values can be returned).

Nor do I see why we should do so — the source code is public, and the only reason I can think of why anyone would need to do this is to work around some weird issue. (We don't even report "transient" failures any more.)

josephlr:
> Meanwhile can you please submit cfg-if and Windows/iOS changes as a separate PR?

Sure! 

After that, should I put that changes that call `errno` directly (and remove the `std::io` dependancy) in this PR or another one.

josephlr:
> Nor do I see why we should do so — the source code is public, and the only reason I can think of why anyone would need to do this is to work around some weird issue. (We don't even report "transient" failures any more.)

That's fair, but if we don't want users to match on the type, then shouldn't the return type be more opaque, not less?

dhardy:
> So this is actually why I think having a raw_os_error() method directly on getrandom::Error for all platforms is a good idea

How does this work on platforms not able to use `io::Error` directly?

> After that, should I put that changes that call errno directly (and remove the std::io dependancy) in this PR or another one.

We can't accept this PR as is for a number of reasons, one of which is the additional dependency on `std`, so if it helps resolve that it makes sense to do so in this PR or before this one.

josephlr:
> How does this work on platforms not able to use `io::Error` directly?

So like the one's that use RDRAND? Then `raw_os_error()` would return `None` indicating that the failure was not due to an OS error. This is essentially what we're already doing with `Error::UNKNOWN` just more explicit.

newpavlov:
Another argument against using target-dependent type is potential future addition of `getrandom` to Rust in the form of lang item à la `#[global_allocator]`. Having a concrete error type shared across all targets will make it significantly easier.

josephlr:
> Another argument against using target-dependent type is potential future addition of `getrandom` to Rust in the form of lang item à la `#[global_allocator]`. Having a concrete error type shared across all targets will make it significantly easier.

That's a really good point I had not thought of.

If that's the case then it might be even easier to just ditch having any information at all in the error type. Error's in `getrandom` would just be a single value indicating that the OS could not provide randomness (and maybe a `TARGET_UNSUPPORTED` value, if we want to not use `compiler_error`). This is how [`ring` handles this sort of problem](https://docs.rs/ring/0.15.0-alpha3/ring/error/struct.Unspecified.html).

newpavlov:
I think `ring` approach is a bit too extreme, and in our case will make debugging of potential errors much harder. Was it descriptor exhaustion? Incorrectly implemented syscall? Did someone screw with `/dev/urandom`? How exactly did hardware RNG misbehaved? Having an error code will significantly reduce scope of potential problems.

None of the 3 reasons stated in the `Unspecified` documentation applies to our case in my opinion.

josephlr:
@newpavlov @dhardy I significantly reworked this PR to take your recommendations. The PR description has been updated. PTAL

This change now focuses more on removing `libstd` deps for error handling, and only makes minor changes to `Error`.

src/ios.rs
```diff
@@ -27,14 +26,16 @@ extern "C" {
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
     let ret = unsafe { SecRandomCopyBytes(kSecRandomDefault, dest.len(), dest.as_mut_ptr()) };
     if ret == -1 {
-        error!("SecRandomCopyBytes call failed");
-        Err(Error::UNKNOWN)
+        Err(Error::from_custom_error(CALL_FAILED))
```
dhardy: I don't know if this case is ever likely to be encountered, but if it is, the new code reports only `UNAVAILABLE` (unless `CALL_FAILED` is supposed to be distinct).

I think we should keep the `error` log? These macros are optimised out completely if logging is not enabled.

src/rdrand.rs
```diff
@@ -33,12 +37,15 @@ unsafe fn rdrand() -> Result<[u8; WORD_SIZE], Error> {
...
                 return Ok(el.to_ne_bytes());
             }
-            error!("RDRAND returned {:X}, CPU RNG may be broken", el);
+            is_bad = true;
             // Keep looping in case this was a false positive.
         }
     }
-    error!("RDRAND failed, CPU issue likely");
-    Err(Error::UNKNOWN)
+    Err(Error::from_custom_error(if is_bad {
```
dhardy: Why (1) is reporting distinct errors if `is_bad` useful, and (2) do we not want to log via `error!`?

src/rdrand.rs
```diff
@@ -12,7 +12,10 @@ use crate::util::LazyBool;
 use crate::Error;
 use core::arch::x86_64::_rdrand64_step;
 use core::mem;
-use core::num::NonZeroU32;
+
+const NO_RDRAND: u16 = 0;
+const BAD_RDRAND: u16 = 1;
+const FAILED_RDRAND: u16 = 2;
```
dhardy: Re-using the same custom codes across multiple platforms doesn't seem useful; it's not like we're short of them. I think it would be better to have a single list of all custom codes in the `error` module?

This implies we could merge all `custom_description` implementations — although that has the disadvantage of defining all error strings in all binaries, so we might also prefer not to do this. (Either way I think the codes should be distinct.)

src/error.rs
```diff
@@ -5,78 +5,119 @@
...
         }
     }
-}
 
-impl fmt::Debug for Error {
-    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
-        match self.msg() {
-            Some(msg) => write!(f, "Error(\"{}\")", msg),
-            None => write!(f, "Error(0x{:08X})", self.0),
+    pub fn custom_code(&self) -> Option<u16> {
```
dhardy: Undocumented public fn.

One thing seems incongruous to me: we have `2<<31 - 1` error codes available, but only allow use of 2<<16 of these (labelled 'custom'). I think the issue is just terminology: these codes are for "getrandom internal" usage.

Perhaps we should use values from `2<<32 - 2<<16` as "special/custom codes"?

src/error.rs
```diff
@@ -5,78 +5,119 @@
...
-    /// Extract the error code.
-    ///
-    /// This may equal one of the codes defined in this library or may be a
-    /// system error code.
-    ///
-    /// One may attempt to format this error via the `Display` implementation.
-    pub fn code(&self) -> NonZeroU32 {
-        self.0
+    pub(crate) fn from_custom_error(custom: u16) -> Self {
+        Self(NonZeroU32::new(custom as u32 + Self::CUSTOM_START).unwrap())
```
dhardy: If we use a thin wrapper around this for `rand_core::Error` then we need some public way of constructing errors with custom codes. Maybe we should even use a pair: `block: u16, code: u16`?

src/rdrand.rs
josephlr: This was a response to: https://github.com/rust-random/getrandom/pull/48#pullrequestreview-256038770, but I now think returning different error codes is more confusing rather than less confusing. Now we just `error!` log suspicious values and return `FAILED_RDRAND` in both cases.

src/ios.rs
josephlr: Now that the codes have been moved to `error.rs`, the `UNAVAILABLE` and `UNKNOWN` errors have been removed. 

As for the `error!` logging. My strategy was to:

- remove logs that just had the same error message as the returned error (to prevent log spew)
- keep logs that added additional context (to aid in debugging)

Do you want me the add the `error!("SecRandomCopyBytes call failed");` lines back?  We're returning an `Error` who's display implementation is `"SecRandomCopyBytes: call failed"`.

src/rdrand.rs
josephlr: I moved all the error constants to `error.rs` and now have a single `internal_desc` function to get the description for each error constant.

I don't think we should worry about a few more string constants in static memory (it's only like O(100) bytes). If a user really cares, they can just remove any calls to `Debug` or `Display` the `Error`, and then LTO will get rid of the strings.

src/error.rs
josephlr: Fixed though use of the `from_kind()` method to mirror the `kind()` method. Now users can create their own errors like:
```rust
const MY_BLOCK: i16 = 2;
const MY_ERROR: u16 = 200;
Error::from_kind(ErrorKind::Custom(MY_BLOCK, MY_ERROR)
```

src/error.rs
josephlr: I reworked how a user gets the error information from an `Error` by introducing an `ErrorKind` enum (and corresponding `Error::kind()` method). This enum represents the three sources of errors:

- `Os` errors (positive `i32` values)
- `Internal` errors (`u16` values)
- `Custom` errors consisting on two parts:
  - `block` - A positive `i16` used to group errors together
  - `code` - The actual `u16` error code within a block.
- In a sense, our `Internal` errors are `Custom` errors in "block 0".

newpavlov:
I will read changes more thoroughly later. First impressions for now as follows: while I like retrieving error descriptions using `libc`, I don't quite understand motivation behind 3 kinds of errors instead of just 2: OS and custom. The highest bit for the former will be equal to zero and for the latter it will be one. Doing so will simplify code a bit.

Also I don't think we need `ErrorKind` as part of our public API. And I would prefer to keep backwards compatibility, especially with the `code` method, since it's already used in `rand`.

josephlr:
> I don't quite understand motivation behind 3 kinds of errors instead of just 2: OS and custom. The highest bit for the former will be equal to zero and for the latter it will be one.

The intention was to make it clear what caused the `Error`, the OS, `getrandom`, or something else. However, I think that if we're considering using this type in `rand_core::Error` (like @dhardy mentioned) then this like between `Internal` and `Custom` get blurred. So eliminating `Internal` makes sense.

> Also I don't think we need `ErrorKind` as part of our public API. And I would prefer to keep backwards compatibility, especially with the `code` method, since it's already used in `rand`.

The main use case for this is `Error::kind()` allowing a user to easily match on an OS error or Custom error. The alternative is to just have two methods `raw_os_error()` and `custom_error()` that return `Option`. But that doesn't really express the fact the exactly one of those functions will return `None`.

An alternative would be to:

- Not have `ErrorKind`
- Not have `Error: From<NonZeroU32>` 
- Have separate functions for creating an error from:
  - An OS error: `from_os_error(errno: i32) -> Error` 
  - A custom block/code pair: `from_custom_error(block: i32, code: u32) -> Error` 
- Have `raw_os_error() -> Option<i32>`
- Don't have a `custom_code()` function.
- Deprecate `code() -> NonZeroU32`
  - Remove it when https://github.com/rust-random/rand/issues/837 is fixed to stop using a raw code.
  - This would make it explicit that you shouldn't `match` on custom error codes.

newpavlov:
>The main use case for this is Error::kind() allowing a user to easily match on an OS error or Custom error.

But what exactly for it will be used? Outside of formatting I don't quite see use-cases for distinguishing OS and custom error.

Also there is a simpler solution: keep `From<NonZeroU32>` impl and `code` method, and add `is_os_error(&self) -> bool` method, which will simply check the highest bit. I don't think we really need a type-level hard-enforcement of the difference between OS and custom errors. Custom error code constants can simply have the highest bit set to one, so we will not need a dedicated `custom_code` method.

josephlr:
@newpavlov I fixed this PR to be on top of #57 and to not make backwards incompatible changes to `Error`. The following public functions are added to `Error`:
  - `from_os_error(errno: i32) -> Self`
  - `custom_error(block: i16, code: u16) -> Self`
  - `raw_os_error(&self) -> Option<i32>`

I'm also not using `ErrorKind` anymore, and the code is much simpler.

> But what exactly for it will be used? Outside of formatting I don't quite see use-cases for distinguishing OS and custom error.

The main reason would be matching on something. We explicitly don't want to have a stable way to match on Custom errors, but we would want a way to match on OS errors so a user can trap on platform specific behavior (otherwise we might as well just have an opaque error type).

src/ios.rs
dhardy: The motivation for redundant messages is https://github.com/rust-random/rand/issues/837 — i.e. they may not be redundant. But of course we're working on solving that...

src/error.rs
```diff
@@ -5,78 +5,166 @@
...
+    pub fn raw_os_error(&self) -> Option<i32> {
+        self.try_os_error().ok()
+    }
+
+    /// Extract the bare error code.
     ///
-    /// One may attempt to format this error via the `Display` implementation.
+    /// This code can either come from the underlying OS, or be a custom error.
+    /// Use [`raw_os_error()`] to disambiguate.
     pub fn code(&self) -> NonZeroU32 {
```
dhardy: Worth adding `#[inline]` here and e.g. to `raw_os_error`? (Without it, inlining is not possible across crate boundaries.)

src/error.rs
```diff
@@ -5,78 +5,166 @@
...
+        Self(NonZeroU32::new(errno as u32).unwrap())
+    }
 
-    /// No generator is available.
-    pub const UNAVAILABLE: Error = Error(unsafe { NonZeroU32::new_unchecked(CODE_UNAVAILABLE) });
+    /// Crate a custom error in the provided block (group of 2^16 error codes).
+    /// The provided block must not be negative, and block 0 is reserved for
+    /// custom errors in the `getrandom` crate.
+    pub fn custom_error(block: i16, code: u16) -> Self {
+        assert!(block >= 0);
```
dhardy: Why use `i16` when you advertise that `2 << 16` blocks are available? IMO there's no reason not to use a `u16` and allow use of all blocks here.

josephlr: So there are only 2^15 blocks available. We need 2^31 - 1 values for the OS errors, and each block is of length 2^16. This leaves only 2^31 value left, so 2^15 blocks. 

Saying "non-negative `i16`" seemed simpler than "`u16` less than 2^15"

dhardy: Your maths are off. `(2^32 - 1) - (2^31 - 1) == (2^32 - 2^31) - (1 - 1) == 2^31`.

josephlr:
> * add fn wrappers: `from_os_error, custom_error, raw_os_error` (though the latter may have to check for both boxed codes and a boxed `io::Error`; maybe we don't need it)

One thing we might consider is having the `rng_core::Error` be a wrapper around `std::io::Error` when using `std` feature. Then having `raw_os_error` would make sense for both, and we wouldn't have to do our own error boxing.

If we did that, then `code(&self) -> Option<NonZeroU32>` would also probobly need to be deprecated along with `From<NonZeroU32>` (I feel like these methods go together and should either both stay or both go).

src/error.rs
dhardy: There's one other issue with `i16`: using `block as u32` is a bit funky since it uses signed extension. Probably just write `block as u16 as u32` then.

newpavlov:
@josephlr 
>We explicitly don't want to have a stable way to match on Custom errors

But why? What is motivation for it? Especially if we'll state that non-OS error codes are unstable and may be changed in future.

Imagine an embedded device with a custom `getrandom` implementation. IIUC your approach creator of such platform-specific crate should use custom errors. He may specify in documentation that some error codes may require actions from user (e.g. RNG device must be powered-up).

src/error.rs
josephlr: Fixed. Added `#[inline]` to all of the `pub` methods, as we want to inline the bit-twiddling across crate boundaries.

src/error.rs
josephlr: Fixed

josephlr:
> Imagine an embedded device with a custom `getrandom` implementation. IIUC your approach, creator of such platform-specific crate should use custom errors. He may specify in documentation that some error codes may require actions from user (e.g. RNG device must be powered-up).

This is a good point, we definitely want to support this use-case. I think the current change works in this regard, as we don't make any backwards incompatible changes, and the changes we do make will work with whatever approach we go with in https://github.com/rust-random/rand/issues/837. I think that issue might be the best place to discuss breaking changes.

I've rebased this on top of the latest master, so the only large changes should just be in `error.rs` and `error_impl.rs`.

josephlr:
@dhardy @newpavlov I've rebased this and changed `util_libc.rs` to use the changes from https://github.com/rust-lang/libc/pull/1432, so that no raw `extern` declarations are needed anymore. Is there anything else you wanted me to do on this PR?

dhardy:
Sorry for the delay — I'd like to review again, but I *think* I'm happy with it. I am however waiting for @newpavlov to reply to my previous comment (about reserving blocks).

newpavlov:
Sorry for the late reply! I still don't see raison d'etre for `from_os_error`, `custom_error`, `internal`, `raw_os_error` and `try_os_error` instead of the simple `From<NonZeroU32>` and `code`. Even worse, you add a potential panic point by using `assert` in `from_os_error`. Compiler can't prove that you will not get a negative `errno`, so this check will be always present, while the current `NonZeroU32::new`-based code is totally panic-free. What exact problem are you trying to solve here? I don't see why we should forbid matching on custom or internal errors. Formatting can be improved without any issues separately from these API changes.

dhardy:
The way I see it, @newpavlov has a good point about avoiding the need for panic and eliminating `from_os_error` and `internal` in favour of `From<NonZeroU32>`, however there *is* value in having structure here: at least documentation suggesting what kind of errors RNGs should return and how error codes can (if necessary) be interpreted. `custom_error` is useful sugar for documenting this and making usage convenient, so I think it might as well stay.

The `raw_os_error` and `try_os_error` functions only really serve one use: converting an error code to the typical `i32`. `raw_os_error` *may* be useful to take advantage of OS-specific `errno` interpretation, but I think `try_os_error` has no other use — we should handle formatting of "custom" errors ourselves (probably something like `0x2F09-3`).

josephlr:
> I still don't see _raison d'etre_ for `from_os_error`, `custom_error`, `internal`, `raw_os_error` and `try_os_error` instead of the simple `From<NonZeroU32>` impl and `code` method. I think specifying in documentation that errors with code smaller than `1 << 31` should be interpreted as OS errors will be good enough

This makes sense, I removed `from_os_error`, `custom_error`, `internal`, and `try_os_error`. This makes the diff smaller and the code cleaner. I left in `raw_os_error` as it's a useful helper and could be used to disambiguate error types (as @dhardy mentioned).

I also just did away with the whole block nonsense. After looking at it with a fresh set of eyes it seems unnecessarily complex. The reservations are now just:
  - `[0, 1 << 31)` - positive `i32`s - OS Errors
  - `[1 << 31, 1 << 31 + 1 << 30)` - codes for `getrandom`, `rand`, etc...
  - `[1 << 31 + 1 << 30, 1 << 32)` - user definable codes

The PR description and docs comments have also been updated/clarified. The `Debug` and `Display` implementations are slightly different, and the `RTL_GEN_RANDOM_FAILED` constants are now just error constants.

src/wasi.rs
```diff
@@ -8,13 +8,14 @@
...
+use core::num::NonZeroU32;
 
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
     let ret =
         unsafe { libc::__wasi_random_get(dest.as_mut_ptr() as *mut libc::c_void, dest.len()) };
-    if ret != 0 {
-        error!("WASI: __wasi_random_get: failed with {}", ret);
-        Err(Error::from_os_error(ret.into()))
+    if let Some(code) = NonZeroU32::new(ret as u32) {
+        error!("WASI: __wasi_random_get: failed with {}", errno);
```
dhardy: Should still be `ret` I think?

josephlr: Fixed

josephlr:
Note that adding in the `Error` constants required bumping the min rustc version to 1.33. Is that OK?

dhardy:
Is there an easy work-around for rustc 1.32? While *most* Rand users are happy to use the latest stable, there are a few less happy about it (for versions < 6 months old).

newpavlov:
What feature do we need from Rust 1.33? Is it about making `internal_error` a `const` function? In that case we can use explicit error codes for the time being.

I will try to review changes on this weekend.

dhardy:
I won't do anything before @newpavlov has a chance to review, but mostly I'm happy about this. Integration into `rand_core` looks fairly easy, but I have one question: which code range should we use for Rand errors?

Perhaps we should consider them *internal* and use from `0xD000_0000 = 1<<31 + 1<<30 + 1<<28` for Rand libs; this leaves plenty for `getrandom` usage and avoids confusion (since `getrandom` is still a `rust-random` lib for now).

josephlr:
> I won't do anything before @newpavlov has a chance to review, but mostly I'm happy about this. Integration into `rand_core` looks fairly easy, but I have one question: which code range should we use for Rand errors?
> 
> Perhaps we should consider them _internal_ and use from `0xD000_0000 = 1<<31 + 1<<30 + 1<<28` for Rand libs; this leaves plenty for `getrandom` usage and avoids confusion (since `getrandom` is still a `rust-random` lib for now).

My thought was to have all the codes in `[1 << 31, 1 << 31 + 1 << 30)` (i.e. between `Error::INTERNAL_START` and `Error::CUSTOM_START`) to be used by all of the Rand libs, `getrandom`, rand-core, rand-*, etc... 

I don't know how explicit you want to make the split between `getrandom` error code ranges and `rand` error code ranges. If you want to make it explicit, we can just split the "Internal" range at `1 << 31 + 1 << 29`, and say `0x8000_0000 - 0xA000_0000` is for `getrandom` and `0xA000_0000 - 0xC000_0000` is for the `rand` libs.

@newpavlov good idea, I'll just turn the function into a macro. The `unsafe` in `const fn` is the only thing this needs from 1.33.

src/openbsd.rs
```diff
@@ -7,24 +7,17 @@
...
-use std::io;
 
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
     for chunk in dest.chunks_mut(256) {
         let ret = unsafe { libc::getentropy(chunk.as_mut_ptr() as *mut libc::c_void, chunk.len()) };
         if ret == -1 {
+            let err = last_os_error();
             error!("libc::getentropy call failed");
-            return Err(io::Error::last_os_error().into());
+            return Err(err);
```
newpavlov: Since we don't use error code in debug message, I think we can save one line by writing `Err(last_os_error())` here and in some other places.

src/util_libc.rs
```diff
@@ -5,12 +5,28 @@
...
+        use libc::___errno as errno_location;
+    } else if #[cfg(any(target_os = "macos", target_os = "freebsd", target_os = "dragonfly"))] {
+        use libc::__error as errno_location;
+    } else if #[cfg(target_os = "haiku")] {
+        use libc::_errnop as errno_location;
+    }
+}
+
+pub fn last_os_error() -> Error {
+    Error::from(NonZeroU32::new(unsafe { *errno_location() } as u32).unwrap())
```
newpavlov: Potential panic point. Maybe write something like:
 ```rust
let errno = unsafe { *errno_location() } as u32;
NonZeroU32::new(errno).map_or(Error::from, Error::ZERO_ERRNO)
```
?

File: src/error.rs [hunk 11]
```diff
@@ -5,78 +5,141 @@
...
-        Error(code)
+        Self(code)
     }
 }
 
-impl From<&Error> for Error {
-    fn from(error: &Error) -> Self {
-        *error
+/// Internal Error constants
+pub(crate) const UNSUPPORTED: Error = internal_error(0);
```
newpavlov: I am a bit unsure about keeping these constants private, especially `UNSUPPORTED`. But I guess we can make them public later without any issues if such need will arise.

src/error.rs
```diff
@@ -5,78 +5,141 @@
...
 impl Error {
-    /// An unknown error.
-    pub const UNKNOWN: Error = Error(unsafe { NonZeroU32::new_unchecked(CODE_UNKNOWN) });
+    /// Codes below this point represent OS Errors (i.e. positive i32 values).
+    /// Codes at or above this point, but below [`Error::CUSTOM_START`] are
+    /// reserved for use by the `rand` and `getrandom` crates.
+    pub const INTERNAL_START: u32 = 1 << 31;
 
-    /// No generator is available.
-    pub const UNAVAILABLE: Error = Error(unsafe { NonZeroU32::new_unchecked(CODE_UNAVAILABLE) });
```
newpavlov: I think we can keep these constants, but mark them deprecated. (yeah, it's somewhat pedantic, but no harm in doing that) For error code you can simply use `Error::UNSUPPORTED` for both constants.

src/openbsd.rs
josephlr: So this was intentional, we want to make sure that any `last_os_error` is always called directly after the relevant function. Something like `error!` might overwrite `errno`.

src/error.rs
josephlr: Seems reasonable for now, not exporting them gives us more flexibility about how/if we want to export them in the future.

src/error.rs
josephlr: Fixed.

src/util_libc.rs
josephlr: I did something slightly different, so that we would also catch if the `errno` is negative. The compiler now can [figure everything out, and remove the `unwrap()` code](https://rust.godbolt.org/z/TeEtmN).

newpavlov:
Great! Thank you!

newpavlov:
One small thing which bother me a bit is `$n as u16 as u32` cast in the `internal_error` macro. Is it intentional or just a leftover from the old code?

dhardy:
I guess it's just to ensure than no constants are too large? It's harmless anyway.

josephlr:
> I guess it's just to ensure than no constants are too large? It's harmless anyway.

Yes, before the function just took a u16, to ensure that the `NonZeroU32::new_unchecked` call is sound. This was just the easiest thing to do the same thing in a macro.

PR #55, opened 2019-07-05
Use cfg-if to simplify lib.rs

josephlr:
This crate uses `cfg-if` to simplify the target switching in `getrandom`. It's already a dependency of `libstd`, so this does not hurt the goals of https://github.com/rust-lang/rust/pull/62082

This change also stops trying to use `errno` (or equivalent) on Windows and iOS. The docs for these functions make no mention of error indications other than their return values. It also correctly handles errno on WASI (it's just the positive returned value, like `cloudabi`).

This removes the `std::io` dependency for these targets.

It also fixes a few `unused` warnings.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ members = ["tests/wasm_bindgen"]
 
 [dependencies]
 log = { version = "0.4", optional = true }
+cfg-if = "0.1"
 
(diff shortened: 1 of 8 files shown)


newpavlov:
I think `Error::UNKNOWN` is fine. Though maybe it may be worth to add description to `error_msg_inner` which will overwrite the default. We can handle it in a separate PR as `error_msg_inner` could use work for other targets as well.

PR #57, opened 2019-07-07
Freebsd: Try getrandom() first

### LINKED ISSUE #35: Use getrandom on FreeBSD

josephlr:
Depends on #54 fixes #35 

Implementation is quite similar to the Solaris/Illumos implementation. @myfreeweb PTAL

Edit: Also adds a `fill_exact` helper method for making repeated OS calls to fill a buffer. This let's us simplify the Linux/Solaris/FreeBSD code, as well as adding some changes useful for #54 and #58

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,7 +16,7 @@
 //! | Windows          | [`RtlGenRandom`][3]
 //! | macOS            | [`getentropy()`][19] if available, otherise [`/dev/random`][20] (identical to `/dev/urandom`)
 //! | iOS              | [`SecRandomCopyBytes`][4]
-//! | FreeBSD          | [`kern.arandom`][5]
+//! | FreeBSD          | [`getrandom()`][21] if available, otherise [`kern.arandom`][5]
 //! | OpenBSD          | [`getentropy`][6]
 //! | NetBSD           | [`/dev/urandom`][7] after reading from `/dev/random` once
 //! | Dragonfly BSD    | [`/dev/random`][8]
@@ -118,6 +118,7 @@
 //! [18]: https://software.intel.com/en-us/articles/intel-digital-random-number-generator-drng-software-implementation-guide
 //! [19]: https://www.unix.com/man-page/mojave/2/getentropy/
 //! [20]: https://www.unix.com/man-page/mojave/4/random/
+//! [21]: https://www.freebsd.org/cgi/man.cgi?query=getrandom&manpath=FreeBSD+12.0-stable
 
 #![doc(
     html_logo_url = "https://www.rust-lang.org/logos/rust-logo-128x128-blk.png",
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,9 +14,9 @@
 //! |------------------|---------------------------------------------------------
 //! | Linux, Android   | [`getrandom`][1] system call if available, otherwise [`/dev/urandom`][2] after reading from `/dev/random` once
 //! | Windows          | [`RtlGenRandom`][3]
-//! | macOS            | [`getentropy()`][19] if available, otherise [`/dev/random`][20] (identical to `/dev/urandom`)
+//! | macOS            | [`getentropy()`][19] if available, otherwise [`/dev/random`][20] (identical to `/dev/urandom`)
 //! | iOS              | [`SecRandomCopyBytes`][4]
-//! | FreeBSD          | [`getrandom()`][21] if available, otherise [`kern.arandom`][5]
+//! | FreeBSD          | [`getrandom()`][21] if available, otherwise [`kern.arandom`][5]
 //! | OpenBSD          | [`getentropy`][6]
 //! | NetBSD           | [`/dev/urandom`][7] after reading from `/dev/random` once
 //! | Dragonfly BSD    | [`/dev/random`][8]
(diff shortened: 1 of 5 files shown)


myfreeweb:
Looks good, works fine.

newpavlov:
Maybe it's worth to make this PR independent from #54? I think it will be some time before we will converge our opinions on that one, while changes in this PR look quite straightforward.

josephlr:
> Maybe it's worth to make this PR independent from #54? I think it will be some time before we will converge our opinions on that one, while changes in this PR look quite straightforward.

The main reason for the change is that this implementation uses the `fill_exact` helper method introduced in #54. I can put those changes in this PR if you want.

newpavlov:
I do not insist on it, you can leave it as-is. I was simply thinking it may be easier for you, since no syncing with #54 will be needed.

josephlr:
> I was simply thinking it may be easier for you, since no syncing with #54 would be needed.

Good point, I added https://github.com/rust-random/getrandom/pull/57/commits/fff1bded4be446ec018a970a979235583ae7ff13 to this PR that adds in the helper function changes.

src/util_libc.rs
```diff
@@ -5,9 +5,30 @@
...
 use crate::util::LazyUsize;
+use crate::Error;
 use core::ptr::NonNull;
+use std::io;
+
+pub fn fill_exact(mut buf: &mut [u8], f: impl Fn(&mut [u8]) -> libc::ssize_t) -> Result<(), Error> {
+    while !buf.is_empty() {
+        let res = f(buf);
+        if res < 0 {
+            let err = io::Error::last_os_error();
```
dhardy: Seems odd to me to assume that calling `last_os_error` is appropriate here given the function name. At least this should be mentioned by function documentation.

src/lib.rs
```diff
@@ -16,7 +16,7 @@
 //! | Windows          | [`RtlGenRandom`][3]
 //! | macOS            | [`getentropy()`][19] if available, otherise [`/dev/random`][20] (identical to `/dev/urandom`)
 //! | iOS              | [`SecRandomCopyBytes`][4]
-//! | FreeBSD          | [`kern.arandom`][5]
+//! | FreeBSD          | [`getrandom()`][21] if available, otherise [`kern.arandom`][5]
```
newpavlov: Typo: otherise -> otherwise

src/solaris_illumos.rs
```diff
@@ -37,11 +34,9 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
...
         // derived platforms for atomically obtaining random data.
         for chunk in dest.chunks_mut(256) {
-            let ret = unsafe { func(chunk.as_mut_ptr(), chunk.len(), 0) };
-            if ret != chunk.len() as _ {
-                error!("getrandom syscall failed with ret={}", ret);
-                return Err(io::Error::last_os_error().into());
-            }
+            fill_exact(chunk, |buf| unsafe {
+                func(buf.as_mut_ptr(), buf.len(), 0) as libc::ssize_t
+            })?
```
newpavlov: I think `fill_exact` is redundant here.

src/util_libc.rs
```diff
@@ -5,9 +5,30 @@
...
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
+extern crate std;
 
 use crate::util::LazyUsize;
+use crate::Error;
 use core::ptr::NonNull;
+use std::io;
+
+pub fn fill_exact(mut buf: &mut [u8], f: impl Fn(&mut [u8]) -> libc::ssize_t) -> Result<(), Error> {
```
newpavlov: Why `libc::ssize_t` instead of `isize`?

src/util_libc.rs
newpavlov: During removal of `std` dependency it will have to be replaced with an explicit `errno`, so I think it's fine to keep it as-is for the time being.

src/solaris_illumos.rs
newpavlov: Ah, never mind, [docs](https://docs.oracle.com/cd/E88353_01/html/E37841/getrandom-2.html) does not specify that it's guaranteed to fill the whole buffer if it's length is less or equal to 256 bytes.

src/util_libc.rs
josephlr: So I changed the function to be named `sys_fill_exact` and added a comment to explain the expected semantics of the `sys_fill` function. @newpavlov is right that this will be replaced with an explicit call to an errno function in #54

src/lib.rs
josephlr: Fixed here and elsewhere

src/util_libc.rs
josephlr: This is mainly because all the functions I'm using this with `getrandom(2)`, `read(2)`, etc.. All return a `libc::ssize_t`, so using that type seemed less confusing. As you correctly point out, it doesn't really matter except for code clarity.

PR #58, opened 2019-07-07
Remove libstd dependancy for Opening and Reading files

josephlr:
Depends on #54 so only look at the last commit if you want to just review this PR.

This PR removes the last part of `libstd` we were depending on (when not using the `std` feature). Specifically:

- We now manually open files with `libc::open(path.as_ptr(), libc::O_RDONLY | libc::O_CLOEXEC)`
  - We don't care about Linux kernels too old to support `libc::O_CLOEXEC` (kernel < 2.6.22)
- All the filenames now have to be null-terminated
- Instead of reading one byte from `/dev/random` we now [`poll(2)`](http://man7.org/linux/man-pages/man2/poll.2.html) on the `/dev/random` file descriptor
  - This still lets us know if `/dev/urandom` is safe
  - This avoids ever actually reading from `/dev/random`
  - This is what [libsodium does](https://github.com/jedisct1/libsodium/blob/60f4bc82120626458183cd1f9902908319af3d5d/src/libsodium/randombytes/sysrandom/randombytes_sysrandom.c#L151-L174)
  - See existing discussion in https://github.com/briansmith/ring/issues/558

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 A Rust library for retrieving random data from (operating) system source. It is
 assumed that system always provides high-quality cryptographically secure random
 data, ideally backed by hardware entropy sources. This crate derives its name
-from Linux's `getrandom` function, but is cross platform, aiming to support
+from Linux's `getrandom` function, but is cross platform, roughly supporting
 the same set of platforms as Rust's `std` lib.
 
 This is a low-level API. Most users should prefer using high-level random-number
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 A Rust library for retrieving random data from (operating) system source. It is
 assumed that system always provides high-quality cryptographically secure random
 data, ideally backed by hardware entropy sources. This crate derives its name
-from Linux's `getrandom` function, but is cross platform, roughly supporting
+from Linux's `getrandom` function, but is cross platform, aiming to support
 the same set of platforms as Rust's `std` lib.
 
 This is a low-level API. Most users should prefer using high-level random-number
@@ -40,7 +40,10 @@ fn get_random_buf() -> Result<[u8; 32], getrandom::Error> {
 
 ## Features
 
-This library is `no_std` compatible, but uses `std` on most platforms.
+This library is `no_std` for every supported target. However, getting randomness
+usually requires calling some external system API. This means most platforms
+will require linking against system libraries (i.e. `libc` for Unix,
+`Advapi32.dll` for Windows, Security framework on iOS, etc...).
 
 The `log` library is supported as an optional dependency. If enabled, error
 reporting will be improved on some platforms.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -152,22 +152,8 @@ mod util;
 #[allow(dead_code)]
 mod util_libc;
 
-// std-only trait definitions (also need for use_file)
-#[cfg(any(
-    feature = "std",
-    target_os = "android",
-    target_os = "dragonfly",
-    target_os = "emscripten",
-    target_os = "freebsd",
-    target_os = "haiku",
-    target_os = "illumos",
-    target_os = "linux",
-    target_os = "macos",
-    target_os = "netbsd",
-    target_os = "openbsd",
-    target_os = "redox",
-    target_os = "solaris",
-))]
+// std-only trait definitions
+#[cfg(feature = "std")]
 mod error_impls;
 
 // These targets read from a file as a fallback method.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,13 +12,13 @@
 //!
 //! | OS               | interface
 //! |------------------|---------------------------------------------------------
-//! | Linux, Android   | [`getrandom`][1] system call if available, otherwise [`/dev/urandom`][2] after reading from `/dev/random` once
+//! | Linux, Android   | [`getrandom`][1] system call if available, otherwise [`/dev/urandom`][2] after successfully polling `/dev/random`
 //! | Windows          | [`RtlGenRandom`][3]
 //! | macOS            | [`getentropy()`][19] if available, otherwise [`/dev/random`][20] (identical to `/dev/urandom`)
 //! | iOS              | [`SecRandomCopyBytes`][4]
 //! | FreeBSD          | [`getrandom()`][21] if available, otherwise [`kern.arandom`][5]
 //! | OpenBSD          | [`getentropy`][6]
-//! | NetBSD           | [`/dev/urandom`][7] after reading from `/dev/random` once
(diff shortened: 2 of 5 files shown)


briansmith:
> * Suggested by @briansmith in [briansmith/ring#558](https://github.com/briansmith/ring/issues/558)

As you can read in the linked *ring* issue, I made so such suggestion.

josephlr:
> > * Suggested by @briansmith in [briansmith/ring#558](https://github.com/briansmith/ring/issues/558)
> 
> As you can read in the linked _ring_ issue, I made so such suggestion.

Whoops sorry about that, meant to say you pointed out the libsodium implementation, changed to be more accurate.

dhardy:
I guess it's time for a rebase — however I believe the only motivation is https://github.com/rust-lang/rust/pull/62082, thus ideally this should wait on a decision whether we should go ahead with that (except that decision may depend on the existence of a viable PR...)

josephlr:
@dhardy I rebased the PR, and the CI is passing. I agree that if we don't do https://github.com/rust-lang/rust/pull/62082, we should not merge this PR.

src/util_libc.rs
```diff
@@ -116,3 +116,21 @@ impl LazyFd {
...
+        use libc::open64 as open;
+    } else {
+        use libc::open;
+    }
+}
+
+// SAFETY: path must be null terminated, FD must be manually closed.
+pub unsafe fn open_readonly(path: &str) -> Option<libc::c_int> {
+    // We don't care about Linux OSes too old to support O_CLOEXEC.
+    let fd = open(path.as_ptr() as *mut _, libc::O_RDONLY | libc::O_CLOEXEC);
```
newpavlov: It may be worth to add a `debug_assert!` which will check that `path` is null terminated. Of course it's not necessary, but if it's possible to check safety conditions, personally I usually add such asserts to be extra safe.

>We don't care about Linux OSes too old to support O_CLOEXEC.

Strictly speaking Rust supports Linux kernel from [2.16.18 and later](https://forge.rust-lang.org/platform-support.html), while `O_CLOEXEC` is available only from 2.6.23. To be safe I think we should run it by the lang team. I will try to ask about this in the rust-lang PR.

newpavlov:
You also have to disable default features for `libc`, otherwise `getrandom` will not be `std-independent.`

josephlr:
> You also have to disable default features for `libc`, otherwise `getrandom` will not be `std`-independent.

Done

README.md
```diff
@@ -10,7 +10,7 @@
 A Rust library for retrieving random data from (operating) system source. It is
 assumed that system always provides high-quality cryptographically secure random
 data, ideally backed by hardware entropy sources. This crate derives its name
-from Linux's `getrandom` function, but is cross platform, roughly supporting
+from Linux's `getrandom` function, but is cross platform, aiming to support
```
dhardy: What does *support* mean — that *it works* or that *it's a bug if it doesn't work*? My interpretation is the latter, thus `aiming to` only adds confusion.

Also — there was never a decision *not* to try to support additional targets (which is what #4 is about), hence why this did say *roughly*.

josephlr: I reverted this `README` change

josephlr:
As https://github.com/rust-lang/rust/issues/62516 might take awhile to reslove. I just added in the RHEL 5 compat code. It only takes 2 lines.

newpavlov:
So I think we can merge this without waiting for decision regarding rust-lang/rust#62082, if `getrandom` will not be used as part of `std` we may revert this change in a later release. But first I think we should decide what to do with the MSRV issue raised in #69.

josephlr:
I slightly changed the `use_file` implementation to use `cfg-if` to determine if we should `poll` `/dev/random`. That seemed less hacky than checking the value of `FILE_PATH`. I also rebased this onto master to make sure there were no merge issues.

> So I think we can merge this without waiting for decision regarding [rust-lang/rust#62082](https://github.com/rust-lang/rust/pull/62082).

Sounds reasonable to me.

newpavlov:
Can you also update the table in `lib.rs` and replace "after reading from `/dev/random` once" to something like "after polling `/dev/random` once"?

josephlr:
> Can you also update the table in `lib.rs` and replace "after reading from `/dev/random` once" to something like "after polling `/dev/random` once"?

Done, I also fixed up the `file_init` loop where we poll, hopefully it's more clear now.

newpavlov:
I think you can simplify the loop even further like this:
```rust
let ret = loop {
    // A negative timeout means an infinite timeout.
    let res = unsafe { libc::poll(&mut pfd, 1, -1) };
    if res == 1 {
        break unsafe { open_readonly("/dev/urandom\0") };
    } else if res < 0 {
        let e = last_os_error().raw_os_error();
        if e == Some(libc::EINTR) || e == Some(libc::EAGAIN) {
            continue;
        }
    }
    break None;
};
unsafe { libc::close(pfd.fd) };
ret
```

josephlr:
> I think you can simplify the loop even further like this:

Done. I totally forgot that rust loops can return a value.

src/lib.rs
```diff
@@ -152,22 +152,8 @@ mod util;
...
-    target_os = "linux",
-    target_os = "macos",
-    target_os = "netbsd",
-    target_os = "openbsd",
-    target_os = "redox",
-    target_os = "solaris",
-))]
+// std-only trait definitions
+#[cfg(feature = "std")]
 mod error_impls;
```
dhardy: What happens in this case when the `getrandom` crate is used both by `std` and by other crates wanting `std` features?

dhardy:
> So I think we can merge this without waiting for decision regarding rust-lang/rust#62082, if getrandom will not be used as part of std we may revert this change in a later release. But first I think we should decide what to do with the MSRV issue raised in #69.

There may not be another way to go about this, so fair enough.

src/lib.rs
newpavlov: IIUC right now `getrandom` will be compiled into `std`, so enabling features for it during usual builds will not influence the copy of the crate used by `std`. Although I am not sure how it will work with rust-lang/rfcs#2663.

josephlr:
@newpavlov if we're going with the build script approach in #69, do we still want to wait on that PR before merging this one?

newpavlov:
No, I think we can merge it right away.

RalfJung:
> We don't care about Linux kernels too old to support libc::O_CLOEXEC (kernel < 2.6.22)

Wasn't the plan to use this crate in libstd? Rust officially [supports](https://forge.rust-lang.org/platform-support.html) Linux 2.6.18+.

mati865:
Description was not updated to reflect https://github.com/rust-random/getrandom/pull/58#issuecomment-516672866

PR #60, opened 2019-07-11
Incorperate feedback on Lazy helpers

josephlr:
In #52 @RalfJung noted that some improvements could be make to the implementations of `LazyUsize` and `LazyFd`. This PR implements those changes.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -53,7 +53,8 @@ impl LazyUsize {
 
     // Synchronously runs the init() function. Only one caller will have their
     // init() function running at a time, and exactly one successful call will
-    // be run. The init() function should never return LazyUsize::ACTIVE.
+    // be run. init() returning UNINIT or ACTIVE will be considered a failure,
+    // and future calls to sync_init will rerun their init() function.
     pub fn sync_init(&self, init: impl FnOnce() -> usize, mut wait: impl FnMut()) -> usize {
         // Common and fast path with no contention. Don't wast time on CAS.
         match self.0.load(Relaxed) {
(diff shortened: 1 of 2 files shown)


src/util.rs
```diff
@@ -65,6 +65,7 @@ impl LazyUsize {
             match self.0.compare_and_swap(Self::UNINIT, Self::ACTIVE, Relaxed) {
                 Self::UNINIT => {
                     let val = init();
+                    assert!(val != Self::ACTIVE);
```
newpavlov: As I wrote earlier in #52 I would prefer to use code like this:
```rust
let mut val = init();
if val == Self::ACTIVE { val = Self::UNINIT }
self.0.store(val, Relaxed);
```

josephlr: Any particular reason why? Wouldn't having `init()` returning `ACTIVE` be a bug? Or are you concerned with the `LazyUsize` getting "stuck" in the `ACTIVE` state after a panic?

newpavlov: Panicking pulls formatting machinery (unless some careful setup is done) and it results in a larger binary. I haven't thought about infinite loop (e.g. if user will catch the panic), but it's indeed another drawback.

I think we both agree that it's extremely unlikely that `init` will return `Self::ACTIVE`, so it does not really matter.

josephlr: I removed the assert and updated the docs. I used a slightly different implementation than above to maintain the invariant that `sync_init()` always returns the value of some run of `init()`.

src/util.rs
```diff
@@ -65,8 +66,13 @@ impl LazyUsize {
             match self.0.compare_and_swap(Self::UNINIT, Self::ACTIVE, Relaxed) {
                 Self::UNINIT => {
                     let val = init();
-                    assert!(val != Self::ACTIVE);
-                    self.0.store(val, Relaxed);
+                    self.0.store(
+                        match val {
+                            Self::UNINIT | Self::ACTIVE => Self::UNINIT,
+                            val => val,
+                        },
```
newpavlov: I think the following alternative will be a bit easier to read:
```rust
let val = match init() {
    Self::UNINIT | Self::ACTIVE => Self::UNINIT,
    val => val,
};
self.0.store(val, Relaxed);
```

newpavlov: Ah, nevermind. I see your comment earlier.

PR #64, opened 2019-07-20
Use wasi-core isntead of libc

newpavlov:
IIUC `wasi-core` will be the main crate for exposing low-level WASI Core API in Rust. We could link to the function directly as with some other targets, but probably we shouldn't do it, as right now `__wasi_random_get` is not considered stable. See: rust-lang/libc#1434

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ cfg-if = "0.1"
 [target.'cfg(any(unix, target_os = "redox"))'.dependencies]
 libc = "0.2.60"
 
-[target.'cfg(any(target_os = "wasi"))'.dependencies]
+[target.'cfg(target_os = "wasi")'.dependencies]
 wasi = "0.5"
 
 [target.wasm32-unknown-unknown.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,8 +21,11 @@ members = ["tests/wasm_bindgen"]
 log = { version = "0.4", optional = true }
 cfg-if = "0.1"
 
-[target.'cfg(any(unix, target_os = "redox", target_os = "wasi"))'.dependencies]
+[target.'cfg(any(unix, target_os = "redox"))'.dependencies]
 libc = "0.2.54"
+
+[target.'cfg(any(target_os = "wasi"))'.dependencies]
+wasi-core = "0.2"
 
 [target.wasm32-unknown-unknown.dependencies]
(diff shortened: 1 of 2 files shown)


dhardy:
Sounds like you know much more about WASI interfaces than myself.

The only thing I can say is that this adds another dependency — but if `wasi-core` becomes a dependency of `std` anyway this doesn't matter.

Since both the `libc` interface and the `wasi-core` interfaces are explicitly unstable there may not be much reason to choose one over the other for now — will breaking changes in each crate require bumping the minor version @sunfishcode?

sunfishcode:
As we thought about it more, the name "WASI core" has caused confusion in some places, so I've now renamed the "wasi-core" crate to "wasi". I'd also request holding off on merging this for just a while longer while we let the module naming scheme within the crate settle in a little more.

But beyond that, I am starting to see the role for this "wasi" crate emerge. It's a minimal and `no_std` crate which we can add as a dependency to libc, and which can track WASI API changes over time, possibly including adding backwards-compatibility interfaces to help smooth over API transitions.

So yeah, I think this PR here is good, just update the crate name and give it a little more time while we settle any remaining organizational issues :-).

newpavlov:
Done! I will wait for your word then before merging this PR. (BTW maybe it's worth to release an empty `wasi-core v0.3` with description "use wasi crate instead" or something similar?)

>the name "WASI core" has caused confusion in some places, so I've now renamed the "wasi-core" crate to "wasi"

Well, I thought `wasi-core` for WASI Core API was a pretty straightforward naming, but oh well. :)

Cargo.toml
```diff
@@ -21,9 +21,12 @@ members = ["tests/wasm_bindgen"]
 log = { version = "0.4", optional = true }
 cfg-if = "0.1"
 
-[target.'cfg(any(unix, target_os = "redox", target_os = "wasi"))'.dependencies]
+[target.'cfg(any(unix, target_os = "redox"))'.dependencies]
 libc = "0.2.60"
 
+[target.'cfg(any(target_os = "wasi"))'.dependencies]
```
josephlr: `any` isn't needed for a single condition here

newpavlov:
@sunfishcode
Can we merge this PR or do you plan to introduce changes to the `wasi` crate in the near future?

sunfishcode:
Sorry for being slow on my end; merging this sounds good to me.

PR #67, opened 2019-07-29
Prepare v0.1.7 release

newpavlov:
I think it will be some time before we will get a decision for #58, so I think it will be better to publish the current version.


josephlr:
Nice! This should unblock https://github.com/rust-lang/rust/pull/61393

PR #69, opened 2019-07-31
Add Windows UWP support

newpavlov:
`RtlGenRandom` is not accessible on UWP targets, so we have to handle them separately right now.

cc @chouquette

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,9 +22,13 @@ environment:
   matrix:
   - TARGET: x86_64-pc-windows-msvc
   - TARGET: i686-pc-windows-msvc
+  - TARGET: x86_64-uwp-windows-gnu
+    CHANNEL: nightly
+  - TARGET: i686-uwp-windows-gnu
+    CHANNEL: nightly
 install:
   - appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
-  - rustup-init.exe -y --default-host %TARGET%
+  - rustup-init.exe -y --default-host %TARGET% --default-toolchain %CHANNEL%
   - set PATH=%PATH%;C:\Users\appveyor\.cargo\bin
   - rustc -V
   - cargo -V
@@ -32,5 +36,5 @@ install:
 build: false
 
 test_script:
-  - cargo test
-  - cargo test --examples
+  - cargo test --features=uwp
+  - cargo test --examples --features=uwp
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,12 +20,14 @@ environment:
   CARGO_HTTP_CHECK_REVOKE: false
 
   matrix:
-  - TARGET: x86_64-pc-windows-msvc
-  - TARGET: i686-pc-windows-msvc
   - TARGET: x86_64-uwp-windows-gnu
     CHANNEL: nightly
   - TARGET: i686-uwp-windows-gnu
     CHANNEL: nightly
+  - TARGET: x86_64-pc-windows-msvc
+    CHANNEL: stable
+  - TARGET: i686-pc-windows-msvc
+    CHANNEL: stable
 install:
   - appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
   - rustup-init.exe -y --default-host %TARGET% --default-toolchain %CHANNEL%
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,14 +21,18 @@ environment:
 
   matrix:
   - TARGET: x86_64-pc-windows-msvc
+    CHANNEL: 1.32.0
+  - TARGET: x86_64-pc-windows-msvc
     CHANNEL: stable
+  - TARGET: x86_64-pc-windows-msvc
+    CHANNEL: nightly
   - TARGET: i686-pc-windows-msvc
-    CHANNEL: stable
-  # TODO: replace with UWP targets in future
+    CHANNEL: nightly
   - TARGET: x86_64-pc-windows-gnu
     CHANNEL: nightly
   - TARGET: i686-pc-windows-gnu
     CHANNEL: nightly
+
 install:
   - appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
   - rustup-init.exe -y --default-host %TARGET% --default-toolchain %CHANNEL%
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -39,5 +39,5 @@ install:
 build: false
 
 test_script:
-  - cargo test --features=uwp
-  - cargo test --examples --features=uwp
+  - cargo test
+  - cargo test --examples
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,14 +20,15 @@ environment:
   CARGO_HTTP_CHECK_REVOKE: false
 
   matrix:
-  - TARGET: x86_64-uwp-windows-gnu
-    CHANNEL: nightly
-  - TARGET: i686-uwp-windows-gnu
-    CHANNEL: nightly
   - TARGET: x86_64-pc-windows-msvc
     CHANNEL: stable
   - TARGET: i686-pc-windows-msvc
     CHANNEL: stable
+  # TODO: replace with UWP targets in future
+  - TARGET: x86_64-pc-windows-gnu
+    CHANNEL: nightly
+  - TARGET: i686-pc-windows-gnu
+    CHANNEL: nightly
 install:
   - appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
   - rustup-init.exe -y --default-host %TARGET% --default-toolchain %CHANNEL%
--- a/build.rs
+++ b/build.rs
@@ -4,9 +4,11 @@ use std::env;
 
 fn main() {
     let target = env::var("TARGET").expect("TARGET was not set");
-    if target.contains("uwp") {
+    if target.contains("-uwp-windows-") {
         // for BCryptGenRandom
         println!("cargo:rustc-link-lib=bcrypt");
+        // to work around unavailability of `target_vendor` on Rust 1.33
+        println!("cargo:rustc-cfg=getrandom_uwp");
     } else if target.contains("windows") {
         // for RtlGenRandom (aka SystemFunction036)
         println!("cargo:rustc-link-lib=advapi32");
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,17 @@
+#![deny(warnings)]
+
+use std::env;
+
+fn main() {
+    let target = env::var("TARGET").expect("TARGET was not set");
+    if target.contains("uwp") {
+        // for BCryptGenRandom
+        println!("cargo:rustc-link-lib=bcrypt");
+    } else if target.contains("windows") {
+        // for RtlGenRandom (aka SystemFunction036)
+        println!("cargo:rustc-link-lib=advapi32");
+    } else if target.contains("apple-ios") {
+        // for SecRandomCopyBytes and kSecRandomDefault
+        println!("cargo:rustc-link-lib=framework=Security");
+    }
+}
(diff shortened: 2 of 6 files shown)


build.rs
```diff
@@ -0,0 +1,12 @@
+#![deny(warnings)]
+
+use std::env;
+
+fn main() {
+    let target = env::var("TARGET").expect("TARGET was not set");
+    if target.contains("uwp") {
+        println!("cargo:rustc-link-lib=bcrypt");
+    } else if target.contains("apple-ios") {
+        println!("cargo:rustc-link-lib=framework=Security");
```
newpavlov: I am not sure if we need this bit for iOS, but to be safe I've followed the `libstd` [example](https://github.com/rust-lang/rust/blob/master/src/libstd/build.rs).

josephlr: If we're following `libstd`, shouldn't we also add `println!("cargo:rustc-link-lib=advapi32");` for `windows` targets? That's [where `RltGenRandom` comes from](https://docs.microsoft.com/en-us/windows/win32/api/ntsecapi/nf-ntsecapi-rtlgenrandom).

src/windows_uwp.rs
```diff
@@ -0,0 +1,42 @@
...
+
+const BCRYPT_USE_SYSTEM_PREFERRED_RNG: u32 = 0x00000002;
+
+extern "system" {
+    fn BCryptGenRandom(
+        hAlgorithm: *mut c_void,
+        pBuffer: *mut u8,
+        cbBuffer: u32,
+        dwFlags: u32,
+    ) -> i32;
```
josephlr: [`NTSTATUS` values](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/87fba13e-bf06-450e-83b1-9241dc81e781) are `u32` not `i32`.

src/windows_uwp.rs
```diff
@@ -0,0 +1,42 @@
...
+        let ret = unsafe {
+            BCryptGenRandom(
+                ptr::null_mut(),
+                chunk.as_mut_ptr(),
+                chunk.len() as u32,
+                BCRYPT_USE_SYSTEM_PREFERRED_RNG,
+            )
+        };
+        if let Some(code) = NonZeroU32::new(ret as u32) {
+            error!("BCryptGenRandom: failed with {}", ret);
```
josephlr: Encode the error value as hex (that's how `NTSTATUS` tables usually write them).

src/windows_uwp.rs
```diff
@@ -0,0 +1,42 @@
...
+    for chunk in dest.chunks_mut(u32::max_value() as usize) {
+        let ret = unsafe {
+            BCryptGenRandom(
+                ptr::null_mut(),
+                chunk.as_mut_ptr(),
+                chunk.len() as u32,
+                BCRYPT_USE_SYSTEM_PREFERRED_RNG,
+            )
+        };
+        if let Some(code) = NonZeroU32::new(ret as u32) {
```
josephlr: My main concern with this is that [`NSSTATUS` values use the entire `u32` space](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/87fba13e-bf06-450e-83b1-9241dc81e781). This means that success/informational statuses will look like OS errors, warning statuses will look like internal errors, and error statuses will look like custom errors. This is not ideal.

We should either:
  - Be more careful with how we map `NTSTATUS` to `Error`
  - Just use a single internal return value, like how we use `RTL_GEN_RANDOM_FAILED` for the normal Windows implementation.

josephlr: [`NTSTATUS` values should not be compared to `STATUS_SUCCESS` (i.e. 0) to check for success](https://docs.microsoft.com/en-us/windows-hardware/drivers/kernel/using-ntstatus-values). They should instead have their most-significant bit checked. High means failure, low means success.

newpavlov: Hm, this is really unfortunate... I am not sure if erasing information completely is better than using a lossy mapping. So one option is to simply zeroize the highest bit.

According to `BCryptGenRandom` [docs](https://docs.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom), it should return `STATUS_SUCCESS` (i.e. 0) if call was successful. To be safe I think it will be better to be conservative and interpret all other codes as errors, even if they were informational or warning.

build.rs
newpavlov: Hm, you are right. But I wonder why does it work without explicit linking...

josephlr: It's not need for building the library, and I would guess it's because any Rust Windows binary has some other dep that links `advapi32`, as it's the base DLL for most Windows system calls.

So this is more a exercise in being cautious.

newpavlov: To be safe linking of `advapi32` is added.

newpavlov:
I've added a correct processing of NTSTATUS codes. Now only codes which start with `0b11` are considered as errors. To make them fit into the rage of OS error code the highest bit is zeroized, potentially it may cause some confusion, but I think it's better than erasing error code completely and replacing it with a custom error.

src/lib.rs
```diff
@@ -216,6 +224,15 @@ cfg_if! {
         #[path = "solaris_illumos.rs"] mod imp;
     } else if #[cfg(target_os = "wasi")] {
         #[path = "wasi.rs"] mod imp;
+    } else if #[cfg(any(
+        // target_vendor was stabilized only in Rust 1.33
+        target = "i686-uwp-windows-gnu",
+        target = "x86_64-uwp-windows-gnu",
+        target = "aarch64-uwp-windows-msvc",
+        target = "x86_64-uwp-windows-msvc",
+        target = "i686-uwp-windows-msvc",
```
newpavlov: Looks like this will not work, see rust-lang/rust#63217. :/ So our choices are:
- Do not support UWP
- Bump MSRV to Rust 1.33
- Add `uwp` or `rust_1_33` feature.

josephlr: I think we should bump the MSRV to 1.33 (and release a new minor version), Rust 1.38 is coming out in a few days (Aug 12 I think), so bumping to 1.33 then seems reasonable.

newpavlov: Personally until something like rust-lang/rfcs#2495 lands I prefer a conservative position stating that MSRV bump should be considered a breaking change.

I think we also could move UWP check to the Windows branch, so MSRV bump will affect only Windows users.

@dhardy 
What dou you think?

build.rs
```diff
@@ -0,0 +1,17 @@
+#![deny(warnings)]
+
+use std::env;
+
+fn main() {
+    let target = env::var("TARGET").expect("TARGET was not set");
+    if target.contains("uwp") {
```
josephlr: We should check for `-uwp-` so this doesn't accidentally trigger if a target arch/os contains "uwp" as a sub-string.

src/lib.rs
josephlr: After looking at #75 and rust-lang/rfcs#2495, I think you're right. Incrementing the minimum supported version _should be considered a breaking change_. However, we can support UWP without bumping the MSRV, adding a feature, or dropping support for XP/Vista. We can just check in the build script.

[`libc` does this](https://github.com/rust-lang/libc/blob/master/build.rs) to maintain a comically low MSRV (Rust 1.13). They just check for the version in `build.rs` and then set `cfg`s appropriately.

So we would do the following:
```rust
    ...
    if target.contains("uwp") {
        println!("cargo:rustc-cfg=getrandom_uwp");
        // for BCryptGenRandom
        println!("cargo:rustc-link-lib=bcrypt");
    ...
```

```rust
// src/lib.rs
    ...
    } else if #[cfg(all(windows, getrandom_uwp))] {
        #[path = "windows_uwp.rs"] mod imp;
    } else if #[cfg(windows)] {
        #[path = "windows.rs"] mod imp;
    }
    ...
```

This prevents exposing any of the details in `getrandom`'s public interface.

src/windows_uwp.rs
```diff
@@ -0,0 +1,55 @@
...
+        };
+        // NTSTATUS codes use two highest bits for severity codes
+        match ret >> 30 {
+            0b01 => info!("BCryptGenRandom: information code 0x{:08X}", ret),
+            0b10 => warn!("BCryptGenRandom: warning code 0x{:08X}", ret),
+            0b11 => {
+                error!("BCryptGenRandom: failed with 0x{:08X}", ret);
+                // We zeroize the highest bit, so the error code will reside
+                // inside the range designated for OS codes.
+                let code = ret & (u32::MAX >> 1);
```
josephlr: This was initially confusing to me, as I expected something like `ret ^ (1 << 31)`. However, this might just be personal preference, so feel free to leave as-is.

src/lib.rs
newpavlov: Good workaround!

src/windows_uwp.rs
newpavlov: I am simply more used to thinking in terms of masks. :) But your variant is closer to the comment, so I will use it instead.

src/lib.rs
dhardy: Using a build.rs file seems appropriate in this case.

newpavlov:
So I think we are good to merge this?

appveyor.yml
```diff
@@ -21,10 +21,17 @@ environment:
 
   matrix:
   - TARGET: x86_64-pc-windows-msvc
+    CHANNEL: stable
   - TARGET: i686-pc-windows-msvc
+    CHANNEL: stable
+  # TODO: replace with UWP targets in future
```
josephlr: Is there no way to just check if the crate will build on UWP? Like we do with Fuchsia, Redox, UEFI, etc...

src/windows_uwp.rs
```diff
@@ -0,0 +1,55 @@
...
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Implementation for Windows UWP targets. After deprecation of Windows XP
+//! and Vista, this can superseed the `RtlGenRandom`-based implementation.
+use crate::Error;
+use core::{ffi::c_void, num::NonZeroU32, ptr, u32};
```
josephlr: Do we need `u32` here? We don't have this in `windows.rs`, which also uses `u32::max_value()`.

appveyor.yml
mati865: You'd have to build your own libstd for this target using Xargo because there are no binaries for it.
I have no idea how much effort it would require.

josephlr: We [currently use `xbuild`](https://github.com/rust-random/getrandom/blob/98934cb4ab9438a9fec76fa22c0f3154890c2bc6/.travis.yml#L137-L139) to work with targets not yet shipping with `rustup`.

Running `cargo xbuild --target=x86_64-uwp-windows-gnu` works on my system, so we should (at the very least) add that and `i686-uwp-windows-gnu` to the travis script.

src/windows_uwp.rs
newpavlov: Ah, I forgot to remove this leftover from the old mask code.

dhardy:
BTW despite the request for review I would prefer to stay out of this one — I know little about the APIs.

newpavlov:
@dhardy 
Can you approve the new appveyor config? With it we now test `x86_64-pc-windows-msvc` on 1.32, stable and nightly; while `i686-pc-windows-msvc`, `x86_64-pc-windows-gnu` and `i686-pc-windows-gnu` are tested only on nightly.

dhardy:
Assuming none of those targets require nightly features, I think it would be more appropriate to use stable and/or beta for some of these tests? Otherwise looks fine (though I'm not very familiar with appveyor).

newpavlov:
In my understanding nightly CI tests are mainly used to catch potential regressions in Rust early to have a chance to fix them before they will hit beta/stable.

PR #71, opened 2019-08-02
Add dummy feature

### LINKED ISSUE #49: Stop using dummy implementation

newpavlov:
Closes #49 

Strictly speaking it may be considered a breaking change. We could make `dummy` feature enabled-by-default, but I am not sure if we should.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,3 +30,7 @@ stdweb = { version = "0.4.18", optional = true }
 
 [features]
 std = []
+# enables dummy implementation for wasm32-unknown-unknown
+wasm_dummy = []
+# enables dummy implementation for unsupported targets
+dummy = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,5 @@ stdweb = { version = "0.4.18", optional = true }
 
 [features]
 std = []
-# enables dummy implementation for wasm32-unknown-unknown
-wasm_dummy = []
 # enables dummy implementation for unsupported targets
 dummy = []
--- a/README.md
+++ b/README.md
@@ -59,9 +59,7 @@ for such target, you can either:
 - Use [`[replace]`][replace] or [`[patch]`][patch] section in your `Cargo.toml`
 to switch to a custom implementation with a support of your target.
 - Enable the `dummy` feature to have getrandom use an implementation that always
-fails on unsupported targets.
-
-We also accept pull requests to support additional targets.
+fails at run-time on unsupported targets.
 
 [replace]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-replace-section
 [patch]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-patch-section
--- a/README.md
+++ b/README.md
@@ -45,11 +45,16 @@ This library is `no_std` compatible, but uses `std` on most platforms.
 The `log` library is supported as an optional dependency. If enabled, error
 reporting will be improved on some platforms.
 
-For WebAssembly (`wasm32`), WASI and Emscripten targets are supported directly; otherwise
-one of the following features must be enabled:
+For `wasm32-unknown-unknown` target one of the following features must be
+enabled:
 
 -   [`wasm-bindgen`](https://crates.io/crates/wasm_bindgen)
 -   [`stdweb`](https://crates.io/crates/stdweb)
+
+By default compiling crate for an unsupported platform will result in a
+compilation error. You may either replace this crate with a custom one, which
+will include support for your target, or enable `dummy` feature to use an
+always erroring dummy implementation.
 
 ## Minimum Supported Rust Version
 
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@ library like [`rand`].
 
 [`rand`]: https://crates.io/crates/rand
 
-
 ## Usage
 
 Add this to your `Cargo.toml`:
@@ -45,21 +44,28 @@ This library is `no_std` compatible, but uses `std` on most platforms.
 The `log` library is supported as an optional dependency. If enabled, error
 reporting will be improved on some platforms.
 
-For `wasm32-unknown-unknown` target one of the following features must be
+For the `wasm32-unknown-unknown` target, one of the following features should be
 enabled:
 
 -   [`wasm-bindgen`](https://crates.io/crates/wasm_bindgen)
 -   [`stdweb`](https://crates.io/crates/stdweb)
 
-By default compiling crate for an unsupported platform will result in a
-compilation error. You may either replace this crate with a custom one, which
-will include support for your target, or enable `dummy` feature to use an
-always erroring dummy implementation.
+By default, compiling `getrandom` for an unsupported target will result in
+a compilation error. If want to build an application which uses `getrandom` for
+such target, you can either:
+- Use [`[replace]`][replace] or [`[patch]`][patch] section in your `Cargo.toml`
+to switch to a custom implementation with a support of your target.
+- Enable the `dummy` feature to have getrandom use an implementation that always
+fails on unsupported targets.
+
+We also accept pull requests to support additional targets.
+
+[replace]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-replace-section
+[patch]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-patch-section
 
 ## Minimum Supported Rust Version
 
 This crate requires Rust 1.32.0 or later.
-
 
 # License
 
@@ -69,4 +75,3 @@ The `getrandom` library is distributed under either of
  * [MIT license](LICENSE-MIT)
 
 at your option.
-
--- a/README.md
+++ b/README.md
@@ -51,8 +51,8 @@ enabled:
 -   [`stdweb`](https://crates.io/crates/stdweb)
 
 By default, compiling `getrandom` for an unsupported target will result in
-a compilation error. If want to build an application which uses `getrandom` for
-such target, you can either:
+a compilation error. If you want to build an application which uses `getrandom`
+for such target, you can either:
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -45,12 +45,17 @@ This library is `no_std` compatible, but uses `std` on most platforms.
 The `log` library is supported as an optional dependency. If enabled, error
 reporting will be improved on some platforms.
 
-For WebAssembly (`wasm32`), WASI and Emscripten targets are supported directly; otherwise
-one of the following features must be enabled:
+For `wasm32-unknown-unknown` target one of the following features must be
```
josephlr: For the `wasm32-unknown-unknown` target,

README.md
```diff
@@ -45,12 +45,17 @@ This library is `no_std` compatible, but uses `std` on most platforms.
...
+enabled:
 
 -   [`wasm-bindgen`](https://crates.io/crates/wasm_bindgen)
 -   [`stdweb`](https://crates.io/crates/stdweb)
 
+By default compiling crate for an unsupported platform will result in a
```
josephlr: By default, compiling this crate

Cargo.toml
```diff
@@ -30,3 +30,7 @@ stdweb = { version = "0.4.18", optional = true }
 
 [features]
 std = []
+# enables dummy implementation for wasm32-unknown-unknown
```
josephlr: Why have `wasm_dummy` and `dummy`. It seems like it would be much better to just have a single `dummy` feature.

README.md
```diff
@@ -45,12 +45,17 @@ This library is `no_std` compatible, but uses `std` on most platforms.
...
+compilation error. You may either replace this crate with a custom one, which
```
josephlr: I think the last sentance should be replace by something like:

> If your application needs to build on targets that are not supported by `getrandom`, you can either:
>  - Use [`cfg` attributes](https://doc.rust-lang.org/rust-by-example/attribute/cfg.html) to switch to a custom implementation on unsupported targets. 
>  - Enable the `dummy` feature to have `getrandom` use an implementation that always fails on unsupported targets.
>
> We also accept pull requests to support additional targets.

This makes it clearer _why_ someone would want to do this.

src/lib.rs
```diff
@@ -231,12 +233,24 @@ cfg_if! {
...
+                ");
             }
         }
-    } else {
+    } else if #[cfg(feature = "dummy")] {
         #[path = "dummy.rs"] mod imp;
+    } else {
+        compile_error!("\
+            target is not supported, you may enable dummy implementation \
+            using `dummy` feature or replace `getrandom` crate with a custom \
```
josephlr: I'd remove/rephrase the reference to the custom crate. Right now the "replace" message is a little confusing.

Cargo.toml
newpavlov: `wasm_dummy` is mainly intended for the `std` inclusion PR. We want use a constant seed only on `wasm32-unknown-unknown` and get a compilation error for all other unsupported targets. Since we can't enable features based on targets, I think using a separate `wasm_dummy` feature will be the most straightforward approach.

josephlr: Ok, so part of my confusion was that I assumed that the implementation for `std` integration would look like:

```rust
fn get_keys() -> [u64; 2] {
    let mut buf = [0u8; 16];

    #[cfg(not(target = "wasm32-unknown-unknown"))]
    getrandom::getrandom(&mut buf).expect("getrandom failure");

    let n = u128::from_ne_bytes(buf);
    [n as u64, (n >> 64) as u64]
}
```

In which case, we wouldn't need the `wasm_dummy` feature, as we wouldn't even compile `getrandom` on `wasm32-unknown-unknown`. We would just have the following in the `Cargo.toml`:
```toml
[target.'cfg(not(target = "wasm32-unknown-unknown"))'.dependencies]
getrandom = { version = "0.1.8", features = ['rustc-dep-of-std', 'dummy'] }
```

newpavlov: Unfortunately `target = ".."` does not work in `cfg` expressions, see: rust-lang/rust#63217. I guess we could write something like `cfg(not(all(target_arch="wasm32", target_os="unknown", target_env="unknown")))`. It's a bit unwieldy, but it indeed allows us to remove the `wasm_dummy` feature.

newpavlov:
@josephlr 
Thank you for the nits! Feels free to propose any other similar improvements!

src/lib.rs
```diff
@@ -46,7 +46,9 @@
 //! methods directly, using either `stdweb` or `wasm-bindgen` depending on what
 //! features are activated for this crate. Note that if both features are
 //! enabled `wasm-bindgen` will be used. If neither feature is enabled,
-//! `getrandom` will always fail.
+//! compiling `getrandom` will result in a compilation error. It can be disabled
+//! by enabling the `dummy` feature, with which `getrandom` will use an always
```
josephlr: s/ with which `getrandom` will / which will make `getrandom` /g

src/lib.rs
```diff
@@ -225,18 +227,30 @@ cfg_if! {
                   target_env = "sgx",
               )))] {
         #[path = "rdrand.rs"] mod imp;
-    } else if #[cfg(target_arch = "wasm32")] {
+    // ideally we would like to use `target = "wasm32-unknown-unknown"`,
```
josephlr: Personally I find something like:
```rust
...
} else if #[cfg(all(target_arch = "wasm32", feature = "wasm-bindgen"))] {
    #[path = "wasm32_bindgen.rs"] mod imp;
} else if #[cfg(all(target_arch = "wasm32", feature = "stdweb"))] {
    #[path = "wasm32_stdweb.rs"] mod imp;
} else if #[cfg(feature = "dummy")] {
    #[path = "dummy.rs"] mod imp;
} else {
    compile_error!("\
...
```
to be a little more readable. It avoids nested `cfg_if!`s, and we don't need the `unknown` specifier. If we get here, the other `wasm32` cases are taken care of.

In theory, a new `wasm32` target could be added that's not addressed in the above, but having the `stdweb`/`wasm-bindgen` features work in that case seems fine (even ideal).

newpavlov: I am not sure about not needing the `unknown` specifier, imagine that we don't have WASI or Emscripten, with this config would not be supported for such "new" targets. But I guess since we can't use `target_vendor` here, we are not completely future proof anyway, so it makes sense to simplify the code.

README.md
```diff
@@ -45,17 +44,29 @@ This library is `no_std` compatible, but uses `std` on most platforms.
...
 -   [`wasm-bindgen`](https://crates.io/crates/wasm_bindgen)
 -   [`stdweb`](https://crates.io/crates/stdweb)
 
+By default, compiling `getrandom` for an unsupported target will result in
+a compilation error. If you want to build an application which uses `getrandom`
+for such target, you can either:
+- Use [`[replace]`][replace] or [`[patch]`][patch] section in your `Cargo.toml`
+to switch to a custom implementation with a support of your target.
+- Enable the `dummy` feature to have getrandom use an implementation that always
+fails on unsupported targets.
```
dhardy: `fails at run-time on unsupported targets`?

README.md
```diff
@@ -45,17 +44,29 @@ This library is `no_std` compatible, but uses `std` on most platforms.
...
+
+We also accept pull requests to support additional targets.
```
dhardy: State the obvious? Actually, we might have additional requirements anyway (on-going support for that platform). I don't think we need this comment here.

src/lib.rs
```diff
@@ -46,7 +46,9 @@
 //! methods directly, using either `stdweb` or `wasm-bindgen` depending on what
 //! features are activated for this crate. Note that if both features are
 //! enabled `wasm-bindgen` will be used. If neither feature is enabled,
-//! `getrandom` will always fail.
+//! compiling `getrandom` will result in a compilation error. It can be disabled
```
dhardy: Instead of disabled, try avoided or worked around?

src/lib.rs
```diff
@@ -225,18 +227,22 @@ cfg_if! {
...
+        #[path = "wasm32_bindgen.rs"] mod imp;
+    } else if #[cfg(all(target_arch = "wasm32", feature = "stdweb"))] {
+        #[path = "wasm32_stdweb.rs"] mod imp;
+    } else if #[cfg(feature = "dummy")] {
         #[path = "dummy.rs"] mod imp;
+    } else {
+        compile_error!("\
+            target is not supported, you may enable dummy implementation \
+            using the `dummy` feature or overwrite `getrandom` crate with \
+            a custom one which supports your target using `[replace]` or \
```
dhardy: Can we make this shorter? E.g.

> target not supported; you may enable compile-only support via `dummy` feature or see getrandom#4 for custom target support

newpavlov: Maybe we should add a section to docs regarding handling of unsupported targets and link it in this message? So it will look like this:
> target is not supported, for more information see: https://docs.rs/getrandom/#unsupported-targets

PR #73, opened 2019-08-04
use arguments of the right type for getrandom syscall

RalfJung:
"0" is a 32-bit 0, but we need pointer-sized 0s here.

Found by Miri.


RalfJung:
Closing in favor of https://github.com/rust-random/getrandom/pull/74 which has more explicit types.

PR #78, opened 2019-08-05
Add rustc-dep-of-std feature

josephlr:
This is how `libc` and `hashbrown` support being dependancies of `libstd`.
This allows `getrandom` to use `core` normally, and use the sysroot `core`
when building `libstd`.

This PR and #58 should be enough to get rust-lang/rust#62082 building without errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,7 +102,7 @@ matrix:
 
     - rust: nightly
       env: DESCRIPTION="cross-platform build only"
-      # Redox: wait for https://github.com/rust-lang/rust/issues/60139
+      # libc on Redox is broken: see https://github.com/rust-lang/libc/pull/1450
       install:
         - rustup target add x86_64-sun-solaris
         - rustup target add x86_64-unknown-cloudabi
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ matrix:
         - cargo test --examples
         # remove cached documentation, otherwise files from previous PRs can get included
         - rm -rf target/doc
-        - cargo doc --no-deps --all --all-features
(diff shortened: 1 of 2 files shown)


newpavlov:
You probably also should remove uses of `--all-features` from CI tests.

I am not familiar with the details of how `std` is build, so I hope someone more knowledgeable will confirm, that it's indeed a correct approach in our case.

cc @alexcrichton

josephlr:
> You probably also should remove uses of `--all-features` from CI tests.

Done

> I am not familiar with the details of how `std` is build, so I hope someone more knowledgeable will confirm, that it's indeed a correct approach in our case.

I know we need _at least_ this to properly use the `core` and `compiler_builtins` libraries in the sysroot. We may also need additional functionality, I'm not sure.

alexcrichton:
I can confirm that for building as part of the standard library the `Cargo.toml` bits are correct and would be required.

.travis.yml
```diff
@@ -46,7 +46,7 @@ matrix:
         - cargo test --examples
         # remove cached documentation, otherwise files from previous PRs can get included
         - rm -rf target/doc
-        - cargo doc --no-deps --all --all-features
+        - cargo doc --no-deps --all --features=std
```
newpavlov: To be safe I think it will be better to use `--features="std log"`.

josephlr: Fixed

josephlr:
This should be ready for merging, I updated the Redox comment to better explain the issue in #79.

I also confirmed that this PR + #58 + a modified version of @newpavlov's rust-lang/rust#62082 is now enough to bootstrap `libstd` without errors.

PR #80, opened 2019-08-06
Add warnings for rust_2018_idioms violations

josephlr:
As part of rust-lang/rust#62082, we need to make sure `getrandom` compiles without warnings when using the Rust warning configuration. Checking for `rust_2018_idioms` and `unused_lifetimes` seems to fix things.

I checked the CI and we don't have any warnings on any targets.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -76,7 +76,7 @@ fn os_err_desc(_errno: i32, _buf: &mut [u8]) -> Option<&str> {
 }
 
 impl fmt::Debug for Error {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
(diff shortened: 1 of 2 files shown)


newpavlov:
Maybe `deny` them then? Or those lints may change in future?

josephlr:
One thing [this article](https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md) mentions is passing `RUSTFLAGS="-D warnings"` for all `cargo` invocations in the CI. That way, the crate stays stable, but we know about warnings, and can fail the CI.

Other than `missing_docs` any other warnings you'd like to add?

newpavlov:
`-D warnings` is indeed a better approach. Can you also modify CI configs as part of this PR? As for other warnings, I don't have any additional suggestions right now.

PR #85, opened 2019-08-13
Fix Redox CI

### LINKED ISSUE #79: Redox support
https://github.com/rust-random/getrandom/blob/199b11576eea7aab93a0162e36571b8b607f0000/.travis.yml#L105

Since https://github.com/rust-lang/rust/issues/60139 is still open, we probably don't want to do anything about this yet. In this case however we also need to disable rand tests ([see this failure](https://travis-ci.org/rust-random/rand/jobs/567814932)).

josephlr:
Fixes #79

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,14 +96,13 @@ matrix:
 
     - name: "cross-platform build only"
       rust: nightly
-      # libc on Redox is broken: see https://github.com/rust-lang/libc/pull/1450
       install:
         - rustup target add x86_64-sun-solaris
         - rustup target add x86_64-unknown-cloudabi
         - rustup target add x86_64-unknown-freebsd
         - rustup target add x86_64-fuchsia
         - rustup target add x86_64-unknown-netbsd
-        # - rustup target add x86_64-unknown-redox
+        - rustup target add x86_64-unknown-redox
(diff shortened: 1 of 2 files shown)


newpavlov:
Hopefully rust-lang/rust#63521 will land soon and we will not have to use `xbuild` for Redox.

newpavlov:
@josephlr 
Can you revert to the old `cargo build`? The builder PR has landed, so we should get pre-built Redox `std` soon.

josephlr:
@newpavlov will do. Looks like redox is now back to building in the latest nightly.
https://rust-lang.github.io/rustup-components-history/x86_64-unknown-redox.html

newpavlov:
Argh, looks like `libc` is still broken on Redox...

josephlr:
> Argh, looks like `libc` is still broken on Redox...

I confirmed locally that 0.2.62 seems to fix things. So this can be merged once the CI passes.

josephlr:
Whoops @newpavlov, I thought you had approved this, so I merged it. We should probably enable the feature that requires review before merging. Would that work for you?

newpavlov:
Yeah, no problem. I am not sure if we need this setting, but either way, I think we will need @dhardy for that.

dhardy:
Done

PR #86, opened 2019-08-14
Add VxWorks support

newpavlov:
I couldn't find a proper documentation for VxWorks, so I've mostly just copied code from `std`.

Relevant issue: openssl/openssl#7946

This PR will not work until rust-lang/libc#1473 lands.

cc @bpangWR @BaoshanPang

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ compiler_builtins = { version = "0.1", optional = true }
 core = { version = "1.0", optional = true, package = "rustc-std-workspace-core" }
 
 [target.'cfg(unix)'.dependencies]
-libc = { version = "0.2.62", default-features = false }
+libc = { version = "0.2.64", default-features = false }
 
 [target.'cfg(target_os = "wasi")'.dependencies]
 wasi = "0.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ cfg-if = "0.1"
 compiler_builtins = { version = "0.1", optional = true }
 core = { version = "1.0", optional = true, package = "rustc-std-workspace-core" }
 
-[target.'cfg(any(unix, target_os = "redox"))'.dependencies]
+[target.'cfg(unix)'.dependencies]
 libc = { version = "0.2.62", default-features = false }
 
 [target.'cfg(target_os = "wasi")'.dependencies]
--- a/src/error.rs
+++ b/src/error.rs
@@ -153,7 +153,7 @@ fn internal_desc(error: Error) -> Option<&'static str> {
         BINDGEN_GRV_UNDEF => Some("wasm-bindgen: crypto.getRandomValues is undefined"),
         STDWEB_NO_RNG => Some("stdweb: no randomness source available"),
         STDWEB_RNG_FAILED => Some("stdweb: failed to get randomness"),
-        RAND_SECURE_FATAL => Some("randSecure: fatal error"),
+        RAND_SECURE_FATAL => Some("randSecure: random number generator module is not initialized"),
         _ => None,
     }
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -138,6 +138,7 @@ pub(crate) const BINDGEN_CRYPTO_UNDEF: Error = internal_error!(7);
 pub(crate) const BINDGEN_GRV_UNDEF: Error = internal_error!(8);
 pub(crate) const STDWEB_NO_RNG: Error = internal_error!(9);
 pub(crate) const STDWEB_RNG_FAILED: Error = internal_error!(10);
+pub(crate) const RAND_SECURE_FATAL: Error = internal_error!(11);
 
 fn internal_desc(error: Error) -> Option<&'static str> {
     match error {
@@ -152,6 +153,7 @@ fn internal_desc(error: Error) -> Option<&'static str> {
         BINDGEN_GRV_UNDEF => Some("wasm-bindgen: crypto.getRandomValues is undefined"),
         STDWEB_NO_RNG => Some("stdweb: no randomness source available"),
         STDWEB_RNG_FAILED => Some("stdweb: failed to get randomness"),
+        RAND_SECURE_FATAL => Some("randSecure: fatal error"),
         _ => None,
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -161,19 +161,22 @@ pub use crate::error::Error;
 #[allow(dead_code)]
 mod util;
 
+#[cfg(target_os = "vxworks")]
+#[allow(dead_code)]
+mod util_libc;
+
 cfg_if! {
     // Unlike the other Unix, Fuchsia and iOS don't use the libc to make any calls.
-    if #[cfg(any(target_os = "android", target_os = "dragonfly", target_os = "emscripten",
-                 target_os = "freebsd", target_os = "haiku",     target_os = "illumos",
-                 target_os = "linux",   target_os = "macos",     target_os = "netbsd",
-                 target_os = "openbsd", target_os = "redox",     target_os = "solaris"))] {
+    if #[cfg(any(
+            target_os = "android", target_os = "dragonfly", target_os = "emscripten",
+            target_os = "freebsd", target_os = "haiku",     target_os = "illumos",
+            target_os = "linux",   target_os = "macos",     target_os = "netbsd",
+            target_os = "openbsd", target_os = "redox",     target_os = "solaris",
+        ))] {
         #[allow(dead_code)]
         mod util_libc;
         // Keep std-only trait definitions for backwards compatiblity
         mod error_impls;
-    } else if #[cfg(target_os = "vxworks")] {
-        #[allow(dead_code)]
-        mod util_libc;
     } else if #[cfg(feature = "std")] {
         mod error_impls;
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,6 +27,7 @@
 //! | Haiku            | `/dev/random` (identical to `/dev/urandom`)
 //! | L4RE, SGX, UEFI  | [RDRAND][18]
 //! | Hermit           | [RDRAND][18] as [`sys_rand`][22] is currently broken.
+//! | VxWorks          | `randBytes`
 //! | Web browsers     | [`Crypto.getRandomValues`][14] (see [Support for WebAssembly and ams.js][14])
 //! | Node.js          | [`crypto.randomBytes`][15] (see [Support for WebAssembly and ams.js][16])
 //! | WASI             | [`__wasi_random_get`][17]
@@ -225,6 +226,8 @@ cfg_if! {
         #[path = "solaris_illumos.rs"] mod imp;
     } else if #[cfg(target_os = "wasi")] {
         #[path = "wasi.rs"] mod imp;
+    } else if #[cfg(target_os = "vxworks")] {
+        #[path = "vxworks.rs"] mod imp;
     } else if #[cfg(all(windows, getrandom_uwp))] {
         #[path = "windows_uwp.rs"] mod imp;
     } else if #[cfg(windows)] {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,7 +27,7 @@
 //! | Haiku            | `/dev/random` (identical to `/dev/urandom`)
 //! | L4RE, SGX, UEFI  | [RDRAND][18]
 //! | Hermit           | [RDRAND][18] as [`sys_rand`][22] is currently broken.
-//! | VxWorks          | `randBytes`
+//! | VxWorks          | `randABytes` after checking entropy pool initialization with `randSecure`
 //! | Web browsers     | [`Crypto.getRandomValues`][14] (see [Support for WebAssembly and ams.js][14])
 //! | Node.js          | [`crypto.randomBytes`][15] (see [Support for WebAssembly and ams.js][16])
 //! | WASI             | [`__wasi_random_get`][17]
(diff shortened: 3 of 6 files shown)


BaoshanPang:
@newpavlov 

Thanks for working this, I am actually looking for the way to port rand for vxWorks with latest version, the layout of version 0.7.0 is quiet different than 0.6.1 which we have ported for.

I don't see any problem with the code, it is good to me.

 Can you teach me how to test it for vxWorks? I am seeing rand 0.6.1(and 0.6.5) is used when runing rust libstd unit test, but I can't figure out where/how rand 0.7.0 is used.

Thanks,
Baoshan

newpavlov:
If I understood you correctly, then you don't have to do anything. Assuming this PR is correct, after publishing the next version of `getrandom` with it, `rand v0.7` should start working on VxWorks automatically.

`libstd` only recently started using `rand v0.7`, I think after this PR: rust-lang/rust#61393.

BaoshanPang:
>>libstd only recently started using rand v0.7, I think after this PR: rust-lang/rust#61393.
I see.  Thanks.

src/vxworks.rs
```diff
@@ -11,17 +11,17 @@ use crate::error::{Error, ERRNO_NOT_POSITIVE};
 use core::num::NonZeroU32;
 
 extern "C" {
-    fn randBytes (buf: *mut u8, length: i32) -> i32;
+    fn randBytes(buf: *mut u8, length: i32) -> i32;
```
n-salim: Yes: confirming that the C prototype of randBytes in VxWorks is:

STATUS randBytes
    (
    unsigned char * randBuf,   /* buffer to be written */
    int             numOfBytes /* number of random bytes requested */
    )

where STATUS is a typedef to int, and int is "i32" on all VxWorks targets.

src/vxworks.rs
```diff
@@ -11,17 +11,17 @@ use crate::error::{Error, ERRNO_NOT_POSITIVE};
...
     // errnoLib.h
     fn errnoGet() -> i32;
 }
 
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
     // Prevent overflow of i32
     for chunk in dest.chunks_mut(i32::max_value() as usize) {
-        let ret = randBytes(chunk.as_mut_ptr(), chunk.len() as i32);
+        let ret = unsafe { randBytes(chunk.as_mut_ptr(), chunk.len() as i32) };
         if ret == -1 {
```
n-salim: Yes: "ERROR" (-1) is returned on error. Note that this can happen even in "normal use" if there isn't enough entropy accumulated. Is that ok?

newpavlov: Not, quite. On Linux and other *NIXes we block in such cases until the entropy pool gets properly initialized. Do you have a dedicated error code to signalize "not enough entropy" error?

src/vxworks.rs
newpavlov: What will happen if you'll provide a negative length? In other words, my question is: is it indeed `i32` used for length and not `u32`?

newpavlov:
@n-salim 
Also do we have to provide any linking arguments to Cargo? See [build.rs](https://github.com/rust-random/getrandom/blob/master/build.rs). Note that in future `getrandom` may be used as a dependency of `std`.

n-salim:
> Also do we have to provide any linking arguments to Cargo?
Good question. Luckily randBytes is part of VxWorks libc.a, which we already always link against.

newpavlov:
>Luckily randBytes is part of VxWorks libc.a, which we already always link against.

How about adding `randBytes` to the `libc` crate then? It would allow us to simplify code in `getrandom` a bit. Also please take look at rust-lang/libc#1469.

BaoshanPang:
>>How about adding randBytes to the libc crate then?

I will do it as part of libc#1469.

src/vxworks.rs
n-salim: I need to check with our "random" expert to make sure that this is the right function to call for our purposes here (I think maybe randABytes would be better), and how to detect "not enough entropy".

n-salim:
Let's expose randABytes and randUBytes too.

src/vxworks.rs
newpavlov: To clarify: we want to block if entropy pool has not initialized (i.e. right after boot), but after that we want to draw entropy even if entropy estimate is low (i.e. we don't want to block like reading from `/dev/random` does on Linux).

n-salim: OK in that case I *think* randUBytes would make more sense:

The documentation for that function says:

> This routine is same as randBytes() but it returns OK even if enough entropy has not been created.

But let's wait till I've heard back from our expert, I don't want to mislead you.

src/vxworks.rs
n-salim: >  is it indeed i32 used for length and not u32?
That's a great question, no i32 is not "really" used for the length.

I don't know the the history of this API (i.e. why didn't we use `size_t`), but from what I can tell, under the hood the argument is cast to size_t (so a negative input will turn into a very large number).

src/vxworks.rs
n-salim: I discussed this with our randomness expert and the conclusion was:

1. Use `randABytes`

> That API will return error if not enough entropy has been collected for the random state to be considered secure. Once that point has been reached, the API will always succeed, regardless of how much random data has been fetched by calling randBytes/randABytes/randUBytes. 

2. Use `randSecure()` to determine whether to give a fatal error or retry after a sleep

```
< 0 : fatal error
== 0: retry, still gathering entropy
> 0: ok
```

newpavlov: Great! I will prepare an appropriate update after the `libc` PR lands on crates.io.

newpavlov: BTW can we assume that all VxWorks targets have atomics? It would be nice to reuse [`unsync_init`](https://github.com/rust-random/getrandom/blob/master/src/util.rs#L44) here.

n-salim: Looking at [core::sync::atomic](https://github.com/rust-lang/rust/blob/master/src/libcore/sync/atomic.rs) it appears that Rust atomics are required to be lock-free (whereas in C11 we're allowed to synthesize atomic operations that don't map to hardware operations using a mutex).

unsync_init appears to specifically require atomic load/store for usize.

The targets we plan to support are Intel (32- and 64-bit), ARM (32-bit and 64-bit), and (eventually) PowerPC (32- and 64-bit). From what I can tell these all support atomic operations on usize.

So in short, yes, we **can** use unsync_init here. And thanks for the question, I learned something in answering it!

BaoshanPang:
> If I understood you correctly, then you don't have to do anything. Assuming this PR is correct, after publishing the next version of `getrandom` with it, `rand v0.7` should start working on VxWorks automatically.
> 
> `libstd` only recently started using `rand v0.7`, I think after this PR: [rust-lang/rust#61393](https://github.com/rust-lang/rust/pull/61393).

I just built the latest rust code , although there are both rand 0.7 and rand 0.6.1 listed in Cargo.lock, but the rand used is still 0.6.1 when building libstd unit test:

$ ./x.py test -i --stage 1 src/libstd
...
  Compiling rand v0.6.1

src/vxworks.rs
newpavlov: Will `randSecure()` report error code via `errnoGet()` or directly in return code?

newpavlov: Also, is it guaranteed that if `randABytes` has returned non-negative value, then the buffer got filled completely or does it return how many bytes have been generated?

n-salim: > Will `randSecure()` report error code via `errnoGet()` or directly in return code?

errno is not set. All status is reported by the return value, and you can't use that return value for anything other than to distinguish the three cases I mentioned:

< 0 : fatal error
== 0: retry, still gathering entropy
> 0: ok

n-salim: > Also, is it guaranteed that if `randABytes` has returned non-negative value, then the buffer got filled completely

Yes (based on the documentation and a cursory scan of the source code).

>  or does it return how many bytes have been generated?

No, the only cases are:

< 0 : failed
== 0 : success (all bytes written)

src/vxworks.rs
```diff
@@ -0,0 +1,36 @@
...
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Implementation for VxWorks
+use crate::error::{Error, RAND_SECURE_FATAL};
+use crate::util_libc::last_os_error;
+use core::sync::atomic::{AtomicBool, Ordering::Relaxed};
+
+static RNG_INIT: AtomicBool = AtomicBool::new(false);
```
josephlr: Just a nit (feel free to ignore), most of our code puts our lazy statics inside the function where they're used (to limit their scope).

src/lib.rs
```diff
@@ -170,6 +171,9 @@ cfg_if! {
         mod util_libc;
         // Keep std-only trait definitions for backwards compatiblity
         mod error_impls;
+    } else if #[cfg(target_os = "vxworks")] {
```
josephlr: See #105 and [my comment](https://github.com/rust-random/getrandom/pull/105#issuecomment-532492625), this won't properly include `error_impls` if `target_os = "vxworks"` and `feature = "std"`. The best bet is to just include it the big `cfg!` block above.

src/vxworks.rs
```diff
@@ -0,0 +1,36 @@
...
+
+pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
+    while !RNG_INIT.load(Relaxed) {
```
josephlr: We should use `LazyBool` here, something like:

```rust
fn rng_init() -> bool {
    loop {
        let ret = unsafe { libc::randSecure() }
        if ret != 0 {
            return ret > 0;
        }
        unsafe { libc::usleep(10) };
    }
}
```

```rust
    static RNG_INIT: LazyBool = LazyBool::new();
    if !RNG_INIT.unsync_init(rng_init) {
        return Err(RAND_SECURE_FATAL);
    }

    // Prevent overflow of i32
    ...
```

You could also use a lambda, if you think it would look better.

Cargo.toml
```diff
@@ -25,7 +25,7 @@ cfg-if = "0.1"
 compiler_builtins = { version = "0.1", optional = true }
 core = { version = "1.0", optional = true, package = "rustc-std-workspace-core" }
 
-[target.'cfg(any(unix, target_os = "redox"))'.dependencies]
+[target.'cfg(unix)'.dependencies]
 libc = { version = "0.2.62", default-features = false }
```
josephlr: We should update this to `0.2.63` to get https://github.com/rust-lang/libc/commit/e1a417f1b2f3055fdbd1054d1835a9071fb7c964

newpavlov: `libc v0.2.63` was yanked due to [a serious bug](https://github.com/rust-lang/libc/issues/1515#issuecomment-531003188). VxWroks support will only be included in (currently unreleased) v0.2.64, so there is no reasons to update it now.

Cargo.toml
```diff
@@ -25,7 +25,7 @@ cfg-if = "0.1"
 compiler_builtins = { version = "0.1", optional = true }
 core = { version = "1.0", optional = true, package = "rustc-std-workspace-core" }
 
-[target.'cfg(any(unix, target_os = "redox"))'.dependencies]
+[target.'cfg(unix)'.dependencies]
```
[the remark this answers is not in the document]
newpavlov: Newer `libc` versions assume that Redox is `unix`, and since we use exactly those versions Redox support on older Rust versions will not work either way.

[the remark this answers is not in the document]
josephlr: seems reasonable to me

Cargo.toml
josephlr: Sounds good, I guess we'll have to wait for `libc v0.2.64`. Not a huge deal.

src/vxworks.rs
newpavlov: @BaoshanPang @n-salim
Is it possible for the system to recover after `randSecure` failure or can we assume that once it has failed it will always fail?

BaoshanPang: Yes, the system can recover after randSecure failure. This is the randSecure's function desciption:

* randSecure - determine if state is secure                                                                                        
*                                                                                                                                  
* This routine determines if state is secure and returns status.                                                                   
* When the system boots, the SW random number lib contains no entropy. Once                                                        
* enough entropy has been collected in order to make the random numbers                                                            
* cryptographically secure, the state is considered to be secure.                                                                  
*                                                                                                                                  
* It will make a system call to a kernel side handler.                                                                             
*                                                                                                                                  
* RETURNS: 1 if the state is secure, 0 if the state is not secure, ERROR if                                                        
* the random number generator module is not initialized.                                                                           
*                                                                                                                                  
* ERRNOS: N/A

src/error.rs
```diff
@@ -152,6 +153,7 @@ fn internal_desc(error: Error) -> Option<&'static str> {
         BINDGEN_GRV_UNDEF => Some("wasm-bindgen: crypto.getRandomValues is undefined"),
         STDWEB_NO_RNG => Some("stdweb: no randomness source available"),
         STDWEB_RNG_FAILED => Some("stdweb: failed to get randomness"),
+        RAND_SECURE_FATAL => Some("randSecure: fatal error"),
```
josephlr: "randSecure: random number generator module is not initialized" to match @BaoshanPang's description below.

src/vxworks.rs
josephlr: > "ERROR if the random number generator module is not initialized."

If I'm understanding this right, the RNG module might not be initialized, but it could become initialized later?

If that's the case, the current code looks good to me.

Issue #87, opened 2019-08-14
0.1.9 seems to have broken wasm32-unknown-unknown


Lokathor:
I just started to get CI breaks today with 0.1.9, saying that wasm32-unknown-unknown is an unsupported platform, but things were fine before today.

Does the new version need to be yanked and then published as 0.2.0?

newpavlov:
This change was introduced in #71 and we decided not to treat it as a breaking one. If you want the old behavior, you can enable the `dummy` feature.

Lokathor:
The use of getrandom is because it's a dev-dependency for benchmarks, but tests build all dev-dependency crates even if tests don't use that crate because cargo is just silly like that. I don't believe any actual tests use getrandom.

If the dummy feature is on does that affect targets _besides_ wasm32-unknown-unknown? I don't mind flipping it on if non wasm targets aren't affected.

Also, you cannot conditional include a feature in cargo as far as I know, so this sort of thing really needs to be a breaking change when you do it.

newpavlov:
Argh... `dummy` features will influence all targets, initially there was also a `wasm_dummy` feature, but it got removed. One option could've been to write:
```toml
[dev-dependencies]
getrandom = { version = "0.1", features = ["dummy"] }
```
But unfortunately dev features leak to non-dev builds, due to the long-standing rust-lang/cargo#1796...

As a temporary solution you can run CI tests with `--features=getrandom/dummy`. I guess we could also add a `dummy` feature to `rand` for convenience.

@dhardy @josephlr 
Thoughts?

Lokathor:
Alright I'll change the CI for now.

Side note: I don't use `rand` at all, so just a fix there will not help this case.

josephlr:
@newpavlov I think we can learn from what other system crates (like `libc`) have done. These crates will _build_ on any target, they will just be empty (or lacking key functions) on unsupported architectures.

I think we should change the behavior of `getrandom` to not have the `getrandom()` function on unsupported platforms, but otherwise build correctly. We should also proprobly make the current compiler error a warning to correctly point people to the docs on how to use unsupported targets.

@Lokathor would this work for your use case? If you don't actually call the getrandom function on unsupported platforms, everything should be fine.

newpavlov:
@josephlr 
I don't think it's a good solutions for us. Don't forget that most users will not use `getrandom` directly, so your approach will result in a confusing compilation error "cannot find function `getrandom`", instead of the human-friendly one.

[...]

Issue #63, opened 2019-07-11
Consider using web_sys and unifying wasm32-unknown-unknown targets


josephlr:
Right now, [`stdweb` depends on `wasm-bindgen`](https://github.com/koute/stdweb/blob/197abe5c4eef74308baefe888d93c7751ce5f6c2/Cargo.toml#L30) so it doesn't really make a whole lot of sense to have two seperate implementations in `getrandom` for this.

A better approach might just be to depend on the [`web_sys` crate](https://crates.io/crates/web-sys) from the [`wasm-bindgen`](https://github.com/rustwasm/wasm-bindgen) repo and eliminate the need for separate wasm32 features.

I'm not sure how this would work with Node, but for Client Web, we could just use [`Crypto::get_random_values_with_u8_array`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.Crypto.html#method.get_random_values_with_u8_array)

newpavlov:
I don't see why we should use `web_sys` while it's built on top of `wasm-bindgen`, instead of directly using the latter.

Previously WASM target was discussed in #17 and we decided to keep it as-is for the time being. Also see this [comment](https://github.com/koute/stdweb/issues/318#issuecomment-480620969).

dhardy:
@koute would it be reasonable to make `wasm-bindgen` a hard dependency for `wasm32-unknown-unknown` now?

najamelan:
If stdweb does not give any advantage over wasm-bindgen, and if it can't lead to less deps being pulled in, this would greatly simplify issues downstream... I mean to just get the random data out of wasm-bindgen without features.

newpavlov:
I am strongly against making `wasm-bindgen` enabled by default for `wasm32-unknown-unknown`. As I wrote several times, this target can not make any assumptions about a target on which it will be executed. It can be a browser, Node.JS, smart-contract, game module system and who knows what else. If we want simplify WASM builds for browsers/Node.JS, let's push for [dedicated targets](https://github.com/rustwasm/team/issues/38).

najamelan:
@newpavlov Does setting `stdweb` work on these other platforms? 

Or are you just thinking other features like `wasm-bindgen` and `stdweb` will maybe be added later? Or do these targets just need to use `dummy` and they are not supported?

newpavlov:
@najamelan 
[...]

PR #89, opened 2019-08-15
Make the documentation for WASM targets clearer

najamelan:
For people who run in a compile error because `wasm-bindgen` and `stdweb` features are not enabled, clarify that they can enable them even if getrandom gets pulled in by one of their dependencies.

I think this isn't necessarily obvious. I didn't realize that I could set features for indirect dependencies. 

What I do wonder about though is if this is the desired workflow, why does the `rand` crate forward this feature? And should a user specify `getrandom` with say `wasm-bindgen` in their Cargo.toml, or should they set the feature on `rand`.

Note also that library authors might have to set this at least in `dev-dependencies` to make tests and examples run.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,6 +62,10 @@
 //! by enabling the `dummy` feature, which will make `getrandom` to use an
 //! always failing implementation.
 //!
+//! If you are seeing this compiler error even though your crate doesn't directly
+//! depend on getrandom (one of your dependencies pulls it in), you can still solve
+//! this by adding getrandom as a dependency to your `Cargo.toml` with the correct feature.
+//!
 //! The WASI target `wasm32-wasi` uses the `__wasi_random_get` function defined
 //! by the WASI standard.
 //!


dhardy:
Thanks for the PR — sounds like a sensible thing to do. Perhaps add a line like the following:
```
getrandom = { version = "0.1.9", features = ["dummy"] }
```

src/lib.rs
```diff
@@ -62,6 +62,10 @@
 //! by enabling the `dummy` feature, which will make `getrandom` to use an
 //! always failing implementation.
 //!
+//! If you are seeing this compiler error even though your crate doesn't directly
+//! depend on getrandom (one of your dependencies pulls it in), you can still solve
+//! this by adding getrandom as a dependency to your `Cargo.toml` with the correct feature.
```
newpavlov: I don't think there is **the** feature to be enabled. Also please take a look at #87. I am not sure if enabling `dummy`, `stdweb` or `wasm-bindgen` feature in an non-application `Cargo.toml` is a correct recommendation.

newpavlov:
So I think the problem will mainly arise in dev-builds, e.g. because benchmarks need system entropy, but not tests and CI job tests for `wasm32-unknown-unknown`. I think a better recommendation would be to add the following line to `Cargo.toml`:
```toml
[dev-dependencies]
getrandom = "0.1"
```

And enable a needed `getrandom` feature manually in CI job, e.g. by using `cargo test --features getrandom/wasm-bindgen` or `cargo bench --features getrandom/dummy`.

najamelan:
@dhardy do you mean add an example line to the docs? I could add that, but I haven't yet really understood the usefulness of the "dummy" feature though. If it's being pulled in, it's because people want it to work, I suppose most users don't want to just turn compile time failure into runtime failure?

Also, I have said it on the futures-rs issue, but one thing that worries me a bit is that when libraries enable this feature, the end use has to change their Cargo.toml, which means it's a breaking change. Currently futures-rs pulls this in with a feature on futures. So if an intermediate library suddenly decides to use that functionality from futures (join or select macro), that might break end user application. Meaning that features like using a join macro now becomes a breaking change, which isn't documented anywhere.

This feature also isn't exactly additive. The docs say if both features are enabled `wasm-bindgen` is chosen. I suppose that is to make them somewhat additive. If libraries should be transparent to this and only application devs should set it, why would they ever set both though? It would also mean that somewhere along the dependency line a library had enabled `stdweb`, which as I understand it they shouldn't have, but also means that they are probably counting on it where as the `wasm-bindgen` feature being enabled as well will silently turn off the functionality of the `stdweb` feature.

> So I think the problem will mainly arise in dev-builds, e.g. because benchmarks need system entropy
@newpavlov I don't know if it's correctly used, but in the futures library, if you turn on the feature that enables the `join` and `select` macros, rand gets pulled which pulls getrandom. They use it in order to randomly select which future to poll first. That's not exactly a dev-build only problem.

dhardy:
From Cargo's point of view, all feature flags are additive. For the `getrandom` crate, only a single API is needed to provide its implementation, and when multiple are available it will choose the best of those; from the user's point of view this does not matter at all.

Yes, true, adding a dependency on `getrandom` or `rand` may now cause build failures on `wasm-unknown-unknown`. I think we just have to accept that. Eventually this target should be deprecated in favour of `wasm-wasi`, where we don't have this problem.

Alternatively maybe we should just assume `wasm-bindgen` for this target now (see #63).

najamelan:
after the discussion in #63 I will try to rephrase the doc PR to be clearer. Does the following seem right (to come after the paragraph about wasm32-unkown-unknown?

> Library authors should not set these features. Application developers should set them even when getrandom gets pulled in by a dependency. 
> 
> **Note** For library authors, this means that if you start pulling in getrandom (which happens if you depend on rand with default features), it will be a breaking change on wasm32 and you should notify your users in documentation that setting the above mentioned features is necessary.

newpavlov:
I am not sure if need note about the breaking change. How about something like this?

Library authors should not set these features outside of dev builds. Meaning if you test `wasm32-unknown-unknown` target in your CI, you can add the following line to your Cargo.toml:
```toml
[dev-dependencies]
getrandom = { version = "0.1", features = ["wasm-bindgen"] }
```

**Note:** features enabled for dev builds leak to non-dev builds (see rust-lang/cargo#1796), but only for the current workspace or path dependencies, meaning you can publish crate with such dependency to crates.io without issues.

Alternatively you can add `getrandom` to your dependencies without enabling any features:
```toml
[dev-dependencies]
getrandom = { version = "0.1", features = ["wasm-bindgen"] }
```
And enable a wanted feature manually in your WASM CI job:
```sh
cargo build --features=getrandom/dummy
```

newpavlov:
Closed in favor of #90.

PR #90, opened 2019-08-16
Always build on wasm32-unknown-unknown

josephlr:
This updates the documentation to explain exacly what we are doing on
this target. Also adds a test that the target builds without features.

Address issues in #87 and makes #89 unnecessary.

CC: @Lokathor

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -61,7 +61,7 @@
 //! Getrandom also supports `wasm32-unknown-unknown` by directly calling
 //! javascript methods. Rust currently has two ways to do this: [bindgen] and
 //! [stdweb]. Getrandom can using either by enabling the `wasm-bindgen` or
-//! `stdweb` crate features. Note that if both features are enabled
+//! `stdweb` crate features. Note that if both features are enabled,
 //! `wasm-bindgen` will be used. If neither feature is enabled, calls to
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -50,21 +51,22 @@
...
+//! `asmjs-unknown-emscripten`, `wasm32-unknown-emscripten`, and
+//! `wasm32-experimental-emscripten` use Emscripten's `/dev/random` emulation.
+//! The WASI target `wasm32-wasi` uses the [`__wasi_random_get`][17] function
+//! defined by the WASI standard.
 //!
-//! The WASI target `wasm32-wasi` uses the `__wasi_random_get` function defined
-//! by the WASI standard.
+//! Getrandom also supports `wasm32-unknown-unknown` by directly calling
+//! javascript methods. Rust currently has two ways to do this: [bindgen] and
+//! [stdweb]. Getrandom can using either by enabling the `wasm-bindgen` or
```
dhardy: can using

newpavlov:
Personally I am not happy about this exception... But I guess we will have to add it for the time being.

src/lib.rs
josephlr: Fixed, also make sure to correctly capitalize JavaScript

PR #92, opened 2019-08-17
Fix CSP error for wasm bindgen

### LINKED ISSUE #91: use with wasm-bindgen triggers CSP error
Due to `let this = Function::new("return this").call(&JsValue::undefined());` I'm currently seeing an error using a sensible CSP:

```
EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'wasm-eval'".
```

`js_sys` uses a more involved method to fetch the global object that eventually falls back to the unsafe version. 
https://rustwasm.github.io/wasm-bindgen/api/src/js_sys/lib.rs.html#4450-4523

Can either `js_sys` be added as a dependency or this logic (or a subset) be reproduced here?

evq:
Fixes #91 

Access the global object via `self` directly instead of the current `return this` function created from a string. Creating a function from a string results in a CSP error as it requires `unsafe-eval`.

See [tweetnacl](https://github.com/dchest/tweetnacl-js/blob/master/nacl.js#L1139) which uses this strategy (although it also checks for `msCrypto` for IE11 compatibility - here we don't because IE11 does not have wasm support).

Tested on node, firefox and brave (chromium) using the wasm-pack test.

### DIFF
--- a/src/wasm32_bindgen.rs
+++ b/src/wasm32_bindgen.rs
@@ -58,54 +58,38 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
 }
 
 fn getrandom_init() -> Result<RngSource, Error> {
-    // First up we need to detect if we're running in node.js or a
-    // browser. To do this we get ahold of the `this` object (in a bit
-    // of a roundabout fashion).
-    //
-    // Once we have `this` we look at its `self` property, which is
-    // only defined on the web (either a main window or web worker).
-    let this = Function::new("return this").call(&JsValue::undefined());
-    assert!(this != JsValue::undefined());
-    let this = This::from(this);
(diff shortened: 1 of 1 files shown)


josephlr:
I think we would be better off using js_sys to get the global object, as that seems to be the more supported way of doing things

evq:
The `js_sys` strategy is a bit overkill here since we only need a way to fetch the global object on browsers, for which `self` should always work. They have to go through a lot of extra hoops because they're trying to make it possible to get the global object even on nodejs and other non-browser platforms.

newpavlov:
I think we can discuss it in #63, but I still don't see a reason why we would want to use crates built on top of `wasm-bindgen` instead of using it directly.

PR #95, opened 2019-08-19
Enable `std` by default

### LINKED ISSUE #94: trait impl disappeared without bumping major version
Sorry to be the bearer of versioning news, `0.1.10` contains a backwards-incompatible change with respect to `0.1.8` and so on - the feature flags around [`error_impls`](https://github.com/rust-random/getrandom/commit/b69f83221041414b30adbbcef13e3934da2c04f7#diff-b4aea3e418ccdb71239b96952d9cddb6L155-L170) changed, and had been accidentally(?) including `error_impls` for specific OSes regardless of `std` being selected or not.

The change around feature flags makes sense, but without bumping major versions, version `0.1.10` can cause breakages: see the end of the log [here](https://travis-ci.org/fastly/lucet/builds/573939344). The missing impl is because `minisign` depends on `getrandom = "0.1.3"`, where `0.1.10` should be a backwards compatible update, but is not due to the missing trait. We can work around that by specifying `std`, but I wanted to make sure y'all knew about this.

josephlr:
Fixes #94

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,15 @@ fn get_random_buf() -> Result<[u8; 32], getrandom::Error> {
 
 ## Features
 
-This library is `no_std` for every supported target. However, getting randomness
+By default, the `std` feature is enabled, allowing getrandom's
+[`Error`](https://docs.rs/getrandom/0.1.10/getrandom/struct.Error.html)
+to implement traits like
+[`std::error::Error`](https://doc.rust-lang.org/std/error/trait.Error.html). To
+use getrandom in a `no_std` environment, the `std` feature must be disabled by
+setting `default-features = false` in `Cargo.toml`. Without this feature, this
+library is `no_std` for every supported target.
+
+However, getting randomness
 usually requires calling some external system API. This means most platforms
 will require linking against system libraries (i.e. `libc` for Unix,
 `Advapi32.dll` for Windows, Security framework on iOS, etc...).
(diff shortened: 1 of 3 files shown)


newpavlov:
You forgot to add `default-features = false` to `no_std` CI test. I think we can merge this and publish an update after the `rand_core`  update lands.

README.md
```diff
@@ -39,7 +39,15 @@ fn get_random_buf() -> Result<[u8; 32], getrandom::Error> {
...
-This library is `no_std` for every supported target. However, getting randomness
+By default, the `std` feature is enabled, allowing getrandom's
+[`Error`](https://docs.rs/getrandom/0.1.10/getrandom/struct.Error.html)
+to implement traits like
+[`std::error::Error`](https://doc.rust-lang.org/std/error/trait.Error.html). To
+use getrandom in a `no_std` environment, the `std` feature must be disabled by
+setting `default-features = false` in `Cargo.toml`. Without this feature, this
+library is `no_std` for every supported target.
+
+However, getting randomness
```
dhardy: Starting a paragraph with *however* is *never* good grammar...

dhardy:
Isn't it the other way around — `rand` libs are waiting on `getrandom` 0.2 to be published? A breaking change is a breaking change, regardless of whether users have updated.

josephlr:
Closing in favor of #96. This change isn't the right way to fix #94

PR #99, opened 2019-08-28
Disable unneeded features

elichai:
Hi,
Fist the default features both these crates give are unneeded here.
Second this is a step towards making `wasm-bindgen` works with `no-std`.

The question is, is there a way to replace `thread_local` with something that can work without std?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,8 +32,8 @@ libc = { version = "0.2.62", default-features = false }
 wasi = "0.5"
 
 [target.wasm32-unknown-unknown.dependencies]
-wasm-bindgen = { version = "0.2.29", optional = true }
-stdweb = { version = "0.4.18", optional = true }
+wasm-bindgen = { version = "0.2.29", default-features = false, optional = true }
+stdweb = { version = "0.4.18", default-features = false, optional = true }
 
 [features]
 std = []


dhardy:
For those not in the know, @elichai is the author of https://github.com/elichai/random-rs.

@elichai, `thread_local` requires some type of synchronisation primitive, and as far as I am aware there is no reason to care about WASM without `std` (at least, this implementation requires either Node.js or a browser to do anything).

newpavlov:
Well, I would also like to make WASM targets `std`-independent, but our hands are tied by `wasm-bindgen` here. In my opinion `wasm32-unknown-unknown` shouldn't be an `std` target in the first place, i.e. it would be fine to provide a default allocator, but it does not make much sense to provide everything else. I guess one of the biggest reasons why it was not done like this is an absence of `HashMap`, which currently is not part of `alloc` crate due to the entropy requirements.

As for `thread_local`, I _think_ we can replace it with atomics in a similar manner to other targets. We just didn't bother with it since currently WASM does not have threading capabilities, so IIUC `thread_local`operations compile down to simple memory loads and stores.

newpavlov:
@elichai 
Because it does not really retrieve system randomness during runtime. For example by default it uses `ahash` which in turn uses [`const-random`](https://crates.io/crates/const-random).

josephlr:
@elichai I think this is a reasonable change, except that it looks like [`wasm-bindgen` does not support `no_std` environments](https://github.com/rustwasm/wasm-bindgen/issues/1745). It would be interesting to see if we could drop the final uses of `std` in `getrandom`, which should be possible with `LazyBool`.

> And it seems that I'm not going to use this crate in random-rs anyway, just wanted to contribute :) 

Thanks! I would assume this is due to our higher minimum Rust version, but if there's some other issue with `getrandom` that's preventing your use, please open an issue. I don't know if you plan on having a cryptographicly secure RNG in `random-rs`.

elichai:
@josephlr Yeah it's the minimum rust version.. I hoped I could maybe PR here support for older compilers but the way this crate currently works is strongly tied with new rust features.

(No saying it's bad, just different use cases)

josephlr:
Closing due to https://github.com/rustwasm/wasm-bindgen/issues/1745, we would need support from the `wasm-bindgen` crate to support `no_std`.

@elichai feel free to reopen this if the situation changes or I missed something. Thanks for the PR.

PR #101, opened 2019-08-29
Update wasi dependency

mati865:
cc https://github.com/rust-lang/rust/pull/63806
It failed with 
```
invalid license Apache-2.0 WITH LLVM-exception in /checkout/obj/build/tmp/distcheck/src/../vendor/wasi/Cargo.toml
```
It could be worked around by adding exception for `wasi` but since it's license has been changed more optimal solution is to just update `wasi`.


newpavlov:
There is already `wasi v0.7`. See #100. ;)

josephlr:
Closing in favor of #100. Thanks @mati865 for the fix!

PR #102, opened 2019-08-29
Proposal for bindgen changes

josephlr:
This is my idea for improving the `wasm32-unknown-unknown` with `bindgen` target. 

Note: This is a Work in Progress and contains many ideas for improving the bindgen target. We don't necessarily have to do all of them.

#### Using `js-sys` and `web-sys`

These crates are maintained [by the same group that maintains `wasm-bindgen`](https://github.com/rustwasm/wasm-bindgen) and are intended to work similar to how the `libc` crate works today. Strictly speaking, the `libc` crate is never necessary. You could always just declare `extern "C"` functions yourself; however, having the `libc` crate ensures that these bindings are correct for your platform.

Similarly, you could always declare interfaces to the JS APIs and Browser APIs manually via `wasm-bindgen`. However, using these binding crates makes our job easier:

- It allows us to properly obtain the global object (doing this correctly [is quite complex](https://docs.rs/js-sys/0.3.27/src/js_sys/lib.rs.html#4450-4523)). See #92 
- When host bindings are supported, we don't have to do the work of wiring everything up.
- These bindings are extremely subtle (even more so than `libc`) due to the complex mapping between Rust and JS types. It's better to just have someone else deal with them and have them fix `bindgen` bugs.

Furthermore, the `web-sys` crate is configurable, so you only need to build what you need to use (we only need `Crypto`, `Window`, and `WorkerGlobalScope`).

I think that if we're OK depending on crates like `libc` that are maintained by the Rust team and just declare platform bindings, we should be OK with using a similar crate for wasm.

#### Better error handling with `catch` bindings

This change also moves to using `catch` bindings (the default in `js-sys` and `web-sys`) for all of our bindings. This allows us to properly return an error when certain APIs fail (or fail to exist) at runtime. 

This also means we no longer need things like `get_random_values_fn`. If the API is not there, the function will just return an error (`undefined` I think). This change does not depend on using `web-sys`/`js-sys`.

#### Removing `std::thread_local`

Right now, the `stdweb` implementation only checks if we are in Node vs a Browser in its `init` function. This does the same for `bindgen`. Given that the cost of retrieving these objects is mostly just the `bindgen` overhead (which we incur regardless), it doesn't make a whole lot of sense to hold onto a `thread_local` `NodeCrypto`. We can just get the object each time we need it.

#### Future changes

We could additionally add `error!` calls to actually print out the underlying JS errors.

We could also try to reduce the duplication between the `stdweb` and `bindgen` code, but that's better for a future PR.

### DIFF
--- a/src/wasm32_bindgen.rs
+++ b/src/wasm32_bindgen.rs
@@ -7,107 +7,54 @@
 // except according to those terms.
 
 //! Implementation for WASM via wasm-bindgen
-extern crate std;
+use js_sys::global;
+use wasm_bindgen::{prelude::*, JsCast};
+use web_sys::{window, Crypto, WorkerGlobalScope};
 
-use core::cell::RefCell;
-use core::mem;
-use std::thread_local;
-
-use wasm_bindgen::prelude::*;
-
-use crate::error::{BINDGEN_CRYPTO_UNDEF, BINDGEN_GRV_UNDEF};
-use crate::Error;
-
-#[derive(Clone, Debug)]
-enum RngSource {
-    Node(NodeCrypto),
-    Browser(BrowserCrypto),
-}
-
-// JsValues are always per-thread, so we initialize RngSource for each thread.
-//   See: https://github.com/rustwasm/wasm-bindgen/pull/955
-thread_local!(
-    static RNG_SOURCE: RefCell<Option<RngSource>> = RefCell::new(None);
-);
+use crate::util::LazyBool;
+use crate::{error, Error};
 
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
-    assert_eq!(mem::size_of::<usize>(), 4);
-
-    RNG_SOURCE.with(|f| {
-        let mut source = f.borrow_mut();
-        if source.is_none() {
-            *source = Some(getrandom_init()?);
+    static IS_NODE: LazyBool = LazyBool::new();
+    if IS_NODE.unsync_init(|| node_crypto().is_some()) {
+        if node_crypto().unwrap().random_fill_sync(dest).is_err() {
+            return Err(error::BINDGEN_NODE_FAILED);
         }
-
-        match source.as_ref().unwrap() {
-            RngSource::Node(n) => n.random_fill_sync(dest),
-            RngSource::Browser(n) => {
-                // see https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
-                //
-                // where it says:
-                //
-                // > A QuotaExceededError DOMException is thrown if the
-                // > requested length is greater than 65536 bytes.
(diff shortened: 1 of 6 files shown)


newpavlov:
Can you provide a motivation section? IIUC they are abstraction layers built on top of `wasm-bindgen`, so bringing-in additional dependencies to simplify code a tiny bit goes against the spirit of `getrandom` in my opinion.

josephlr:
> Can you provide a motivation section? IIUC they are abstraction layers built on top of `wasm-bindgen`, so bringing-in additional dependencies to simplify code a tiny bit goes against the spirit of `getrandom` in my opinion.

Done, PTAL. I tried to split up the proposed changes into sections.

newpavlov:
Hm, you argumentation seems reasonable. I don't have a strong opinion here, so I will leave decision to @dhardy.

Note: if this change will get merged, in `rand`'s Cargo.toml we probably should use:
```toml
wasm-bindgen = ["bindgen"]
binidgen = ["getrandom/bindgen"]
```

src/wasm32_bindgen.rs
```diff
@@ -7,107 +7,54 @@
...
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
-    assert_eq!(mem::size_of::<usize>(), 4);
-
-    RNG_SOURCE.with(|f| {
-        let mut source = f.borrow_mut();
-        if source.is_none() {
-            *source = Some(getrandom_init()?);
+    static IS_NODE: LazyBool = LazyBool::new();
+    if IS_NODE.unsync_init(|| node_crypto().is_some()) {
+        if node_crypto().unwrap().random_fill_sync(dest).is_err() {
```
dhardy: Why `.unwrap()` and not `.ok_or(..)?` ?

alexcrichton:
This seems fine to me either way personally. Writing `extern "C"` shorthand is fine most of the time if you're only importing one or two functions, so I would consider using `#[wasm_bindgen]` directly to be basically equivalent for one or two functions. I agree that getting the global object is pretty tricky, though, and using `js-sys` for that is probably worth it.

newpavlov:
BTW I think we should not rename the feature. I think the first association with `bindgen` will be [a wrong one](https://crates.io/crates/bindgen). We still use `wasm-bindgen` under the hood, so it should be fine to continue to use the old name. Plus it will reduce amount of v0.2 breakage.

josephlr:
Closing this for now. Thanks everyone for the feedback. My idea for solving #4 conflicts with this, so we should resolve those issues before dealing with this one.

> BTW I think we should not rename the feature. I think the first association with `bindgen` will be [a wrong one](https://crates.io/crates/bindgen). 

That's a fair point, sticking with the name `wasm-bindgen` (for whatever approach we go with) seems like the best bet.

PR #105, opened 2019-09-18
Fix building issue

BaoshanPang:
I saw this error when building rust for vxWorks targets:
/folk/prj-rust-dev/bpang/ala-diab-pb19l/latest/latest/.cargo/registry/src/github.com-1ecc6299db9ec823/rand_core-0.5.1/src/err\
or.rs:150:28
    |
150 |             Error { inner: Box::new(error) }
    |                            ^^^^^^^^^^^^^^^ the trait `std::error::Error` is not implemented for `getrandom::error::Error`
    |
    = note: required for the cast to the object type `dyn std::error::Error + std::marker::Send + std::marker::Sync`

And after adding "mod error_impls;" the issue disappears.

r? @newpavlov 
cc @n-salim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.10] - 2019-08-18
+### Changed
+- Use the dummy implementation on `wasm32-unknown-unknown` even with the
+disabled `dummy` feature. [#90]
+
+### Fixed
+- Fix CSP error for `wasm-bindgen`. [#92]
+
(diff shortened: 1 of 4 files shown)


josephlr:
@BaoshanPang thanks for opening this. This is an error caused by my change. Essentially when `feature = "std"` and `target_os = "vxworks"`, we don't add `error_impls` when we should.

Ideally we would fix the `cfg_if` declaration so that `mod error_impls` is included when `feature = "std"` is set, instead of this change, but that's going to make things messier. So I guess this change is OK.

Note that we're going to remove the automatic `error_impls` inclusion in `0.2`. However, this shouldn't affect you as you're only using `getrandom` though `rand`/`rand_core` which sets the features correct regardless.

Can you rebase this change onto the current master, then I can approve and merge it.

josephlr:
This should just be part of #86, as that PR hasn't merged yet. However, this is an important point, and should be accounted for in review.

PR #108, opened 2019-09-20
Improve CI for web targets

josephlr:
This PR uses simplifies the testing story for `getrandom`. All targets now use the same testing file (eliminating the need for a separate `tests/wasm_bindgen` crate).

We also use `.cargo/config` to allow users to run `cargo test` normally for `wasm32-wasi` or `wasm32-unknown bindgen`. This just requires that [`wasmtime`](https://github.com/CraneStation/wasmtime) or [`wasm-bindgen-test-runner`](https://github.com/rustwasm/wasm-bindgen/tree/master/crates/test)  exist in the user's path.

The CI for these targets has been overhauled:
  - Tests are now actually run a larger number of targets:
    - `wasm32-wasi`
    - `stdweb`: Node and Chrome
    - `wasm-bindgen`: Node, Firefox, and Chrome
  - Tests are built/linked for `asmjs-unknown-emscripten`
  - We download pre-built binaries for all our dependancies

Results:
  - Less code
  - Less duplication
  - More tests are run
  - WASM CI is much faster. Decreased from 13 mins to 4.5 mins

This is not part of #98, so it can be merged into master.


Cargo.toml
```diff
@@ -35,9 +32,14 @@ wasi = "0.7"
...
+[target.wasm32-unknown-unknown.dev-dependencies]
+wasm-bindgen-test = "0.2"
+
 [features]
 std = []
 # Enables dummy implementation for unsupported targets
 dummy = []
 # Unstable feature to support being a libstd dependency
 rustc-dep-of-std = ["compiler_builtins", "core"]
+# Usstable feature for testing
```
newpavlov:
```suggestion
# Unstable feature for testing
```

josephlr:
Updated description to reflect the scope of these changes. 

@dhardy this is ready, the CI runs significantly faster now.

dhardy:
I would prefer that someone more familiar with WASM could review. @stevebob wrote this test originally (rand#696).

Issue #84, opened 2019-08-13
Feature for using CPU-based RNG


josephlr:
For some custom x86 targets I am working on (https://github.com/intel/rust-hypervisor-firmware), it would be nice to have a feature that says "just use the on-CPU RNG". 

Having a feature for this seems better than adding more targets to the list of targets that use RDRAND. It would also be useful more conventional targets (i.e. Linux) that don't want to use a syscall to get random data (i.e. for performance).

newpavlov:
I am not sure about this. I think it falls under using a custom `getrandom` implementation, so right now such projects should use `[patch]` or `[replace]`, and hopefully eventually we will get a lang item for it.

>don't want to use a syscall to get random data (i.e. for performance)

I think if you want performance you should use `ThreadRng`, not overwrite `getrandom`.

josephlr:
> I am not sure about this. I think it falls under using a custom `getrandom` implementation, so right now such projects should use `[patch]` or `[replace]`, and hopefully eventually we will get a lang item for it.

So this is totally possible, but it requires a patch to the `cfg_if` statement in `lib.rs`. Essentially, if my target is `x86_64-unknown-none`, I have to patch `getrandom` to use the implementation in `rdrand.rs`. A lang-item also wouldn't work, because I wouldn't want to substitute my own custom implementation. I would just want `getrandom` to use a different, already existing implementation.

That's why I think a feature might make more sense.

> Do you suggest using a user-space CSPRNG? What do you plan to use for the initial seed? Either way, `getrandom` conception is a minimal crate for retrieving _system_ entropy, and I don't think that user-space RNGs fit here.

Oh I agree, `getrandom` should only be about getting the entropy from some target-specific source (either the OS or the CPU).

> I think if you want performance you should use `ThreadRng`, not overwrite `getrandom`.

Correct, `ThreadRng` is way better and (IIRC) more performant. My point about Linux was wrong.

newpavlov:
>A lang-item also wouldn't work, because I wouldn't want to substitute my own custom implementation.

As I see it we would have a `getrandom-rdrand` crate which will contain only RDRAND implementation, so you will not have to write custom implementation yourself.

I guess we can add this feature until we get this hypothetical lang item. Can you prepare a PR?

dhardy:
> As I see it we would have a getrandom-rdrand crate which will contain only RDRAND implementation, so you will not have to write custom implementation yourself.

I guess this makes sense. Then users can use something like following:
```toml
[...]

PR #109, opened 2019-09-20
Add support for Custom RNGs

### LINKED ISSUE #4: embedded source compatibility

josephlr:
This is a breaking change for #98, so it's being merged against the 0.2 branch.

This change consists on several significant changes, so reviewing on a per commit basis may be the easiest. The changes (in rough commit order) are:

- The `getrandom::Error` constants are moved to be associated constants, and they are now all public and documented.
- The features related to `wasm32` implementations (`wasm-bindgen` and `stdweb`) are removed from the main crate
- Functionality is added (behind the `custom` Cargo feature) to enable registering and using a custom RNG implementation. This is done by having the custom function be declared with `#[no_mangle]`, and having the function called via `extern "C"`.
- To demonstrate the soundness of the Custom RNG approch, the following Custom RNGs are added:
  - `stdweb-getrandom`: Replacing the `stdweb` feature
  - `wasm-bindgen-getrandom`: Replacing the `wasm-bindgen` feature
- Tests/CI are updated to work with the new WASM Custom RNGs

Fixes #4 and allows for having a `getrandom-rdrand` Custom RNG (which provides a better solution for #84)

### DIFF
--- a/custom/stdweb/Cargo.toml
+++ b/custom/stdweb/Cargo.toml
@@ -1,0 +1,26 @@
+[package]
+name = "stdweb-getrandom"
+version = "0.1.0"
+edition = "2018"
+authors = ["The Rand Project Developers"]
+license = "MIT OR Apache-2.0"
+description = "Custom shim for using getrandom with stdweb"
+documentation = "https://docs.rs/stdweb-getrandom"
+repository = "https://github.com/rust-random/getrandom"
+categories = ["wasm"]
+
+[dependencies]
+getrandom = { path = "../..", version = "0.2", features = ["custom"] }
+stdweb = "0.4.18"
+
+# Test-only features allowing us to reuse most of the code in common.rs
+[features]
+default = ["test-stdweb"]
+test-stdweb = []
+
+[[test]]
+name = "common"
+path = "../../tests/common.rs"
+
+[package.metadata.docs.rs]
+default-target = "wasm32-unknown-unknown"
--- a/custom/stdweb/src/lib.rs
+++ b/custom/stdweb/src/lib.rs
@@ -6,15 +6,16 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-//! Implementation for WASM via stdweb
-extern crate std;
+#![recursion_limit = "128"]
+#[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
+compile_error!("This crate is only for the `wasm32-unknown-unknown` target");
 
 use core::mem;
 use std::sync::Once;
 
 use stdweb::js;
 
-use crate::Error;
+use getrandom::{register_custom_getrandom, Error};
 
 #[derive(Clone, Copy, Debug)]
 enum RngSource {
@@ -22,7 +23,9 @@ enum RngSource {
     Node,
 }
 
-pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
+register_custom_getrandom!(getrandom_inner);
+
+fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
     assert_eq!(mem::size_of::<usize>(), 4);
     static ONCE: Once = Once::new();
     static mut RNG_SOURCE: Result<RngSource, Error> = Ok(RngSource::Node);
--- a/src/custom.rs
+++ b/src/custom.rs
@@ -1,0 +1,45 @@
+// Copyright 2018 Developers of the Rand project.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! An implementation which calls out to an externally defined function.
+use crate::Error;
+use core::num::NonZeroU32;
+
+/// Register a function to be invoked by `getrandom` on custom targets.
+///
+/// This function will only be invoked on targets not supported by `getrandom`.
+/// This prevents crate dependencies from either inadvertently or maliciously
+/// overriding the secure RNG implementations in `getrandom`.
+///
+/// *This API requires the following crate features to be activated: `custom`*
+#[macro_export]
+macro_rules! register_custom_getrandom {
+    ($path:path) => {
+        // We use an extern "C" function to get the guarantees of a stable ABI.
+        #[no_mangle]
+        extern "C" fn __getrandom_custom(dest: *mut u8, len: usize) -> u32 {
+            let slice = unsafe { ::std::slice::from_raw_parts_mut(dest, len) };
+            match $path(slice) {
+                Ok(()) => 0,
+                Err(e) => e.code().get(),
+            }
+        }
+    };
+}
+
+#[allow(dead_code)]
+pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
+    extern "C" {
+        fn __getrandom_custom(dest: *mut u8, len: usize) -> u32;
+    }
+    let ret = unsafe { __getrandom_custom(dest.as_mut_ptr(), dest.len()) };
+    match NonZeroU32::new(ret) {
+        None => Ok(()),
+        Some(code) => Err(Error::from(code)),
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -134,7 +134,10 @@ extern crate cfg_if;
 
 mod error;
 mod util;
-
+// To prevent a breaking change when targets are added, we always export the
+// register_custom_getrandom macro, so old Custom RNG crates continue to build.
+#[cfg(feature = "custom")]
+mod custom;
 #[cfg(feature = "std")]
 mod error_impls;
 
@@ -201,6 +204,8 @@ cfg_if! {
                 ");
             }
         }
+    } else if #[cfg(feature = "custom")] {
+        use custom as imp;
     } else {
         compile_error!("\
             target is not supported, for more information see: \
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,7 +126,6 @@
     html_root_url = "https://rust-random.github.io/rand/"
 )]
 #![no_std]
-#![cfg_attr(feature = "stdweb", recursion_limit = "128")]
 #![warn(rust_2018_idioms, unused_lifetimes, missing_docs)]
 
 #[macro_use]
@@ -191,19 +190,6 @@ cfg_if! {
                   target_env = "sgx",
               )))] {
         #[path = "rdrand.rs"] mod imp;
-    } else if #[cfg(all(target_arch = "wasm32", target_os = "unknown"))] {
-        cfg_if! {
-            if #[cfg(feature = "wasm-bindgen")] {
-                #[path = "wasm32_bindgen.rs"] mod imp;
-            } else if #[cfg(feature = "stdweb")] {
-                #[path = "wasm32_stdweb.rs"] mod imp;
-            } else {
-                compile_error!("\
-                    Enable crate features to use the wasm32-unknown-unknown target, see: \
-                    https://docs.rs/getrandom/#support-for-webassembly-and-asmjs\
-                ");
-            }
-        }
     } else if #[cfg(feature = "custom")] {
         use custom as imp;
     } else {
--- a/tests/common.rs
+++ b/tests/common.rs
@@ -1,3 +1,7 @@
+// Explicitly use the Custom RNG crates to link them in.
+#[cfg(feature = "test-stdweb")]
+use stdweb_getrandom as _;
+
 #[cfg(feature = "wasm-bindgen")]
 use wasm_bindgen_test::*;
 
--- a/tests/common.rs
+++ b/tests/common.rs
@@ -1,8 +1,10 @@
 // Explicitly use the Custom RNG crates to link them in.
 #[cfg(feature = "test-stdweb")]
 use stdweb_getrandom as _;
+#[cfg(feature = "test-bindgen")]
+use wasm_bindgen_getrandom as _;
 
-#[cfg(feature = "wasm-bindgen")]
+#[cfg(feature = "test-bindgen")]
 use wasm_bindgen_test::*;
 
 use getrandom::getrandom;
@@ -10,14 +12,14 @@ use getrandom::getrandom;
 #[cfg(feature = "test-in-browser")]
 wasm_bindgen_test_configure!(run_in_browser);
 
-#[cfg_attr(feature = "wasm-bindgen", wasm_bindgen_test)]
+#[cfg_attr(feature = "test-bindgen", wasm_bindgen_test)]
 #[test]
 fn test_zero() {
     // Test that APIs are happy with zero-length requests
     getrandom(&mut [0u8; 0]).unwrap();
 }
 
-#[cfg_attr(feature = "wasm-bindgen", wasm_bindgen_test)]
+#[cfg_attr(feature = "test-bindgen", wasm_bindgen_test)]
 #[test]
 fn test_diff() {
     let mut v1 = [0u8; 1000];
@@ -35,7 +37,7 @@ fn test_diff() {
     assert!(n_diff_bits >= v1.len() as u32);
 }
 
-#[cfg_attr(feature = "wasm-bindgen", wasm_bindgen_test)]
+#[cfg_attr(feature = "test-bindgen", wasm_bindgen_test)]
 #[test]
 fn test_huge() {
     let mut huge = [0u8; 100_000];
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,8 @@ matrix:
         - cargo web test --nodejs --target=wasm32-unknown-unknown
         - cargo web test --target=wasm32-unknown-unknown
         # wasm-bindgen tests (Node, Firefox, Chrome)
-        - cargo test --target wasm32-unknown-unknown --features=wasm-bindgen
+        - cd ../wasm-bindgen
+        - cargo test --target wasm32-unknown-unknown
         - GECKODRIVER=$HOME/geckodriver cargo test --target wasm32-unknown-unknown --features=test-in-browser
         - CHROMEDRIVER=$HOME/chromedriver cargo test --target wasm32-unknown-unknown --features=test-in-browser
 
@@ -84,6 +85,7 @@ matrix:
       rust: nightly
       os: linux
       install:
+        - rustup target add wasm32-unknown-unknown
         - cargo --list | egrep "^\s*deadlinks$" -q || cargo install cargo-deadlinks
         - cargo deadlinks -V
       script:
@@ -96,6 +98,7 @@ matrix:
         # remove cached documentation, otherwise files from previous PRs can get included
         - rm -rf target/doc
         - cargo doc --no-deps --features=std,custom
+        - cargo doc --no-deps --manifest-path=custom/wasm-bindgen/Cargo.toml --target=wasm32-unknown-unknown
         - cargo deadlinks --dir target/doc
         # also test minimum dependency versions are usable
         - cargo generate-lockfile -Z minimal-versions
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,13 +91,14 @@ matrix:
         - cargo test --benches
         # Check that setting various features does not break the build
         - cargo build --features=std
+        - cargo build --features=custom
         # remove cached documentation, otherwise files from previous PRs can get included
         - rm -rf target/doc
-        - cargo doc --no-deps --features=std
+        - cargo doc --no-deps --features=std,custom
         - cargo deadlinks --dir target/doc
         # also test minimum dependency versions are usable
         - cargo generate-lockfile -Z minimal-versions
-        - cargo test --features=std
+        - cargo test --features=std,custom
 
     - <<: *nightly_and_docs
       name: "OSX, nightly, docs"
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,8 +71,9 @@ matrix:
         # wasi tests
         - cargo test --target wasm32-wasi
         # stdweb tests (Node, Chrome)
-        - cargo web test --nodejs --target=wasm32-unknown-unknown --features=stdweb
-        - cargo web test --target=wasm32-unknown-unknown --features=stdweb
+        - cd custom/stdweb
+        - cargo web test --nodejs --target=wasm32-unknown-unknown
+        - cargo web test --target=wasm32-unknown-unknown
         # wasm-bindgen tests (Node, Firefox, Chrome)
         - cargo test --target wasm32-unknown-unknown --features=wasm-bindgen
         - GECKODRIVER=$HOME/geckodriver cargo test --target wasm32-unknown-unknown --features=test-in-browser
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ appveyor = { repository = "rust-random/getrandom" }
 [workspace]
 members = [
     "custom/stdweb",
+    "custom/wasm-bindgen",
 ]
 
 [dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,21 +27,12 @@ libc = { version = "0.2.64", default-features = false }
 [target.'cfg(target_os = "wasi")'.dependencies]
 wasi = "0.9"
 
-[target.wasm32-unknown-unknown.dependencies]
-wasm-bindgen = { version = "0.2.29", optional = true }
-stdweb = { version = "0.4.18", optional = true }
-
-[target.wasm32-unknown-unknown.dev-dependencies]
-wasm-bindgen-test = "0.2"
-
 [features]
 std = []
 # Feature to enable custom RNG implementations
 custom = []
 # Unstable feature to support being a libstd dependency
 rustc-dep-of-std = ["compiler_builtins", "core"]
-# Unstable feature for testing
-test-in-browser = ["wasm-bindgen"]
 
 [package.metadata.docs.rs]
 features = ["std", "custom"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,11 @@ exclude = ["utils/*", ".*", "appveyor.yml"]
 travis-ci = { repository = "rust-random/getrandom" }
 appveyor = { repository = "rust-random/getrandom" }
 
+[workspace]
+members = [
+    "custom/stdweb",
+]
+
 [dependencies]
 cfg-if = "0.1.2"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,10 +36,12 @@ wasm-bindgen-test = "0.2"
 
 [features]
 std = []
+# Feature to enable custom RNG implementations
+custom = []
 # Unstable feature to support being a libstd dependency
 rustc-dep-of-std = ["compiler_builtins", "core"]
 # Unstable feature for testing
 test-in-browser = ["wasm-bindgen"]
 
 [package.metadata.docs.rs]
-features = ["std"]
+features = ["std", "custom"]
--- a/custom/wasm-bindgen/Cargo.toml
+++ b/custom/wasm-bindgen/Cargo.toml
@@ -1,0 +1,30 @@
+[package]
+name = "wasm-bindgen-getrandom"
+version = "0.1.0"
+edition = "2018"
+authors = ["The Rand Project Developers"]
+license = "MIT OR Apache-2.0"
+description = "Custom shim for using getrandom with wasm-bindgen"
+documentation = "https://docs.rs/wasm-bindgen-getrandom"
+repository = "https://github.com/rust-random/getrandom"
+categories = ["wasm"]
+
+[dependencies]
+getrandom = { path = "../..", version = "0.2", features = ["custom"] }
+wasm-bindgen = "0.2.29"
+
+[dev-dependencies]
+wasm-bindgen-test = "0.2"
+
+# Test-only features allowing us to reuse most of the code in common.rs
+[features]
+default = ["test-bindgen"]
+test-bindgen = []
+test-in-browser = ["test-bindgen"]
+
+[[test]]
+name = "common"
+path = "../../tests/common.rs"
+
+[package.metadata.docs.rs]
+default-target = "wasm32-unknown-unknown"
--- a/custom/wasm-bindgen/src/lib.rs
+++ b/custom/wasm-bindgen/src/lib.rs
@@ -6,8 +6,8 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-//! Implementation for WASM via wasm-bindgen
-extern crate std;
+#[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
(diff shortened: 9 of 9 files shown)


src/error.rs
```diff
@@ -35,10 +35,14 @@ impl Error {
     pub(crate) const RTL_GEN_RANDOM_FAILED: Error = internal_error!(4);
     pub(crate) const FAILED_RDRAND: Error = internal_error!(5);
     pub(crate) const NO_RDRAND: Error = internal_error!(6);
-    pub(crate) const BINDGEN_CRYPTO_UNDEF: Error = internal_error!(7);
-    pub(crate) const BINDGEN_GRV_UNDEF: Error = internal_error!(8);
-    pub(crate) const STDWEB_NO_RNG: Error = internal_error!(9);
-    pub(crate) const STDWEB_RNG_FAILED: Error = internal_error!(10);
+    #[doc(hidden)]
+    pub const BINDGEN_CRYPTO_UNDEF: Error = internal_error!(7);
```
dhardy: Why hide these? They are defined constants either way.

src/custom.rs
```diff
@@ -0,0 +1,31 @@
...
+        fn __getrandom_custom(dest: &mut [u8]) -> Result<(), Error> {
+            $path(dest)
+        }
+    };
+}
+
+pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
+    extern "Rust" {
+        #[allow(improper_ctypes)] // See rust-lang/rust#64593
+        fn __getrandom_custom(dest: &mut [u8]) -> Result<(), Error>;
```
dhardy: Looks like a nice solution.

The call stack is now potentially `getrandom::getrandom` → `getrandom::getrandom_inner` → `__getrandom_custom` → `provider::getrandom_inner`, but use of `#[inline]` in a couple of places reduces this to 2.

src/error.rs
josephlr: I'll need to go back an document these (this was just to make the CI happy). There's also the separate question of if we should expose all of the `Error` constants or just those necessary to make the custom RNGs work. What do you think?

src/custom.rs
josephlr: I think even without the `#[inline]` Rust will manage to get rid of all the intermediate function calls except the `__getrandom_custom`, but I should run a compile test to make sure.

src/error.rs
dhardy: There's also the issue of `Debug`/ `Display` implementations (error strings) when moving implementations out of the main crate.

src/custom.rs
dhardy: Yes, and I think `__getrandom_custom` can never be inlined, but it shouldn't be a big issue.

josephlr:
> I get why you built this on top of your CI changes, but it doesn't require your other PR (removing `dummy`), and I'm not convinced we want that (although I believe the main objection to making `wasm32-unknown-unknown` a compile failure was that it's a breaking change).

Ya the previous discussion is why I proposed it. I think having the dummy impl is almost never what a user wants. I think that regardless of if we implement `wasm-bindgen`/`stdweb` as Custom RNGs or keep them in the main crate, we should make `wasm32-unknown-unknown` with no implementation a compile-time error.

Of course, what to do for no-implementation `wasm32-unknown-unknown` is separate from the issue of removing the `dummy` Cargo feature. https://github.com/rust-random/getrandom/pull/71 added it to help users who were relying on the 0.1.0 - 0.1.8 behavior where we always used a dummy impl for unsupported targets.

One interesting concept is that a "dummy" impl could be just another Custom RNG  

> 
> Although it's a nice demonstration of custom RNGs, is there much point in moving the WASM implementations out of the base crate? Dependency reduction?

dhardy:
We can always re-add the `dummy` feature later if need be, so it is safe to remove now. I guess then it makes sense to merge #107 @newpavlov.

josephlr:
> Although it's a nice demonstration of custom RNGs, is there much point in moving the WASM implementations out of the base crate? Dependency reduction?

The main reason (other than making the dependency story easier) would be to eliminate the need for crates to keep reexporting the `stdweb` and `wasm-bindgen` features.

Thinking of a large Rust application, where `getrandom` might be at the end of a very long dependency chain. It's the end-user, not the immediate library using `getrandom`, that's in the best position to determine what type of `wasm32-unknown-unknown` environment they are in.

I'm not sure how this calculation changes if we are able to detect at build-time if we are in `cargo-web` vs `wasm-bindgen` vs neither. I'll have to do some more looking.

src/error.rs
josephlr: I'm not sure what you mean? The error strings would stay in the main `getrandom`, I guess it means Custom RNGs not in this repo just have to use numbers instead of words. That's not ideal, but fixing it is probably just too hard.

dhardy:
The alternative being, as we tell `rand` users now, that they should depend on `getrandom` directly just to add the feature flag. However that potentially breaks if any crate in the build depends on a different version of `getrandom`, whereas with an external crate implementing the feature it doesn't.

The complication is that we have more crate versions to deal with. If we release `getrandom 0.3` in the future, we need to make a new release of `getrandom-wasm-bindgen` just to update the version of `getrandom` it depends on, since the crate is actually a dependency (assuming the `Error` type and `__getrandom_custom` prototype don't change).

josephlr:
> The alternative being, as we tell `rand` users now, that they should depend on `getrandom` directly just to add the feature flag. However that potentially breaks if any crate in the build depends on a different version of `getrandom`, whereas with an external crate implementing the feature it doesn't.

External crates have another benefit: it makes custom `getrandom` implementations look and feel more similar to custom allocators or panic handlers. Intermediate crates can depend on allocation or panicking without knowing exactly how it will be dealt with by the application.

Of course, those have specific `lang_item`s, but our macro-based approach seems like the best thing to do without compiler support for custom `lang_item`s.

> The complication is that we have more crate versions to deal with. If we release `getrandom 0.3` in the future, we need to make a new release of `getrandom-wasm-bindgen` just to update the version of `getrandom` it depends on, since the crate is actually a dependency (assuming the `Error` type and `__getrandom_custom` prototype don't change).

So the situation isn't too bad provided two conditions are met:
  - `__getrandom_custom` signature/name doesn't change (i.e. we don't need more arguments)
  - `Error` stays a `NonZeroU32`

For example, let's say we have the following versions of `getrandom`:
  - `getrandom v0.2.5` - Last version before semver-trick
  - `getrandom v0.2.6` - Shim via semver-trick to `v0.3`
  - `getrandom v0.3.0` - Backwards incompatible, but not violating the above conditions

Then we can have:
  - `getrandom-wasm-bindgen v0.1.0` depending on `getrandom v0.2`
  - `getrandom-wasm-bindgen v0.1.1` depending on `getrandom v0.3`

and nothing should break.

The following use cases all work how you would expect:
  - A program depends on `getrandom v0.2.5` and `getrandom-wasm-bindgen v0.1.1`
    - `getrandom v0.3.0` is built, but never called.
    - `getrandom-wasm-bindgen`'s function is invoked by `getrandom v0.2.5` without issues.
  - A program depends on `getrandom v0.3.0` and `getrandom-wasm-bindgen v0.1.0`
    - Some version of `getrandom v0.2` is built, but never called.
    - `getrandom-wasm-bindgen`'s function is invoked by `getrandom v0.3.0` without issues.
  - A user has `getrandom v0.3.0` _and_  `getrandom v0.2.5` and _either_ version `getrandom-wasm-bindgen`
    - Both versions of `getrandom` will be built and (potentially) called.
    - Both versions of `getrandom` (when built for a custom target) will link against the same function declared by `getrandom-wasm-bindgen`.
    - Thus, the custom RNG will control the behavior for both versions of `getrandom`.

The only issue would be if there are multiple Custom RNG handlers in a crate's dependency tree, then you'd get a linking error. However, that shouldn't happen as Cargo should only select on of `getrandom-wasm-bindgen v0.1.0` or `getrandom-wasm-bindgen v0.1.1` as they are semver compatible.

This all works because the `extern`/`#[no_mangle]` namespace is global across crates and across versions. It also only needs the representations to be consistent it can't do version checking or anything.

dhardy:
Agreed. Thus we have extra incentive to make sure the API is stable; this looks fine to me but given the scarcity of doc around `extern "Rust"` it may be worth @RalfJung taking a quick look.
```rust
extern "Rust" {
    fn __getrandom_custom(dest: &mut [u8]) -> Result<(), Error>;
}
```

Once #107 and #108 are ready lets rebase this.

josephlr:
> Once #107 and #108 are ready lets rebase this.

Done, I rebased the `0.2` branch onto latest master as well.

src/error.rs
dhardy: Fixing it is https://github.com/rust-random/rand/issues/837 which it looks like we will not solve. But for these WASM crates we should keep the error stings in `getrandom`.

Unless we add an auxilliary extern Rust function:
```rust
fn __getrandom_error_msg(code: NonZeroU32) -> Option<&'static str>;
```
I don't believe it's worth it for now though (and it doesn't solve error messages for anything outside `getrandom`).

josephlr:
Continuing the discussion from @newpavlov and @dhardy's [comments](https://github.com/rust-random/getrandom/pull/107#issuecomment-534053850) in #107 

> @josephlr
> So if I understand correctly you intend to replace the `dummy` feature with the custom impl crate, right? But I still don't understand how you intend to work around `wasm32-unknown-unknown` breakage in downstream crates, without a clear plan I don't think we should remove the `dummy` impl.

I poorly communicated how I wanted to fix the potential `wasm32-unknown-unknown` breakage in upstream crates.  So here's my solution:

First, this PR will add in a `getrandom-dummy` crate that implements a custom RNG that always fails. I didn't initially think this was necessary, but it makes a lot of the code simpler for downstream crates. This required changing the `register_custom_getrandom!` macro to always be present if the `custom` feature is active. This is probably good, as it allows us to add support for targets in later versions of `getrandom` without breaking back-compat.  

Second, we will consider a _simplified_ version of the `rand` crate. This crate has two features `wasm-bindgen` and `stdweb` that (right now) just forward to the `getrandom v0.1` features. `rand` wants to use `getrandom v0.2` without making any breaking changes. Specifically (on `wasm32-unknown-unknown`) it wants to:
  - If `stdweb` feature is enabled, use `stdweb` 
  - If `wasm-bindgen` feature is enabled, use `wasm-bindgen`
  - If both are enabled, use `wasm-bindgen`
  - If neither are enabled, use a dummy implementation.

To do this, `rand` would have the following dependencies in its `Cargo.toml`:

```toml
[dependencies]
getrandom = "0.2"

[target.wasm32-unknown-unknown.dependencies]
wasm-bindgen = { package = "getrandom-wasm-bindgen", version = "0.1", optional = true }
stdweb = { package = "getrandom-stdweb", version = "0.1", optional = true }
getrandom-dummy = "0.1"
```

and the following code in its `src/lib.rs`:

```rust
#[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
cfg_if! {
    if #[cfg(feature = "wasm-bindgen")] {
        extern crate wasm_bindgen;
    } else if #[cfg(feature = "stdweb")] {
        extern crate stdweb;
    } else {
        extern crate getrandom_dummy;
    }
}
```
ensuring that for all `wasm32-unknown-unknown` builds, only a single custom RNG gets linked in.

josephlr:
As for `rand_core`, it only needs to depend on `getrandom::Error` not `getrandom::getrandom`. This means we can use a (on by default) Cargo feature `imp` to gate `getrandom::getrandom`. Then, `rand_core` can just have the following `Cargo.toml`:
```toml
[features]
std = ["getrandom", "getrandom/std"]

[dependencies]
getrandom = { version = "0.2", default-features = false, optional = true }
```
This then allows `rand_core` to compile for any Rust target, as `getrandom::Error` doesn't depend on anything other than `core`. It also allows a future version of `rand_core` to do:
```toml
[features]
std = ["getrandom/std"]

[dependencies]
getrandom = { version = "0.2", default-features = false }
```
and unconditionally depend on `getrandom`, solving issues raised in #110.

This `imp` feature should probably have its own PR, I just wanted to show it would fix these issues.

dhardy:
@josephlr the use of `extern crate` does nothing except introduce the crate's API into the build's namespace; in Edition 2018 it does nothing at all except in a few corner cases (standard libraries, `#[macro_use]`). So there will still be linker conflicts with both `stdweb` and `wasm-bindgen` features enabled and I don't know a way of fixing this (*again*, it's a Cargo limitation).

Maybe we can get away with simply telling people *don't enable both*; I don't know.

Why add a `getrandom-dummy` crate over the `dummy` feature? I see no sense; part of the rartionale for this feature was that you don't have to worry about getting configuration right (when not actually calling `getrandom`), but with this crate the build would fail if enabling another `getrandom-*` provider.

Since `rand_core` 0.5.1 it does use `getrandom` directly (to implement `OsRng`).

josephlr:
> Since `rand_core` 0.5.1 it does use `getrandom` directly (to implement `OsRng`).

Crap, I forgot about this. Scratch the stuff about `rand_core` above.

> Why add a `getrandom-dummy` crate over the `dummy` feature? I see no sense; part of the rartionale for this feature was that you don't have to worry about getting configuration right (when not actually calling `getrandom`), but with this crate the build would fail if enabling another `getrandom-*` provider.

This crate would mainly be used to allow crates to migrate to the new system, while exposing their old APIs, while using less code. It also allow you to work around a cargo limitation where you can't have enable features on a crate based on specific targets. The final reason is that it would allow us to actually remove the `dummy` feature from the main crate, and tell people to only use `getrandom-dummy` in specific, limited circumstances.

josephlr:
> @josephlr the use of `extern crate` does nothing except introduce the crate's API into the build's namespace; in Edition 2018 it does nothing at all except in a few corner cases (standard libraries, `#[macro_use]`). So there will still be linker conflicts with both `stdweb` and `wasm-bindgen` features enabled and I don't know a way of fixing this (_again_, it's a Cargo limitation).

So this isn't true, but I thought the same thing (that `extern crate` did essentially nothing in Rust 2018). It turns out, if there are no references to a crate (either by using a symbol, a `use` statement, or a `extern crate` declaration), then it won't be linked in, regardless of the Cargo dependencies. So the code in https://github.com/rust-random/getrandom/pull/109#issuecomment-534281770 works fine (I tested it).

I also make sure this wasn't a particular linker property. But the `wasm-bindgen` linker, `cargo-web` linker, the default Linux linker, etc... all have this property.  They will complain about duplicate symbols if two custom implementations are linked in. They will complain about a missing symbol if no custom implementation is linked in.

dhardy:
> @josephlr the use of extern crate does nothing except 

From testing, it appears that my understanding is wrong on this and that linking only happens if (a) `extern crate ..` is used or (b) if some symbol is referenced, so your `cfg` thing may work. *However* for some reason "multiple definition" errors are not *always* emitted when both sub-crates are linked. Demonstration [code here](https://github.com/dhardy/rust-sublib-linking).

Note that one can also write `use getrandom_stdweb as _;` which appears to be equivalent to `extern crate`.

Furthermore, this means that an application adding a dependency on, say, `getrandom-wasm-bindgen`, is not enough to enable this backend, thus this form of configuration is not as convenient as I had assumed.

The language spec doesn't directly cover this unfortunately. The section on [extern crates](https://doc.rust-lang.org/reference/items/extern-crates.html) does talk about linking, but it is not clear whether this has been updated for Edition 2018. The sections on *use declarations* and *linking* also do not clear this up.

josephlr:
> > @josephlr the use of extern crate does nothing except

Looks like we submitted at the exact same time 😁

> _However_ for some reason "multiple definition" errors are not _always_ emitted when both sub-crates are linked.

I also ran into this sometimes, apparently the key is to make sure the custom RNG crates are of type `dylib`, at least that fixed some issues for me. Does it for you?

> Furthermore, this means that an application adding a dependency on, say, `getrandom-wasm-bindgen`, is not enough to enable this backend, thus this form of configuration is not as convenient as I had assumed.

I agree that it could be a little more convenient, but it's not too bad (you just need one line in your lib's `src/lib.rs`, we will just have to document exactly how to _use_ these crates as well as how to create your own Custom RNGs.

Do you think this is still the way to go for Custom RNGs?

> The language spec doesn't directly cover this unfortunately.

I agree, but it does seem like the referencing of a crate (not its declaration as a Cargo dependency) is what controls all aspects of linking.

josephlr:
@dhardy @newpavlov I rebased this, updated the description, and better split the changes into commits. Let me know if you think any of these changes should be in their own PR.

dhardy:
I would prefer to keep anything to-do with `dummy` out of this PR.

> This crate would mainly be used to allow crates to migrate to the new system, while exposing their old APIs, while using less code. It also allow you to work around a Cargo limitation where you can't have enable features on a crate based on specific targets. The final reason is that it would allow us to actually remove the dummy feature from the main crate, and tell people to only use getrandom-dummy in specific, limited circumstances.

The Cargo features issue shouldn't apply here (`rand_core` would always depend on the `dummy` feature) and the less-code argument isn't convincing (if even true). It comes down to (a) whether we wish to keep the feature and (b) whether we wish to gracefully handle `dummy` and another backend without the `cfg_if` in user-code. So lets leave it until later.

josephlr:
> It comes down to (a) whether we wish to keep the feature and (b) whether we wish to gracefully handle `dummy` and another backend without the `cfg_if` in user-code. So lets leave it until later.

Make sense, `dummy` stuff is removed.

custom/stdweb/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+edition = "2018"
+authors = ["The Rand Project Developers"]
+license = "MIT OR Apache-2.0"
+description = "Custom shim for using getrandom with stdweb"
+documentation = "https://docs.rs/getrandom-stdweb"
+repository = "https://github.com/rust-random/getrandom/tree/master/custom/stdweb"
+categories = ["wasm"]
+
+[dependencies]
+getrandom = { path = "../..", features = ["custom"] }
```
newpavlov: I think it's wroth to add `version = "0.2"`, same with the `getrandom-wasm-bindgen` crate.

custom/stdweb/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "getrandom-stdweb"
+version = "0.1.0"
+edition = "2018"
+authors = ["The Rand Project Developers"]
+license = "MIT OR Apache-2.0"
+description = "Custom shim for using getrandom with stdweb"
+documentation = "https://docs.rs/getrandom-stdweb"
+repository = "https://github.com/rust-random/getrandom/tree/master/custom/stdweb"
```
newpavlov: I think `https://github.com/rust-random/getrandom` should be enough.

custom/stdweb/src/lib.rs
```diff
@@ -6,26 +6,28 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-//! Implementation for WASM via stdweb
-extern crate std;
+//! `getrandom` implementation for WASM via stdweb
+#![cfg(all(target_arch = "wasm32", target_os = "unknown"))]
```
newpavlov: I think `compilation_error!` would be a bit more clear, something liek this::
```rust
#[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
compilation_error!("This crate can be used only on `wasm32-unknown-unknown` target");
```

custom/stdweb/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+log = "0.4"
+
+# Unstable features used to build tests correctly
+[features]
+default = ["test-stdweb"]
+test-stdweb = []
+
+[[test]]
+name = "common"
+path = "../../tests/common.rs"
```
newpavlov: Do we really need it? IIUC we already test `stdweb` and `wasm-bindgen` via `getrandom`. Or is it for removing WASM-related test features to the custom crates?

custom/stdweb/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+categories = ["wasm"]
+
+[dependencies]
+getrandom = { path = "../..", features = ["custom"] }
+stdweb = "0.4.18"
+log = "0.4"
+
+# Unstable features used to build tests correctly
+[features]
+default = ["test-stdweb"]
```
newpavlov: It's a bit strange to enable an unstable feature by default. Can't we do it in CI job somehow?

RalfJung:
> given the scarcity of doc around extern "Rust" it may be worth @RalfJung taking a quick look

I am afraid I know little about that. All I know is that `extern "Rust" fn foo()`" and `fn foo()` are the same type; `extern` just controls the ABI. The Rust call ABI is unstable but I suppose you can use it as long as you are sure the other side is compiled with the exact same version of rustc? libstd does this, but that may be exploiting its special status as (libstd can rely on unspecified details of the rsutc versions it come with in ways that user code cannot).

@gnzlbg and/or @rkruppe would probably know more.

RalfJung:
It seems to me like this "have a hook that can be implemented once by some crate in the crate tree" could be a separate concern solved by a dedicated create (I am half-surprised @dtolnay does not already have a crate for this. ;)

There's been an RFC to make this a native Rust feature, but it's been postponed recently: https://github.com/rust-lang/rfcs/pull/2492. @jethrogb proposed some macro-based solution there, so they might also be able to give some useful feedback here.

jethrogb:
> It seems to me like this "have a hook that can be implemented once by some crate in the crate tree" could be a separate concern solved by a dedicated create

This is basically the [extern-existential](https://crates.io/crates/extern-existential) crate, but this should really be a language feature.

josephlr:
@newpavlov given the above feedback on `extern "Rust"` are you OK using it? Specifically, are you OK assuming that `getrandom` and some custom RNG will both be compiled with the same Rust version?

If we aren't comfortable using the Rust ABI, we could always just use the C ABI and have code like:
```rust
    extern "C" {
        fn __getrandom_custom(dest: *mut u8, len: usize) -> u32;
    }
    let ret = unsafe { __getrandom_custom(dest.as_mut_ptr(), dest.len()) };
    match NonZeroU32::new(ret) {
        None => Ok(()),
        Some(code) => Err(Error(code)),
    }
}
```
and
```rust
macro_rules! register_custom_getrandom {
    ($path:path) => {
        #[no_mangle]
        fn extern "C" __getrandom_custom(dest: *mut u8, len: usize) -> u32 {
            let slice = unsafe { from_raw_parts_mut(dest, len) };
            match $path(slice) {
                Ok(()) => 0,
                Err(e) => e.code().get(),
            } 
        }
    };
}
```
It's a bit longer, but would still keep the details internal to `getrandom`.

newpavlov:
@josephlr 
>given the above feedback on extern "Rust" are you OK using it? Specifically, are you OK assuming that getrandom and some custom RNG will both be compiled with the same Rust version?

I think that the raised assumptions are reasonable, so personally I am OK with using `extern "Rust"`. But I will be a a tiny bit more comfortable with `extern "C"` just to be extra safe. IIUC it does not really have any drawbacks except being slightly more verbose and having two `unsafe` lines instead of none.

custom/stdweb/Cargo.toml
josephlr: We do need this. When we move the wasm/stdweb functionality to Custom RNGs, they can no longer be tested by `getrandom` main test code. These `[[test]]`s simply  restore the old testing behavior.

custom/stdweb/Cargo.toml
josephlr: We could, but then a users running `cargo test` on these repos would have things break.

These features only affect tests, so it's not that bad. Ideally, we would have a way to enable these features in tests. I'll look into https://github.com/rust-lang/cargo/issues/2911 and `required-features`.

custom/stdweb/Cargo.toml
josephlr: Done, also bumped `getrandom` to 0.2

custom/stdweb/src/lib.rs
josephlr: Done

custom/stdweb/Cargo.toml
josephlr: I changed the comments to clarify that these are non "unstable" but that they are only for tests. I'm actually fine with the features being stable so that "cargo test" just works.

We could use `required-features` here to skip the tests if you haven't enabled the features, but that would just make the tests appear to pass when the features aren't set. To actually solve this we would need https://github.com/rust-lang/cargo/issues/2911. Having the test-only features on by default seems like the easiest for users and doesn't hurt anything.

josephlr:
@newpavlov @dhardy I've rebased `0.2` onto latest master (to fix the CI) and rebased this PR onto that. I think I've addressed all current feedback and the CI is passing, PTAL 

> I think that the raised assumptions are reasonable, so personally I am OK with using `extern "Rust"`. But I will be a a tiny bit more comfortable with `extern "C"` just to be extra safe. IIUC it does not really have any drawbacks except being slightly more verbose and having two `unsafe` lines instead of none.

Done, switched the macro and the `custom` `getrandom_inner` to use `extern "C"`.

dhardy:
> I've rebased 0.2 onto latest master

This is supposed to be a dev branch, not a WIP PR; please don't rebase it (I should probably enable branch protections). A merge should suffice.

josephlr:
> This is supposed to be a dev branch, not a WIP PR; please don't rebase it (I should probably enable branch protections). A merge should suffice.

Sorry about that, I'll make sure to do that from here on out, and branch protections would be very reasonable.

src/lib.rs
```diff
@@ -251,6 +240,28 @@ cfg_if! {
     }
 }
 
+/// Reister a function to be invoked by `getrandom` on custom targets. This
+/// function will only be invoked on targets not supported by `getrandom`. This
+/// prevents crate dependancies from either inadvertantly or maliciously
```
newpavlov: Typos: reister -> Register, dependancies -> dependencies

src/lib.rs
```diff
@@ -251,6 +240,28 @@ cfg_if! {
...
+/// overriding the secure RNG implementations in `getrandom`.
+///
+/// *This API requires the following crate features to be activated: `custom`*
+#[macro_export]
+#[cfg(feature = "custom")]
+macro_rules! register_custom_getrandom {
+    ($path:path) => {
+        // We use an extern "C" function to get the garuntees of a stable ABI.
```
newpavlov: garuntees -> guarantees

tests/common.rs
```diff
@@ -1,19 +1,25 @@
-#[cfg(feature = "wasm-bindgen")]
+// Explictly use the Custom RNG crates to link them in.
```
newpavlov: Explictly -> Explicitly

josephlr: Fixed

src/lib.rs
josephlr: Fixed

custom/stdweb/Cargo.toml
newpavlov: You have misunderstood me, I suggested to replace `getrandom = { path = "../..", features = ["custom"] }` with `getrandom = { version = "0.2", path = "../..", features = ["custom"] }`. I am fine with `getrandom-*` crates not synchronizing versions with `getrandom`.

newpavlov:
@koute
Will you be interested in maintaining the `getrandom-stdweb` crate as part of the `stdweb` project? We could rename it to `stdweb-getrandom` to better match the naming scheme used in your project.

src/error.rs
```diff
@@ -17,15 +17,45 @@ use core::num::NonZeroU32;
...
+    pub const STDWEB_RNG_FAILED: Error = internal_error!(10);
+
     #[deprecated(since = "0.1.7")]
     /// Unknown error.
-    pub const UNKNOWN: Error = UNSUPPORTED;
+    pub const UNKNOWN: Error = Error::UNSUPPORTED;
     #[deprecated(since = "0.1.7")]
     /// System entropy source is unavailable.
-    pub const UNAVAILABLE: Error = UNSUPPORTED;
+    pub const UNAVAILABLE: Error = Error::UNSUPPORTED;
```
newpavlov: I think we can remove these two deprecated constants as part of this PR.

alexcrichton:
Yes if that's the desired course of action here I think it's reasonable to move the wasm-bindgen piece into the rustwasm org, and we'd be happy to maintain it.

src/lib.rs
```diff
@@ -251,6 +240,28 @@ cfg_if! {
...
 
+/// Register a function to be invoked by `getrandom` on custom targets. This
+/// function will only be invoked on targets not supported by `getrandom`. This
+/// prevents crate dependencies from either inadvertently or maliciously
+/// overriding the secure RNG implementations in `getrandom`.
+///
+/// *This API requires the following crate features to be activated: `custom`*
+#[macro_export]
+#[cfg(feature = "custom")]
+macro_rules! register_custom_getrandom {
```
dhardy: Note that Edition 2018 macro rules enable us to place this macro in the `custom` sub-module, then `pub use` it. We require Rustc >= 32 so I don't see any problems doing this, though the keeping it here is also fine (just slightly messier).

src/error.rs
josephlr: Done

src/lib.rs
josephlr: So we have to be careful, as the marco needs to always be exported, even if the custom implementation isn't used. Otherwise adding support for a target becomes a breaking change.

To get this working I did:

In `src/lib.rs`:
```rust
#[cfg(feature = "custom")]
mod custom;

// ...

cfg_if! {
    // ...
    } else if #[cfg(feature = "custom")] {
        use custom as imp;
    } else {
    // ...
}
```

In `src/custom.rs`:
```rust
#[macro_export]
macro_rules! register_custom_getrandom {
    // ...
}

#[allow(dead_code)]
    // ...
}
```

Does this seem reasonable?

custom/stdweb/Cargo.toml
josephlr: Fixed, I should have [read the docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies).

src/lib.rs
dhardy: Seems I mis-remembered how macros get exported ([silly example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7ad67c2ad78ff237d79545259c6a3bbc)). So this is as good as we can get.

Any idea why we need `#[alllow(dead_code)]` now?

josephlr: `getrandom_inner` isn't always called. If you have `custom` selected on a supported target, you'll get a dead code warning.

josephlr:
I've rebased this onto #120 and improved the commit messages to be more readable. I also renamed the Custom RNG implementations to be `wasm-bindgen-getrandom` and `stdweb-getrandom` (to reflect their potential future ownership). I like the idea of (post 0.2 release) moving the Custom RNG crates to their respective owners.

I've addressed all of the outstanding feedback, so I think the only thing left here is to update the module documentation and `README.md`.

src/lib.rs
dhardy: Sorry, I get it now.

So, this change results in slightly more code overall, but does but both parts of this `extern "C"` interface within the same module and make `lib.rs` smaller, so is worthwhile IMO. :+1:

newpavlov:
@josephlr 
Can you resolve the conflicts? After that I think we should be good to merge.

PR #114, opened 2019-09-30
Fix and Improve WASM CI

### LINKED ISSUE #113: CI is broken for wasm-bindgen
This is currently blocking https://github.com/rust-random/getrandom/pull/112

The CI error message [here](https://travis-ci.org/rust-random/getrandom/jobs/591702671) actually gives a fairly good description:

```
Currently the bindgen format is unstable enough that these two version must
exactly match, so it's required that these two version are kept in sync by
either updating the wasm-bindgen dependency or this binary.
```

The solution here is simple, just always download the latest binary from Github instead of pinning the version.

josephlr:
Fixes #113 

- Always use latest Firefox
  - We have to use latest Chrome, so we might as well do the same thing here.
- Explicitly pin `wasmtime` and `cargo-web`
  - We don't need a particular version here, so we might as well pin to avoid CI breakage.
- Make sure our `wasm-bindgen-test-runner` and `wasm-bindgen` versions match
  - This uses the `cargo metadata` command (which is an awesome thing I just learned about).

We now consistently use the `$VERSION` environment variable to make our intentions explicit (and to avoid repeating version numbers).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,9 +32,6 @@ matrix:
       addons:
         firefox: latest
         chrome: stable
-        apt:
-          packages:
-          - jq
       install:
         - rustup target add wasm32-unknown-unknown
         - rustup target add wasm32-unknown-emscripten
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,29 +30,36 @@ matrix:
     - name: "WASM via emscripten, stdweb, wasm-bindgen and WASI"
       rust: nightly
       addons:
-        firefox: 69.0
+        firefox: latest
         chrome: stable
+        apt:
+          packages:
+          - jq
       install:
         - rustup target add wasm32-unknown-unknown
         - rustup target add wasm32-unknown-emscripten
         - rustup target add asmjs-unknown-emscripten
         - rustup target add wasm32-wasi
-        # Get geckodriver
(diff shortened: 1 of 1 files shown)


josephlr:
Hey @newpavlov @dhardy I still have Write access to this repo, but I wasn't able to merge this PR ("You’re not authorized to merge this pull request.") even though it was approved. Is this intentional, or did something change? I was able to merge https://github.com/rust-random/getrandom/pull/108 just fine a few days ago.

For the Google repos I maintain, I set the branch protection rules to be:
  - `master` is protected, so no force pushing
  - All PRs against `master` require approval
  - PRs require approval from at least one other user with Write permission
  - Once approved, anyone with Write access can merge the PR 

I thought that's what `getrandom` had as well, but it seems like that's not the case.

dhardy:
Sorry @josephlr, that was me; I demoted you from "Maintain" to Write access; according to the description this should still let you manage PRs so I don't know why this happened.

I checked the branch protection (merge review, push, and now status checks); these shouldn't be an issue. Hmm, I'll bump you back to Maintain then.

josephlr:
> Sorry @josephlr, that was me; I demoted you from "Maintain" to Write access; according to the description this should still let you manage PRs so I don't know why this happened.

That's really weird, I agree that the docs suggest I _should_ only need Write permission to merge approved PRs. It might be a bug in Github's new permissions settings, or just some unclear docs. I haven't dealt that much with Triage/Maintain permissions as the new permissions Beta hasn't rolled out to the Google org yet.

PR #116, opened 2019-10-24
Merge branch 'master' into '0.2'

josephlr:
Not too bad on merge conflicts, only one was in `src/lib.rs`. Fixed by using the fact that [`vxworks` is in the `unix` family](https://github.com/rust-lang/rust/blob/55e00631e5bc5b16d40232914e57deeea197a8e4/src/librustc_target/spec/vxworks_base.rs#L26).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,7 +18,7 @@
 //! | iOS              | [`SecRandomCopyBytes`][4]
 //! | FreeBSD          | [`getrandom()`][21] if available, otherwise [`kern.arandom`][5]
 //! | OpenBSD          | [`getentropy`][6]
-//! | NetBSD           | [`/dev/urandom`][7] after successfully polling `/dev/random`
+//! | NetBSD           | [`kern.arandom`][7]
 //! | Dragonfly BSD    | [`/dev/random`][8]
 //! | Solaris, illumos | [`getrandom`][9] system call if available, otherwise [`/dev/random`][10]
 //! | Fuchsia OS       | [`cprng_draw`][11]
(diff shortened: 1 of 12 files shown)


newpavlov:
I think with this PR you are trying to merge '0.2' into `master`, not the other way around. We probably should release `v0.1.13` with the VxWorks and NetBSD changes first. And after that it could be reasonable to merge `0.2` into `master`.

src/lib.rs
```diff
@@ -161,24 +159,20 @@ pub use crate::error::Error;
...
-#[cfg(target_os = "vxworks")]
+#[cfg(all(
+    unix,
+    not(any(
+        target_os = "ios",
+        target_os = "fuchsia",
+        target_os = "hermit",
+        target_os = "l4re"
+    ))
+))]
```
newpavlov: BTW personally I would prefer to use an explicit enumeration of targets. Size-wise it will not be that bad:
```rust
#[cfg(any(
    target_os = "android", target_os = "dragonfly", target_os = "emscripten",
    target_os = "freebsd", target_os = "haiku",     target_os = "illumos",
    target_os = "linux",   target_os = "macos",     target_os = "netbsd",
    target_os = "openbsd", target_os = "redox",     target_os = "solaris",
    target_os = "vxworks",
))]
#[allow(dead_code)]
mod util_libc;
```

josephlr:
> I think with this PR you are trying to merge `0.2` into `master`, not the other way around. We probably should release `v0.1.13` with the VxWorks and NetBSD changes first. And after that it could be reasonable to merge `0.2` into `master`.

Ya I selected the wrong base, this is supposed to just update `0.2` to incorporate the latest changes from `master`. We should wait a bit before merging `0.2` into `master`.

josephlr:
@newpavlov this should probably wait on #117 to merge first.

src/lib.rs
josephlr: I agree, but think this should be fixed in a separate `0.2` PR, I'll open one when this gets merged.

josephlr: Here's the basic idea: https://github.com/josephlr/getrandom/commit/e47483b3a82481d2ed7c544284584708510b08ac

PR #119, opened 2019-10-24
Improve src/lib.rs cfgs

josephlr:
Suggested by @newpavlov here: https://github.com/rust-random/getrandom/pull/116#discussion_r338433563

Right now for each of `util_libc` and `use_file` we have a list of
`target_os` configs to determine if we build the module.

This PR moves these mod declarations into the main `cfg_if` statement
(the one that selects which implementation we use). This way, the mod
statements are kept in-sync with the implementations that use them.

Also, I merged together `target_os` cfgs that have the same
implementation. The downside to this is that the targets are no longer
in alphabetical order.

Also, this is only being applied to `0.2` as the `0.1` cfgs still have
to keep `std` around.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -154,73 +154,49 @@ cfg_if! {
 }
 
 mod error;
-pub use crate::error::Error;
-
-#[allow(dead_code)]
 mod util;
-
-#[cfg(all(
-    unix,
-    not(any(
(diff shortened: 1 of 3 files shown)


newpavlov:
>Suggested by @newpavlov here

Well, to be fair the main idea is yours. :)

PR #120, opened 2019-10-26
0.2: Error and Testing improvements

josephlr:
#109 contains a lot of fixes to the `Error` class and to our testing infrastructure that makes it harder to focus on the important changes in that PR.

I split out the changes unrelated to Custom RNGs into this PR. I also fixed up the commits/messages to explain these changes (we should be sure to "merge" this PR not "squash" it). Also, this PR should be merged before #109

This PR:

- Improves the `Error` struct
  - Removes deprecated constants
  - Exposes new `Error` constants
  - Some other minor nits
- Improves our Testing code
  - Remove use of `--examples`, we do not have examples
  - Install drivers to $HOME
  - Use YAML expansion to avoid repeating testing configs
  - Ensure crates build with each feature set individually
  - Add missing targets to cross-platform build step
  - Check minimum version compatiblity with features _on_
  - Gerate docs with "std" feature set
- Makes sure that the `std` feature will be enabled when using [docs.rs](https://docs.rs/).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,6 @@ matrix:
         - rustup target add aarch64-apple-ios
       script:
         - cargo test
-        - cargo test --examples
         - cargo build --target aarch64-apple-ios
 
     - name: "Linux, beta"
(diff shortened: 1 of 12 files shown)


newpavlov:
BTW maybe we should feature-gate match arms in the `internal_desc` depending on the platform (or even constants themselves)? Otherwise all error description strings will get compiled on all platforms, even though most of them will never be used. Yes, it will be a really small improvement of a final binary size, but I guess every bit counts.

josephlr:
> Looks good to me.
> 
> One addition I'd like to see (@newpavlov?) is to prefix more of these error constants (e.g. `RAND_SECURE_FATAL`) with the platform name (in this case VxWorks, which is mentioned in the error message).

Done, changed to `IOS_SEC_RANDOM`, `WINDOWS_RTL_GEN_RANDOM`, `VXWORKS_RAND_SECURE`. I also updated the error messages/docs



> BTW maybe we should `cfg` match arms in the `internal_desc` depending on the platform (or even constants themselves)?

So removing the strings makes the release binary smaller, but removing the constants themselves should only make the `rlib` smaller but have no affect on the final binary size (as they are constants, they don't have a place in memory).

I don't think we should remove the constants. As for removing the strings, doing it correctly for all of them is tricky (involving some gross `cfg-if`). I think we should discuss binary size reduction in a separate PR, and decide which metrics we care about.

> And I think we don't use `From<io::Error>` impl anymore, so we probably can remove it together with `UNKNOWN_IO_ERROR`.

Done, removed.

@dhardy, @newpavlov if you like the changes, feel free to merge this (once the CI is green).

dhardy:
> I don't think we should remove the constants. As for removing the strings, doing it correctly for all of them is tricky (involving some gross cfg-if). I think we should discuss binary size reduction in a separate PR, and decide which metrics we care about.

Agreed. Duplicating `cfg-if` spaghetti sounds like something best saved until later (potentially never, unless someone can make a good argument for it, but feel free to open an issue).

Changes look good to me; merge if you're happy @newpavlov.

PR #122, opened 2019-10-27
Readme draft

0xSilene:
New to pull requests / git so sorry about the mistakes. Formatting of the Readme Code had problems. Not sure if all my commits will show or not. Sorry about that.

I added some more documentation for using getrandom with vectors and passing in parameters for the bytes generated.


dhardy:
I'm glad you worked out how to fill a vector using a function which takes a mutable slice argument.
I don't see how this helps users however — this crate is aimed at doing a low-level job well with no extras, and is only really intended for use by people who already know how to do this.

The [rand](https://github.com/rust-random/rand) crate does all those extras, including converting to a `u32` or `f64`, unbiased sampling from a range, shuffling a list, and making all this fast, thread- and fork-safe.

I don't mean to be unfriendly, but suggest you should probably start reading `rand` examples or [the book](https://rust-random.github.io/book/).

PR #125, opened 2020-01-02
use_file: Remove use of spin-locks

josephlr:
Remove the general purpose spin-lock from getrandom, and don't spin
when polling /dev/random. We also remove the use of spin locks when
opening the persistent fd for platforms that require it.

For both these cases, we can just use the pthread lock/unlock methods
in libc. We also do some minor cleanup to better make use of `Result`
types and `DropGuard`s.

Essentially, this change does the "standard" Mutex based synchronization approach, it just has to do it without `libstd`. With this change, we continue to have the property that `getrandom` uses at most one file descriptor per process.

**EDIT:** Updated description to indicate that we are _removing_ all uses of spin locks.

Thanks to @matklad for pointing this out. See [his blog post about `getrandom`](https://matklad.github.io//2020/01/02/spinlocks-considered-harmful.html) and https://github.com/matklad/once_cell/issues/61.

### DIFF
--- a/src/use_file.rs
+++ b/src/use_file.rs
@@ -7,8 +7,11 @@
 // except according to those terms.
 
 //! Implementations that just need to read from a file
-use crate::util_libc::{last_os_error, open_readonly, sys_fill_exact, LazyFd};
+use crate::util::LazyUsize;
+use crate::util_libc::{open_readonly, sys_fill_exact};
 use crate::Error;
+use core::cell::UnsafeCell;
+use core::sync::atomic::{AtomicUsize, Ordering::Relaxed};
 
 #[cfg(target_os = "redox")]
 const FILE_PATH: &str = "rand:\0";
@@ -21,10 +24,11 @@ const FILE_PATH: &str = "rand:\0";
     target_os = "illumos"
 ))]
 const FILE_PATH: &str = "/dev/random\0";
+#[cfg(any(target_os = "android", target_os = "linux"))]
+const FILE_PATH: &str = "/dev/urandom\0";
 
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
-    static FD: LazyFd = LazyFd::new();
-    let fd = FD.init(init_file).ok_or_else(last_os_error)?;
+    let fd = get_rng_fd()?;
     let read = |buf: &mut [u8]| unsafe { libc::read(fd, buf.as_mut_ptr() as *mut _, buf.len()) };
 
     if cfg!(target_os = "emscripten") {
@@ -38,36 +42,96 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
     Ok(())
 }
 
-cfg_if! {
-    if #[cfg(any(target_os = "android", target_os = "linux"))] {
-        fn init_file() -> Option<libc::c_int> {
-            // Poll /dev/random to make sure it is ok to read from /dev/urandom.
-            let mut pfd = libc::pollfd {
-                fd: unsafe { open_readonly("/dev/random\0")? },
-                events: libc::POLLIN,
-                revents: 0,
-            };
+// Returns the file descriptor for the device file used to retrieve random
+// bytes. The file will be opened exactly once. All successful calls will
+// return the same file descriptor. This file descriptor is never closed.
+fn get_rng_fd() -> Result<libc::c_int, Error> {
+    static FD: AtomicUsize = AtomicUsize::new(LazyUsize::UNINIT);
+    fn get_fd() -> Option<libc::c_int> {
+        match FD.load(Relaxed) {
+            LazyUsize::UNINIT => None,
+            val => Some(val as libc::c_int),
+        }
+    }
 
-            let ret = loop {
-                // A negative timeout means an infinite timeout.
-                let res = unsafe { libc::poll(&mut pfd, 1, -1) };
-                if res == 1 {
-                    break unsafe { open_readonly("/dev/urandom\0") };
-                } else if res < 0 {
-                    let e = last_os_error().raw_os_error();
-                    if e == Some(libc::EINTR) || e == Some(libc::EAGAIN) {
-                        continue;
-                    }
-                }
-                // We either hard failed, or poll() returned the wrong pfd.
-                break None;
-            };
-            unsafe { libc::close(pfd.fd) };
-            ret
+    // Use double-checked locking to avoid acquiring the lock if possible.
+    if let Some(fd) = get_fd() {
+        return Ok(fd);
+    }
+
+    // SAFETY: We use the mutex only in this method, and we always unlock it
+    // before returning, making sure we don't violate the pthread_mutex_t API.
+    static MUTEX: Mutex = Mutex::new();
+    unsafe { MUTEX.lock() };
+    let _guard = DropGuard(|| unsafe { MUTEX.unlock() });
+
+    if let Some(fd) = get_fd() {
+        return Ok(fd);
+    }
+
+    // On Linux, /dev/urandom might return insecure values.
+    #[cfg(any(target_os = "android", target_os = "linux"))]
+    wait_until_rng_ready()?;
+
+    let fd = unsafe { open_readonly(FILE_PATH)? };
+    // The fd always fits in a usize without conflicting with UNINIT.
+    debug_assert!(fd >= 0 && (fd as usize) < LazyUsize::UNINIT);
+    FD.store(fd as usize, Relaxed);
+
+    Ok(fd)
+}
+
+// Succeeds once /dev/urandom is safe to read from
+#[cfg(any(target_os = "android", target_os = "linux"))]
+fn wait_until_rng_ready() -> Result<(), Error> {
+    // Poll /dev/random to make sure it is ok to read from /dev/urandom.
+    let fd = unsafe { open_readonly("/dev/random\0")? };
+    let mut pfd = libc::pollfd {
+        fd,
+        events: libc::POLLIN,
+        revents: 0,
+    };
+    let _guard = DropGuard(|| unsafe {
+        libc::close(fd);
+    });
+
+    loop {
+        // A negative timeout means an infinite timeout.
+        let res = unsafe { libc::poll(&mut pfd, 1, -1) };
+        if res >= 0 {
+            assert_eq!(res, 1); // We only used one fd, and cannot timeout.
+            return Ok(());
         }
-    } else {
-        fn init_file() -> Option<libc::c_int> {
-            unsafe { open_readonly(FILE_PATH) }
+        let err = crate::util_libc::last_os_error();
+        match err.raw_os_error() {
+            Some(libc::EINTR) | Some(libc::EAGAIN) => continue,
+            _ => return Err(err),
         }
+    }
+}
+
+struct Mutex(UnsafeCell<libc::pthread_mutex_t>);
+
+impl Mutex {
+    const fn new() -> Self {
+        Self(UnsafeCell::new(libc::PTHREAD_MUTEX_INITIALIZER))
+    }
+    unsafe fn lock(&self) {
+        let r = libc::pthread_mutex_lock(self.0.get());
+        debug_assert_eq!(r, 0);
+    }
+    unsafe fn unlock(&self) {
+        let r = libc::pthread_mutex_unlock(self.0.get());
+        debug_assert_eq!(r, 0);
+    }
+}
+
+unsafe impl Sync for Mutex {}
+
+struct DropGuard<F: FnMut()>(F);
+
+impl<F: FnMut()> Drop for DropGuard<F> {
+    fn drop(&mut self) {
+        self.0()
     }
 }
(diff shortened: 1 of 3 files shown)


src/use_file.rs
```diff
@@ -38,36 +51,93 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
...
+        // read() or poll() syscalls, the open() syscall will not block if the
+        // file is a device. This is true even if the kernel's entropy pool
+        // isn't initialized. In that case, only reads will block.
+        loop {
+            // Relaxed ordering is fine, as we only have a single atomic variable.
+            match self.0.compare_and_swap(Self::EMPTY, Self::OPENING, Relaxed) {
+                Self::EMPTY => {
+                    let fd = unsafe { open_readonly(FILE_PATH) };
+                    let val = match fd {
+                        // OK as val >= 0 and val <= c_int::MAX < Self::OPENING
```
matklad: Comment could be changed to an assert

josephlr: Done

josephlr:
> > This is fine, as the open() syscall will never block when opening a device
> 
> This is better, but, in theory, and if you are **really** unlucky, can still lead to a priority inversion, if the thread is preempted.

I agree, after looking over the code, I think removing spinning entirely is the best bet (and also makes the code more readable). In the very rare case we race on the `open` call, it's fine to open `/dev/urandom` twice and then just close one.

matklad:
It just occurred to me that if we are using the open syscall here then we ... can just call pthread_lock as well?

josephlr:
> It just occurred to me that if we are using the open syscall here then we ... can just call pthread_lock as well?

I looked into doing that originally, and then again when I was fixing this issue. It seems much more complicated to do than the current code. There are a bunch of pthread methods you have to call (or at least the stdlib calls them, I'm not sure the complexity (vs this PR) is worth it.

matklad:
>There are a bunch of pthread methods you have to call (or at least the stdlib calls them

stdlib needs to guard agains reentrant locking, which we don't have to, so the implementation is pretty straight forward (modulo the lack of `UnsafeSyncCell`):

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=053aa21299f80833e3ee779a05bf55c5

I like it because it doges the other edge case -- unbounded number of concurrently opened file descriptors (not that my assertion from the blog that an init functions would be run at most #cores times is wrong).

matklad:
>stdlib needs to guard agains reentrant locking, which we don't have to

Well, which I *think* we don't need to do, for the following reasons:

* we don't run any user code while holding the mutex (in particular, no closure, generics or custom `Drop`s are in sight)
* so the only reason why we might reenter the function is signal handlers, but we don't advertise `getrandom` as async safe

gnzlbg:
> I like it because it doges the other edge case -- unbounded number of concurrently opened file descriptors

I was just going to mention this. File-descriptors are a finite resource on most OSes, and I'd be more comfortable with a solution that does not exhaust them.

josephlr:
@matklad @gnzlbg I think you are both right. The code to use pthreads wasn't that bad (after I added a `DropGuard` helper type) and it avoids some weird pathological edge cases.

Let me know what you think.

src/use_file.rs
```diff
@@ -38,36 +41,83 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
...
-            let mut pfd = libc::pollfd {
-                fd: unsafe { open_readonly("/dev/random\0")? },
-                events: libc::POLLIN,
-                revents: 0,
-            };
+// Returns the file descriptor for the device file used to retrieve random
+// numbers. The file will be opened exactly once. All successful calls will
+// return the same file descriptor. This file descriptor is never closed.
+fn get_rng_fd() -> Result<libc::c_int, Error> {
+    static FD: AtomicUsize = AtomicUsize::new(LazyUsize::UNINIT);
```
gnzlbg: I'm not sure how often this needs to happen, or whether this happening is worth optimizing, but a different way to implement this is to do something like this to keep the common path branchless: https://github.com/calebzulawski/multiversion/pull/6/files#diff-0a878480aac95b54dd822d02c4ad345eR76 

cc @TethysSvensson - it might be a thing worth attempting in a subsequent PR after an approach that works "correctly" gets merged.

TethysSvensson: @gnzlbg The reason why that trick works in multiversion is, that the value we are trying to lazily compute is already a function pointer. This means that the best-case performance is always going to involve **at least**  1 indirect function call. The trick makes sure that the common path does not pay anything **in addition** to that 1 indirect function.

Here it looks like you are trying to lazily compute a file descriptor. As far as I can tell, the current cost in the common path is a single conditional branch on a relaxed load. This appears to me to be cheaper than alternative cost of 1 indirection function call on a relaxed load.

TethysSvensson: I do have one suggestion though, assuming that you would ever want to inline `get_rng_fd`. I think in that case, you would want to inline the initial check, while keeping the slow initializer un-inlined. I am imagining something like this:

```rust
#[inline]
fn get_rng_fd() -> Result<libc::c_int, Error> {
    static FD: AtomicUsize = AtomicUsize::new(LazyUsize::UNINIT);

    if let Some(fd) = get_fd() {
        Ok(fd)
    } else {
        initialize_slow()
    }

    #[inline(always)]
    fn get_fd() -> Option<libc::c_int> { ... }
    #[inline(always)]
    fn initialize_slow() -> Result<libc::c_int, Error> { ... }
}
```

However if I understand this code correctly, the I don't think this function is meant to be inlined(?), so in that case it does not really matter either way.

josephlr: > However if I understand this code correctly, the I don't think this function is meant to be inlined(?), so in that case it does not really matter either way.

This is correct, the only external function of this crate is `getrandom::getrandom` which is not marked `#[inline]`, so I don't think marking internal functions `#[inline]` really does anything.

In fact it looks like (in release mode) the compiler just inlines everything in this file into `use_file::getrandom_inner`. I checked with the current nightly build.

src/use_file.rs
```diff
@@ -38,36 +41,83 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
...
+    }
+
+    // Use double-checked locking to avoid acquiring the lock if possible.
+    if let Some(fd) = get_fd() {
+        return Ok(fd);
+    }
+
+    // SAFETY: Only a raw pointer is created from MUTEX to pass to libc.
+    static mut MUTEX: libc::pthread_mutex_t = libc::PTHREAD_MUTEX_INITIALIZER;
+    let r = unsafe { libc::pthread_mutex_lock(&mut MUTEX as *mut _) };
```
matklad: I am not sure that `&mut MUTEX as *mut _` is formally sound. Here, we create a mutable reference to `MUTEX` *before* locking it, so, at runtime, we can get aliasing. The guaranteed ok solution is to put a mutex into a `Cell` / `UnsafeCell`, and store that in a non-mut static. `UnsafeCell` is not `Sync` though, so you'll need a custom wrapper struct to assert `Sync` (like I did in the playground link).

See also https://github.com/rust-lang/rust/issues/53639

dhardy: Isn't the purpose of mutexes to handle aliasing — i.e. that one needs to obtain a `*mut _` local to the thread (and aliasing the equivalent in any other thread) in order to use the mutex?

dhardy: Aha, the [`UnsafeCell`](https://doc.rust-lang.org/std/cell/struct.UnsafeCell.html) docs clarify:

> The compiler makes optimizations based on the knowledge that &T is not mutably aliased or mutated, and that &mut T is unique.

>  The UnsafeCell API itself is technically very simple: it gives you a raw pointer *mut T to its contents. It is up to you as the abstraction designer to use that raw pointer correctly.

So it wrapping via `UnsafeCell` just allows us to skip the `&mut _` part.

josephlr: This should be fixed now. I added a `Mutex` abstraction that wraps an `UnsafeCell`. By adding `lock`/`unlock` methods to this `Mutex`, the code also ends up looking much cleaner.

src/use_file.rs
```diff
@@ -38,36 +42,98 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
...
+
+    loop {
+        // A negative timeout means an infinite timeout.
+        let res = unsafe { libc::poll(&mut pfd, 1, -1) };
+        if res == 1 {
+            return Ok(());
         }
+        let err = last_os_error();
+        if res >= 0 {
+            return Err(err); // poll() returned the wrong pfd
```
matklad: if `res >=`, the `err` does not make sense, no? But this was probably per-existing.

src/util_libc.rs
```diff
@@ -129,15 +98,15 @@ cfg_if! {
 }
 
 // SAFETY: path must be null terminated, FD must be manually closed.
-pub unsafe fn open_readonly(path: &str) -> Option<libc::c_int> {
+pub unsafe fn open_readonly(path: &str) -> Result<libc::c_int, Error> {
     debug_assert!(path.as_bytes().last() == Some(&0));
     let fd = open(path.as_ptr() as *mut _, libc::O_RDONLY | libc::O_CLOEXEC);
```
matklad: unrelated to the PR, but

```suggestion
    let fd = open(path.as_ptr() as *const _, libc::O_RDONLY | libc::O_CLOEXEC);
```

?

src/use_file.rs
josephlr: Good catch. We should just `assert` that the return value is 1 if `poll(2)` succeeds. [Per the man page](http://man7.org/linux/man-pages/man2/poll.2.html), this is the only value we should get give the parameters we are feeding in.

src/util_libc.rs
josephlr: Fixed.

josephlr:
@dhardy @newpavlov This PR is ready for review/merging. Comments have been addressed, and the CI is passing.

`get_rng_fd` is the main functional change, most of the remaining changes are either nits or code moving around. I'd either "rebase" or "merge" this PR as to not lose the commit history (where I tried to split up this change).

src/use_file.rs
```diff
@@ -38,36 +42,96 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
...
+    // Use double-checked locking to avoid acquiring the lock if possible.
+    if let Some(fd) = get_fd() {
+        return Ok(fd);
+    }
+
+    // SAFETY: We use the mutex only in this method, and we always unlock it
+    // before returning, making sure we don't violate the pthread_mutex_t API.
+    static MUTEX: Mutex = Mutex::new();
+    unsafe { MUTEX.lock() };
+    let _guard = DropGuard(|| unsafe { MUTEX.unlock() });
```
newpavlov: Is it guaranteed that destructor will run right before function exit (be it `return` or panic)? For some reason I thought that with NLL drop place can change depending on liveness analysis.

josephlr: NLL (and its future extensions w/ Polonious or what not) is just for Lifetimes. `Drop` is still scope based, so the `drop` method is always executed when the object leaves scope. As the guard is at function scope, it `drop`s on `return` or on `panic` unwinding.

The liveness analysis is just used to determine if a set of borrows are permitted. This is why (modulo compiler bugs) NLL was introduced without it being a breaking change.

Docs: https://doc.rust-lang.org/book/ch15-03-drop.html

src/use_file.rs
```diff
@@ -38,36 +42,96 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
...
+
+struct Mutex(UnsafeCell<libc::pthread_mutex_t>);
+
+impl Mutex {
+    const fn new() -> Self {
+        Self(UnsafeCell::new(libc::PTHREAD_MUTEX_INITIALIZER))
+    }
+    unsafe fn lock(&self) {
+        let r = libc::pthread_mutex_lock(self.0.get());
+        debug_assert_eq!(r, 0);
```
dhardy: Should we not check the return value in all builds? (Also `unlock`.)

src/use_file.rs
```diff
@@ -38,36 +42,96 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
...
+    }
+    unsafe fn unlock(&self) {
+        let r = libc::pthread_mutex_unlock(self.0.get());
+        debug_assert_eq!(r, 0);
+    }
+}
+
+unsafe impl Sync for Mutex {}
+
+struct DropGuard<F: FnMut()>(F);
```
dhardy: I'm surprised this isn't a part of the `core` library!

src/use_file.rs
matklad: stdlib also uses `debug_assert`, and man page specifically says that lock/unlock basically can't fail.

src/use_file.rs
matklad: Often times, this does not work in higher level code, because closure borrows the environment, and so you can't modify it. We only able to use it because we close over low-level file descriptor, which is `Copy`. 

To be clear, it is an oftentimes useful utility, but it is not as useful in practice as it could seem.

src/use_file.rs
dhardy: Aha. Closer reading of the possible error values shows that none should be applicable in this case, and if any were that would be a code error, so okay. (Assuming `EAGAIN` is specific to recursive locks.)

PR #126, opened 2020-01-02
wasi: Update and Fix Travis failures

josephlr:
It looks like Rust is now shipping a different version of the WASI toolchain. This toolchain now requires the corresponding runner to have `wasi_snapshot_preview1` as part of its environment.

Updating the version of the wasi runner to `0.8` allows us to run the binaries created by newer Rust toolchains. See: https://github.com/bytecodealliance/wasi/issues/37

We also update the `wasi` version to `0.9` to get on a stable interface. This requires us to use `unsafe` again (as the current API does not have slice-based helpers anymore). The new API also gets ride of `error_str`, so we have to go though `wasi::Error`.  

CC: @sunfishcode

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ matrix:
         - gunzip cargo-web.gz
         - chmod +x cargo-web
         # Get wasmtime
-        - export VERSION=v0.3.0 # Pin version for stability
+        - export VERSION=v0.8.0 # Pin version for stability
         - wget -O wasmtime.tar.xz https://github.com/CraneStation/wasmtime/releases/download/$VERSION/wasmtime-$VERSION-x86_64-linux.tar.xz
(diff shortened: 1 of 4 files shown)


sunfishcode:
Could you provide more details? I'm able to build this crate and run all the tests with wasm32-wasi with both stable and nightly Rust.

josephlr:
> Could you provide more details? I'm able to build this crate and run all the tests with wasm32-wasi with both stable and nightly Rust.

Sorry it's not building that's a problem, it's running the code. I keep getting a "Import module wasi_snapshot_preview1 was not found" even when using the never versions of wasi (like 0.9.0).

josephlr:
> I keep getting a "Import module wasi_snapshot_preview1 was not found" even when using the never versions of wasi (like 0.9.0).

Looks like the issue was that the runner version (not the library version) which was causing issues. Let's see if this fixes things.

Issue #124, opened 2019-12-30
Compilation failure on DragonFlyBSD


paragw:
error[E0432]: unresolved import `libc::__error`
  --> /home/vmuser/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.1.13/src/util_libc.rs:22:13
   |
22 |         use libc::__error as errno_location;
   |             ^^^^^^-------^^^^^^^^^^^^^^^^^^
   |             |     |
   |             |     help: a similar name exists in the module: `ferror`
   |             no `__error` in the root

error: aborting due to previous error

For more information about this error, try `rustc --explain E0432`.
error: Could not compile `getrandom`.

josephlr:
I was able to reproduce this with `cargo xbuild --target x86_64-unknown-dragonfly`. For context, we attempted to get `errno` working correctly on DragonFlyBSD via https://github.com/rust-lang/libc/pull/1432 but that only allows it to work when `libc` is a dep of `libstd`.  The is because:
  - DragonflyBSD (unlike every other UNIX target) uses a thread-local static variable directly for its errono (as opposed to going through a helper method like `__error` or `_errno_locaiton`.
  - Rust thread-local statics are not stable, see https://github.com/rust-lang/rust/issues/29594

So we will probably just not bother with getting errno information on Dragonfly, which should fix the issue.

PR #128, opened 2020-01-05
Prepare release v0.1.14

newpavlov:
Ideally it would be nice to include changes in #125 into this release as well.


josephlr:
It think it's reasonable to get a release containing #126 and #125. I might take a look at fixing #124 before the next release.

PR #131, opened 2020-01-09
log: Remove optional log dependancy

josephlr:
This feature isn't enabled by rand/rand_core and provides very little
error information that isn't already conveyed through our Error values.

This also simplifies the supported configuration space for getrandom.

See this comment by @newpavlov: https://github.com/rust-random/getrandom/issues/98#issuecomment-546977404

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,14 +91,13 @@ matrix:
         - cargo test --benches
         # Check that setting various features does not break the build
         - cargo build --features=std
-        - cargo build --features=log
         # remove cached documentation, otherwise files from previous PRs can get included
         - rm -rf target/doc
         - cargo doc --no-deps --features=std
         - cargo deadlinks --dir target/doc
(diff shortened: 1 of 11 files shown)


josephlr:
> Concern: the Windows case now never reports the raw NTSTATUS code. However, relying on the logging system for this was never an ideal solution anyway, so it's not sufficient reason to avoid this PR.

I think this is the only significant ase where we actually took advantage of `log`'s functionality, but I don't think this Windows method will actually return a warning, so we probably weren't getting much out of it regardless.

dhardy:
Correct, it seems that MS do not document what the error codes are. OTTH [this guy does](https://davidvielmetter.com/tips/ntstatus-error-code-list/) and it means people can't then search for the error correctly. But for the most part this probably isn't really an issue (RNGs don't usually fail, and if it does, then a Rust app is probably not going to be the only one to notice).

PR #133, opened 2020-02-13
Add Cargo feature for CPU randomness

### LINKED ISSUE #84: Feature for using CPU-based RNG

josephlr:
Closes #84 

## `cpu` Cargo feature
Many CPU architectures support instructions to get randomness:
- `x86`/`x86_64` have the well-known [RDRAND/RDSEED](https://en.wikipedia.org/wiki/RDRAND)
- ARMv8.5-RNG recently added [RNDR](https://developer.arm.com/docs/ddi0595/c/aarch64-system-registers/rndr) ([LLVM patch](https://reviews.llvm.org/D52481), [Linux patch](https://lwn.net/Articles/809929/))
- PowerPC ([version 3.0B of the ISA](https://openpowerfoundation.org/?resource_lib=power-isa-version-3-0), i.e. POWER9 and later) has DARN (Deliver A Random Number)  
- [S390 supports PRNO](https://fossies.org/linux/binutils/opcodes/s390-opc.txt) (perform PseudoRandom Number Operation) [Linux code](https://elixir.bootlin.com/linux/v5.5.4/source/arch/s390/include/asm/cpacf.h#L443)

This PR adds a `cpu` Cargo feature, which calls those instructions if there is not a supported implementation. Right now we only support `x86_64`. This `cpu` feature is on by default (as there's no real reason to disable it, other than wanting to make sure all buildable targets have a concrete implementations). This feature takes lower priority than `custom` implementations. This is necessary to have specific RNG implementations on unsupported platforms.

We can now drop explicit support for targets where we used RDRAND simply because we didn't know what else to use. This includes:
- `x86_64-unknown-hermit`
- `x86_64-unknown-l4re-uclibc`
- `x86_64-unknown-uefi`

As the `cpu` feature is on by default, these targets will keep working by default, but they can now be overridden by `custom` implementations.

We continue to _explicitly_ support all SGX targets. This means that `custom` implementations _cannot_, override the RDRAND implementation. Intel makes it very clear you have to use RDRAND to get randomness on SGX. The other targets don't have this mandate. For example, UEFI could get it's randomness form either RDRAND or [EFI_RNG_PROTOCOL](https://docs.microsoft.com/en-us/windows-hardware/drivers/bringup/efi-rng-protocol).

## Testing improvements

 I also took some time to fix the test structure. Now, there is a `test_common` module, which various test files import. This still allows for us to write tests once, but removes the need for hacky Cargo features. I then cleaned up some of the testing in `.travis.yml`. I also removed duplicate tests for minimum dependency versions.

We also now test the CPU-based RNG implementation on all x86 targets. This makes it possible to directly test on Linux, instead of needing a custom target.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,16 @@
 language: rust
-sudo: false
+os: linux
 
-matrix:
+# Targets that we just build (rather than run and test)
+env:
+  global:
+    - STD_TARGETS="x86_64-sun-solaris x86_64-unknown-cloudabi x86_64-unknown-freebsd x86_64-fuchsia x86_64-unknown-netbsd x86_64-unknown-redox x86_64-fortanix-unknown-sgx"
+    - NO_STD_TARGETS="x86_64-unknown-uefi x86_64-unknown-hermit x86_64-unknown-l4re-uclibc x86_64-uwp-windows-gnu x86_64-wrs-vxworks"
+
+jobs:
   include:
     - name: "Linux, 1.32.0"
       rust: 1.32.0
-      os: linux
 
     - name: "OSX, 1.32.0"
       rust: 1.32.0
@@ -68,41 +73,40 @@ matrix:
         # This is actually useful as it finds stuff such as rust-random/rand#669
         - cargo web test --target wasm32-unknown-emscripten --no-run
         - cargo web test --target asmjs-unknown-emscripten --no-run
-        # wasi tests
         - cargo test --target wasm32-wasi
-        # stdweb tests (Node, Chrome)
-        - cd custom/stdweb
-        - cargo web test --nodejs --target=wasm32-unknown-unknown
-        - cargo web test --target=wasm32-unknown-unknown
-        # wasm-bindgen tests (Node, Firefox, Chrome)
-        - cd ../wasm-bindgen
-        - cargo test --target wasm32-unknown-unknown
-        - GECKODRIVER=$HOME/geckodriver cargo test --target wasm32-unknown-unknown --features=test-in-browser
-        - CHROMEDRIVER=$HOME/chromedriver cargo test --target wasm32-unknown-unknown --features=test-in-browser
+        # stdweb (wasm32-unknown-unknown) tests (Node, Chrome)
+        - cargo web test --package stdweb-getrandom
+        - cargo web test --package stdweb-getrandom --nodejs
+        # wasm-bindgen (wasm32-unknown-unknown) tests (Node, Firefox, Chrome)
+        - cargo test --package wasm-bindgen-getrandom
+          --target wasm32-unknown-unknown --test node
+        - GECKODRIVER=$HOME/geckodriver
+          cargo test --package wasm-bindgen-getrandom
+          --target wasm32-unknown-unknown --test web
+        - CHROMEDRIVER=$HOME/chromedriver
+          cargo test --package wasm-bindgen-getrandom
+          --target wasm32-unknown-unknown --test web
 
     - &nightly_and_docs
       name: "Linux, nightly, docs"
       rust: nightly
-      os: linux
       install:
         - rustup target add wasm32-unknown-unknown
         - cargo --list | egrep "^\s*deadlinks$" -q || cargo install cargo-deadlinks
         - cargo deadlinks -V
       script:
-        # Check that our tests pass in the default configuration
-        - cargo test
-        - cargo test --benches
         # Check that setting various features does not break the build
         - cargo build --features=std
         - cargo build --features=custom
         # remove cached documentation, otherwise files from previous PRs can get included
         - rm -rf target/doc
         - cargo doc --no-deps --features=std,custom
-        - cargo doc --no-deps --manifest-path=custom/wasm-bindgen/Cargo.toml --target=wasm32-unknown-unknown
+        - cargo doc --no-deps --package wasm-bindgen-getrandom --target=wasm32-unknown-unknown
         - cargo deadlinks --dir target/doc
-        # also test minimum dependency versions are usable
+        # Check that our tests pass in the default/minimal configuration
+        - cargo test --tests --benches
         - cargo generate-lockfile -Z minimal-versions
-        - cargo test --features=std,custom
+        - cargo test --tests --benches
 
     - <<: *nightly_and_docs
       name: "OSX, nightly, docs"
@@ -111,49 +115,23 @@ matrix:
     - name: "cross-platform build only"
       rust: nightly
       install:
-        - rustup target add x86_64-sun-solaris
-        - rustup target add x86_64-unknown-cloudabi
-        - rustup target add x86_64-unknown-freebsd
-        - rustup target add x86_64-fuchsia
-        - rustup target add x86_64-unknown-netbsd
-        - rustup target add x86_64-unknown-redox
-        - rustup target add x86_64-fortanix-unknown-sgx
+        - echo $STD_TARGETS | xargs -n1 rustup target add
         # For no_std targets
         - rustup component add rust-src
-        - cargo install cargo-xbuild || true
+        - cargo --list | egrep "^\s*xbuild$" -q || cargo install cargo-xbuild
       script:
-        - cargo build --target=x86_64-sun-solaris
-        - cargo build --target=x86_64-unknown-cloudabi
-        - cargo build --target=x86_64-unknown-freebsd
-        - cargo build --target=x86_64-fuchsia
-        - cargo build --target=x86_64-unknown-netbsd
-        - cargo build --target=x86_64-unknown-redox
-        - cargo build --target=x86_64-fortanix-unknown-sgx
-        - cargo xbuild --target=x86_64-unknown-uefi
-        - cargo xbuild --target=x86_64-unknown-hermit
-        - cargo xbuild --target=x86_64-unknown-l4re-uclibc
-        - cargo xbuild --target=x86_64-uwp-windows-gnu
-        - cargo xbuild --target=x86_64-wrs-vxworks
+        # We test that getrandom builds for all targets
+        - echo $STD_TARGETS | xargs -t -n1 cargo build --target
+        - echo $NO_STD_TARGETS | xargs -t -n1 cargo xbuild --target
         # also test minimum dependency versions are usable
         - cargo generate-lockfile -Z minimal-versions
-        - cargo build --target=x86_64-sun-solaris
-        - cargo build --target=x86_64-unknown-cloudabi
-        - cargo build --target=x86_64-unknown-freebsd
-        - cargo build --target=x86_64-fuchsia
-        - cargo build --target=x86_64-unknown-netbsd
-        - cargo build --target=x86_64-unknown-redox
-        - cargo build --target=x86_64-fortanix-unknown-sgx
-        - cargo xbuild --target=x86_64-unknown-uefi
-        - cargo xbuild --target=x86_64-unknown-hermit
-        - cargo xbuild --target=x86_64-unknown-l4re-uclibc
-        - cargo xbuild --target=x86_64-uwp-windows-gnu
-        - cargo xbuild --target=x86_64-wrs-vxworks
+        - echo $STD_TARGETS | xargs -t -n1 cargo build --target
+        - echo $NO_STD_TARGETS | xargs -t -n1 cargo xbuild --target
 
     # Trust cross-built/emulated targets. We must repeat all non-default values.
     - name: "Linux (MIPS, big-endian)"
       env: TARGET=mips-unknown-linux-gnu
       rust: stable
-      sudo: required
       dist: trusty
       services: docker
       install:
@@ -165,7 +143,6 @@ matrix:
     - name: "Android (ARMv7)"
       env: TARGET=armv7-linux-androideabi
       rust: stable
-      sudo: required
       dist: trusty
       services: docker
       install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -122,11 +122,11 @@ jobs:
       script:
         # We test that getrandom builds for all targets
         - echo $STD_TARGETS | xargs -t -n1 cargo build --target
-        - echo $NO_STD_TARGETS | xargs -t -n1 cargo xbuild --target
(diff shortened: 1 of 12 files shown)


custom/cpurand/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+
+#[macro_use]
+extern crate cfg_if;
+
+use getrandom::{register_custom_getrandom, Error};
+
+cfg_if! {
+    if #[cfg(target_arch = "x86_64")] {
+        #[path = "../../../src/util.rs"] mod util;
+        #[path = "../../../src/rdrand.rs"] mod imp;
```
dhardy: Do we want to use out-of-crate modules? We do have a reasonable option to resolve this in the future (copy `util`, move `rdrand`), so I guess it's acceptable, though probably any use of `#[path]` on modules is bad style.

Alternatively for `util` we could export `LazyBool` under some feature flag.

.travis.yml
```diff
@@ -122,20 +122,8 @@ matrix:
...
-        - cargo build --target=x86_64-unknown-netbsd
-        - cargo build --target=x86_64-unknown-redox
-        - cargo build --target=x86_64-fortanix-unknown-sgx
-        - cargo xbuild --target=x86_64-unknown-uefi
-        - cargo xbuild --target=x86_64-unknown-hermit
-        - cargo xbuild --target=x86_64-unknown-l4re-uclibc
-        - cargo xbuild --target=x86_64-uwp-windows-gnu
-        - cargo xbuild --target=x86_64-wrs-vxworks
-        # also test minimum dependency versions are usable
-        - cargo generate-lockfile -Z minimal-versions
```
dhardy: I think we should keep this.

.travis.yml
```diff
@@ -122,20 +122,8 @@ matrix:
...
+        # We test that getrandom and cpurand build for all x86 targets
+        - cd custom/cpurand
```
dhardy: I don't think there are any issues combining `--all` and `--target` options — so long as we're not using `--feature` I think it makes sense to test sub-crates via `--all`.

src/lib.rs
```diff
@@ -183,12 +181,7 @@ cfg_if! {
...
-    } else if #[cfg(all(target_arch = "x86_64", any(
-                  target_os = "hermit",
-                  target_os = "l4re",
-                  target_os = "uefi",
-                  target_env = "sgx",
-              )))] {
+    } else if #[cfg(all(target_arch = "x86_64", target_env = "sgx"))] {
         #[path = "rdrand.rs"] mod imp;
     } else if #[cfg(feature = "custom")] {
         use custom as imp;
```
newpavlov: In the OP you wrote the following:

>This crate also directly exposes a cpurand function for cases where a user want CPU-based RNG, even on platforms that have other options (like Linux)

But with the current configuration custom RNG will never be used for targets with a supported OS. To make it work as intended we may allow custom RNGs for all targets (i.e. move `if #[cfg(feature = "custom")]` to the very beginning), but it will have certain security implications, since users will be able to overwrite randomness source with a potentially insecure one even on well supported targets like Linux and Windows. I don't think it's a big deal and it even may be useful for niche use cases like deterministically testing code which uses `getrandom`, but it should be an explicit choice.

.travis.yml
josephlr: Done

.travis.yml
josephlr: You're right, but unfortunately we can't use `--all` with this crate. Our custom RNGs only compile on certain arches.

However, we can use `--package` to avoid changing the working directory. I did this where it makes sense.

src/lib.rs
josephlr: So I rephrased the OP to be more clear. The `getrandom-cpu` crate adds a custom `getrandom` handler that only runs on supported architectures. It also directly exposes a function `getrandom_cpu::cpurand` that allows a user of the crate to directly access the CPU randomness. 

This made it possible to test the RDRAND implementation on Linux. It also makes sure that supported targets cannot be overridden. This seems like the best of both worlds. Does this seem reasonable?

josephlr:
> Good rationale for using a separate `cpurand-getrandom` crate, though maybe it should be called `getrandom-cpu`?

@dhardy, good idea.  I switched the crates to be named `getrandom-*` instead of `*-getrandom`. This changes the name scheme we talked about in TODO, but that seems fine. Having `getrandom`, `getrandom-cpu`, `getrandom-stdweb`, etc... make it easier to see that these crates go together.

> The code looks reasonable. The tests — I think you removed min-dependency-version tests?

Fixed. I also cleaned up the Travis scripts in general, let me know what you think. 

> AFAIK ARM does not have a dedicated CPU instruction for retrieving random numbers and instead dedicated peripherals are used, so you can't rely on target triplet here. I also couldn't find PowerPC instruction (although it was a shallow search). I think it may be worth to introduce `getrandom-rdrand` crate instead and work on other targets in separate crates.

@newpavlov I updated the OP to document the instructions on ARM, Power, and s390.  At lot of these are very new, but were essentially introduced to be the counterparts of Intel's RDRAND. Particularly in the case of ARM and Power, these CPU instructions were introduced to avoid needing separate drivers for each board (i.e. the current state of affairs on ARM/Power).

Given these instructions, do you think it makes sense to have a `getrandom-cpu` and just add arch support as needed?

custom/cpurand/src/lib.rs
josephlr: I mean we do use `#[path]` a ton in the main `src/lib.rs`, but I get your point. 

I wasn't able to figure out how to structure the code to avoid the use of out-of-crate modules completely. Feature flags could solve this, and I even tried this initially, but it made the code much messier and harder to understand.

Do you have a way you would like this to be organized? The existing layout isn't _great_, but I couldn't figure out a better way.

src/lib.rs
newpavlov: I am not sure about usefulness of `getrandom_cpu::cpurand` in the proposed form. At the very least instead of independent function it could be better to define a struct which will implement `RngCore + CryptoRng`, so it will be possible to use it as part of the `rand` stack on "supported" platforms.

newpavlov:
@josephlr 
Thank you for the links! IIUC we will not be able to properly use it until rust-lang/rfcs#2725 gets accepted and implemented, correct?

So we have two use-cases (see my other [message](https://github.com/rust-random/getrandom/pull/133#discussion_r379915673)). Ideally we should separate "use a custom CPU-based `getrandom`" and "give me a CPU-based RNG without overwriting `getrandom`". The second use-case is already accomplished by the `rdrand` crate (modulo potential ARM/PowerPC support). We may introduce a competitor crate for `rdrand` with a wider range of supported platforms (let's call it `rand-cpu` for now), then I think `getrandom-cpu` should focus solely on the first use-case. To reduce duplication `getrandom-cpu` can use `rand-cpu` under the hood.

dhardy:
> "give me a CPU-based HWRNG without overwriting getrandom"

I'm not sure that we should care about this use-case? Is there a significant set of users? As you say, the `rdrand` crate already does provide this.

> It also directly exposes a function `getrandom_cpu::cpurand` ...

Exposing this directly for testing / niche purposes seems acceptable to me (even though it is of limited utility). It is a small API extension with no extra code, and I don't think there's any reason we'd need to make a breaking change here (without also breaking the main `getrandom` function).

custom/cpurand/src/lib.rs
dhardy: I guess it's not really important.

The plan is to remove `rdrand` from `getrandom` core, right? Or were you planning on keeping it for SGX support within `getrandom`? In the latter case, there seems to be less reason to make `getrandom-cpu` a separate crate instead of merely a feature flag.

The proposal in #84 appears to be somewhat different: using the `getrandom-cpu` crate directly, but renamed `getrandom`. This of course has the limitation that it only applies to the current user of `getrandom` (i.e. won't fix crates importing `rand` like usual). Perhaps best to take this back to #84...

newpavlov:
>I'm not sure that we should care about this use-case?

We may support wider range of targets with such crate, plus it would allow us to fix bug faster (e.g. `rdrand` still does not contain a workaround for the AMD bug). As for potential users, I am not sure.

>Exposing this directly for testing / niche purposes seems acceptable to me (even though it is of limited utility).

For testing it should be enough to use module-level tests, no need for exposing this function publicly. I am not sure why you brought breaking changes into this discussion. The approach, which I've drafted, will not introduce any breaking changes.

As for keeping RDRAND code for SGX targets, I think it may be worth to remove it from `getrandom` and tell users to use `getrandom-cpu` or other custom option instead. Since IIUC "the official" way for reading randomness on those targets is to use `sgx_read_rand`.

dhardy:
From the OP:

> We can now drop support for targets where we used RDRAND simply because we didn't know what else to use.

It may still be desirable to make these targets work by default. Note that libs should definitely not be importing a crate like `getrandom-cpu` since it precludes the use of a more specific implementation, such as a hypothetical `getrandom-sgx` crate. This means that there is no (good) way for a lib to include `rand` and have it "just work" on these targets.

custom/cpurand/src/lib.rs
josephlr: This is fixed. Now, the only use of out-of-crate modules are the custom RNG integration tests.

src/lib.rs
josephlr: I think you're correct here, we don't need to export another function. With the changes to use a `cpu` Cargo feature, exporting any `cpurand` function is no longer needed.

josephlr:
@newpavlov @dhardy I implemented my plan from https://github.com/rust-random/getrandom/issues/84#issuecomment-587180349, I rewrote the commit history, and the tests are now passing.

PTAL, each of the 5 commits is mostly self contained (be sure to keep the history when merging). 

The only remaining question is: What do we do for SGX targets? Right now, they are marked as explicitly supported, as RDRAND is how Intel implements [`sgx_read_rand`](https://github.com/intel/linux-sgx/blob/7b9f974482cf05a9ef1e133225b36b53be21d0e6/common/src/sgx_read_rand.cpp#L78-L101) (the supported way to get randomness on SGX w/ C/C++). We could just lump SGX in with the other x86_64 freestanding targets, but that doesn't seem quite right to me. Thoughts?

.travis.yml
```diff
@@ -68,41 +73,40 @@ matrix:
...
-        - cargo web test --nodejs --target=wasm32-unknown-unknown
-        - cargo web test --target=wasm32-unknown-unknown
-        # wasm-bindgen tests (Node, Firefox, Chrome)
-        - cd ../wasm-bindgen
-        - cargo test --target wasm32-unknown-unknown
-        - GECKODRIVER=$HOME/geckodriver cargo test --target wasm32-unknown-unknown --features=test-in-browser
-        - CHROMEDRIVER=$HOME/chromedriver cargo test --target wasm32-unknown-unknown --features=test-in-browser
+        # stdweb (wasm32-unknown-unknown) tests (Node, Chrome)
+        - cargo web test --package stdweb-getrandom
+        - cargo web test --package stdweb-getrandom --nodejs
```
dhardy: These two tests result in output:
```
test result (async): ok. 0 passed; 0 failed
```

.travis.yml
```diff
@@ -111,49 +115,23 @@ matrix:
...
-        - cargo build --target=x86_64-unknown-netbsd
-        - cargo build --target=x86_64-unknown-redox
-        - cargo build --target=x86_64-fortanix-unknown-sgx
-        - cargo xbuild --target=x86_64-unknown-uefi
-        - cargo xbuild --target=x86_64-unknown-hermit
-        - cargo xbuild --target=x86_64-unknown-l4re-uclibc
-        - cargo xbuild --target=x86_64-uwp-windows-gnu
-        - cargo xbuild --target=x86_64-wrs-vxworks
+        # We test that getrandom builds for all targets
+        - echo $STD_TARGETS | xargs -n1 cargo build --target
```
dhardy: Seems to work fine but it's not clear which target is being built. Can we echo the target name?

.travis.yml
josephlr: Correct, but this is the async output section. The main tests run after this (printing `All tests passed!`). This initially confused me, but making the tests intentionally fail (by changing `<` to `>` for example) does make this command fail.

.travis.yml
josephlr: I just added `-t` to the xargs invocation, it now prints the command being run.

.travis.yml
dhardy: Strange; usually the tests do get listed. Well, if you confirmed failure that's fine.

Cargo.toml
```diff
@@ -34,7 +34,10 @@ libc = { version = "0.2.64", default-features = false }
 wasi = "0.9"
 
 [features]
+default = ["cpu"]
```
newpavlov: Why enable it by default? De-facto it will be almost always enabled, since many crates will not use `default-featuers = false`, thus making the feature pointless. So I think we wither should make it a disabled by default feature, or do not feature-gate CPU-based implementation in the first place.

josephlr: I agree with this reasoning, and I think we should have the feature disabled by default. If a target is unsupported, we should not build it by default.

josephlr:
> > What do we do for SGX targets?
> 
> I guess we can leave it as-is, or if the `cpu` feature will be removed (see my comment below), then it can be lumped together with other x86-64 targets without issues.
> 
> UPD: Essentially we have to decide if our support of SGX via RDRAND is on the same level as for targets with OS or not, i.e. should we give users an ability to overwrite it or not. I am fine with both options.

If we have the CPU feature off by default, we should still have the SGX targets build by default, as the fortanix sgx taget is Tier 2 with full 'std' support. So I think having SGX be "fully suported" is fine.

Cargo.toml
josephlr: Fixed, the `cpu` feature is now off by default.

PR #135, opened 2020-02-26
Update docs for 0.2 Release 

josephlr:
Final PR for 0.2 adding all the docs about the `"rdrand"`, `"js"`, and `"custom"` features.

To preview what will go up on [crates.io](https://crates.io), run:
```bash
git clone --single-branch --branch docs https://github.com/josephlr/getrandom.git
cd getrandom
cargo doc --no-deps --features=std,custom --open`
```

### DIFF
--- a/src/custom.rs
+++ b/src/custom.rs
@@ -78,8 +78,9 @@ macro_rules! register_custom_getrandom {
         // We use an extern "C" function to get the guarantees of a stable ABI.
         #[no_mangle]
         extern "C" fn __getrandom_custom(dest: *mut u8, len: usize) -> u32 {
+            let f: fn(&mut [u8]) -> Result<(), ::getrandom::Error> = $path;
             let slice = unsafe { ::core::slice::from_raw_parts_mut(dest, len) };
-            match $path(slice) {
+            match f(slice) {
                 Ok(()) => 0,
                 Err(e) => e.code().get(),
             }
--- a/src/custom.rs
+++ b/src/custom.rs
@@ -25,12 +25,12 @@ use core::num::NonZeroU32;
 /// [dependencies]
 /// getrandom = { version = "0.2", features = ["custom"] }
 /// ```
-/// 
+///
 /// Next, in `dummy-getrandom/src/lib.rs`, we define our custom implementation and register it:
 /// ```rust
 /// use core::num::NonZeroU32;
 /// use getrandom::{Error, register_custom_getrandom};
-/// 
+///
 /// const MY_CUSTOM_ERROR_CODE: u32 = Error::CUSTOM_START + 42;
 /// fn always_fail(buf: &mut [u8]) -> Result<(), Error> {
 ///     let code = NonZeroU32::new(MY_CUSTOM_ERROR_CODE).unwrap();
@@ -43,10 +43,9 @@ use core::num::NonZeroU32;
 /// [`getrandom::getrandom`](crate::getrandom).
 ///
 /// Now any user of `getrandom` (direct or indirect) on this target will use the
-/// above custom implementation. Note that if you are using a helper-crate, some
-/// crate in the build needs to depend on `dummy-getrandom` via a
-/// `use dummy_getrandom;` statement. Failure to do this will result
-/// in linker errors.
+/// above custom implementation. See the
+/// [usage documentation](index.html#use-a-custom-implementation) for information about
+/// _using_ such a custom implementation.
 #[macro_export]
 macro_rules! register_custom_getrandom {
     ($path:path) => {
--- a/src/custom.rs
+++ b/src/custom.rs
@@ -12,40 +12,66 @@ use core::num::NonZeroU32;
 
 /// Register a function to be invoked by `getrandom` on unsupported targets.
 ///
-/// *This API requires the following Cargo features to be activated: `"custom"`*
+/// *This API requires the `"custom"` Cargo feature to be activated*.
 ///
-/// ## Writing your own custom `getrandom` implementation
+/// ## Writing a custom `getrandom` implementation
 ///
-/// Users can define custom implementations either in their root crate or in a
-/// target specific helper-crate. We will use the helper-crate approach in this
-/// example, defining `dummy-getrandom`, an implementation that always fails.
+/// The function to register must have the same signature as
+/// [`getrandom::getrandom`](crate::getrandom). The function can be defined
+/// wherever you want, either in root crate or a dependant crate.
 ///
-/// First, in `dummy-getrandom/Cargo.toml` we depend on `getrandom`:
+/// For example, if we wanted a `failure-getrandom` crate containing an
+/// implementation that always fails, we would first depend on `getrandom`
+/// (for the [`Error`] type) in `failure-getrandom/Cargo.toml`:
+/// ```toml
+/// [dependencies]
+/// getrandom = "0.2"
+/// ```
+/// Note that the crate containing this function does **not** need to enable the
+/// `"custom"` Cargo feature.
+///
+/// Next, in `failure-getrandom/src/lib.rs`, we define our function:
+/// ```rust
+/// use core::num::NonZeroU32;
+/// use getrandom::Error;
+///
+/// // Some application-specific error code
+/// const MY_CUSTOM_ERROR_CODE: u32 = Error::CUSTOM_START + 42;
+/// pub fn always_fail(buf: &mut [u8]) -> Result<(), Error> {
+///     let code = NonZeroU32::new(MY_CUSTOM_ERROR_CODE).unwrap();
+///     Err(Error::from(code))
+/// }
+/// ```
+///
+/// ## Registering a custom `getrandom` implementation
+///
+/// Functions can only be registered in the root binary crate. Attempting to
+/// register a function in a non-root crate will result in a linker error.
+/// This is similar to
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -27,75 +27,96 @@
...
+//! - Enable `getrandom`'s `"cpu"` feature
+//!   - This makes `getrandom` fallback to using CPU-specific RNG instructions.
+//!   - Currently only `x86`/`x86_64` targets are suppored. On these targets,
+//!     [RDRAND][18] will be used to generate the random data.
+//!   - If `getrandom` is not a direct dependancy of your crate, you can still
+//!     enable this fallback by adding the following to your crate's
+//!     `[dependancies]` section:
+//!     ```toml
+//!     getrandom = { version = "0.2", features = ["custom"]
+//!     ```
```
dhardy: This isn't really a list, is it? A sub-heading might work better.

src/lib.rs
```diff
@@ -27,75 +27,96 @@
...
-//! enabled, `wasm-bindgen` will be used. If neither feature is enabled, calls
-//! to `getrandom` will always fail at runtime.
-//!
-//! [bindgen]: https://github.com/rust-lang/rust-bindgen
+//! This crate supports the following `wasm32`/`asmjs` targets:
+//! - `wasm32-wasi`
+//! - `asmjs-unknown-emscripten`
+//! - `wasm32-unknown-emscripten`
+//!
+//! For `wasm32-unknown-unknown`, the story ismessy. This target can interface
```
dhardy: "ismessy"

src/lib.rs
```diff
@@ -27,75 +27,96 @@
...
+//! with JavaScript via either [wasm-bindgen] or [stdweb]. For this reason,
+//! **`wasm32-unknown-unknown` is an unsupported target**.
+//!
+//! However, as with any unsupported target, users can use custom `getrandom`
+//! implementation crates. For `wasm32-unknown-unknown`, there are:
+//! - [`wasm-bindgen-getrandom`][22], for use with [wasm-bindgen]
+//! - [`stdweb-getradnom`][23], for use with [stdweb]
```
dhardy: Better something like this IMO:

> The target `wasm32-unknown-unknown` is not directly supported since, from the target name alone, we cannot deduce which JavaScript interface is in use. Instead, the following crates provide support via the "custom `getrandom` implementation" mechanism (see above):
>
> - [`wasm-bindgen-getrandom`][22] via the [wasm-bindgen] JavaScript bridge
> ...

newpavlov:
@dhardy 
This PR is half-year old. We probably should commit your edits and move forward with 0.2 release. I would like to go back to rust-lang/rust#62079.

josephlr:
@newpavlov I 100% agree. I'll get this ready for final review ASAP. I also now have a strong work reason (https://github.com/briansmith/ring/issues/744#issuecomment-670804704) to make sure getrandom 0.2 is out asap

src/lib.rs
josephlr: Fixed

src/lib.rs
josephlr: Updated

josephlr:
@newpavlov @dhardy the docs are complete and are ready for review

src/custom.rs
```diff
@@ -10,13 +10,43 @@
...
+///
+/// register_custom_getrandom!(always_fail);
+/// ```
+/// the registered function must have the same type signature as
+/// [`getrandom::getrandom`](crate::getrandom).
+///
+/// Now any user of `getrandom` (direct or indirect) on this target will use the
+/// above custom implementation. Note that if you are using a helper-crate, some
+/// crate in the build needs to depend on `dummy-getrandom` via a
+/// `use dummy_getrandom;` statement. Failure to do this will result
```
newpavlov, line 48: I think it's worth to add recommendation that such helper crate should not be used in library crates as a non-dev dependency.

src/stdweb.rs
```diff
@@ -10,7 +10,8 @@ use crate::Error;
 extern crate std;
 use std::thread_local;
 
-use stdweb::js;
+#[macro_use]
+extern crate stdweb;
```
newpavlov, line 14: Is there a reasons for this change?

newpavlov (approved):

src/stdweb.rs
josephlr: Whoops, nope. It was a rebase error. Removed

src/custom.rs
josephlr: Oh good point, I'll add that.

josephlr:
@stevenrutherford you've used this branch before, can you take a look at the docs.

src/custom.rs
josephlr: Added

josephlr:
@newpavlov one last question (then we can merge this).

Right now the documented way to use a custom-RNG crate is:
  1. In the custom RNG crate (i.e. `dummy-getrandom`)
     - Depend on `getrandom` in the `Cargo.toml` with the `"custom"` feature enabled
     - Define a function (i.e. `always_fail`) with signature `fn() -> Result<(), getrandom::Error>`
     - Call `register_custom_getrandom!(always_fail);`
  2. In the binary crate
     - Depend on `dummy-getrandom` in the `Cargo.toml`.
     - Use the crate in `src/main.rs` by having `use dummy_getrandom;` at the top of the file.
        - Failure to do this step creates linker errors.
     - Binary crate doesn't need to depend on `getrandom`.

An **alternative** way to do things would be:
  1. In the custom RNG crate (i.e. `dummy-getrandom`)
     - Depend on `getrandom` in the `Cargo.toml` (no need to select any features)
     - Define a **public** function (i.e. `always_fail`) with signature `fn() -> Result<(), getrandom::Error>`
  2. In the binary crate
     - Depend on `dummy-getrandom` in the `Cargo.toml`.
     - Depend on `getrandom` in the `Cargo.toml` with the `"custom"` feature enable
     - Have the following code in `src/main.rs`:
       ```rust
       use getrandom::register_custom_getrandom;
       use dummy_getrandom::always_fail;
       register_custom_getrandom!(always_fail);
       ```

Do you think we should recommend the current approach or the alternative approach? I think the alternative approch better mimics common patterns with `#[panic_handler]` and `#[global_allocator]`.

newpavlov:
Hm, I agree that the second approach is closer to the global allocator and panic handlers, plus it feels  "less magic". We could reduce amount of setup a bit, by suggesting that `dummy-getrandom` could depend on `getrandom` with the enabled `custom` feature and re-export it, so users could write:
```rust
use dummy_getrandom::{always_fail, getrandom::register_custom_getrandom};
register_custom_getrandom!(always_fail);
```
But I am not sure if it's worth the trouble.

josephlr:
> Hm, I agree that the second approach is closer to the global allocator and panic handlers, plus it feels "less magic".

That was my thinking as well, docs have been updated to show the second approach. I also added a check in the macro that the function has the exact correct signature. This makes the error messages much better if you feed it a bad function.

> But I am not sure if it's worth the trouble.

Ya I agree. I think keeping it sort of close to how `#[global_allocator]` works is for the best.

dhardy:
Moving the binding to the binary crate does avoid possible confusion about how custom RNG crates work, and about what happens if multiple custom RNG crates were enabled. On the other hand it's a bit strange and inconvenient putting the binding in a binary crate if the binary crate doesn't use `getrandom` itself (e.g. if just using `rand`).

josephlr:
> On the other hand it's a bit strange and inconvenient putting the binding in a binary crate if the binary crate doesn't use `getrandom` itself (e.g. if just using `rand`).

I agree it's a little inconvenient, but it does make things less magical. It's also consistent with the "js" and "rdrand" features, which are also enabled by directly depending on getrandom (even if the user was only using rand).

Fundamentally, if a user is doing this, they need to know they are overriding getrandom, otherwise they'll have issues/conflicts down the road.

I also like this approach because it's compatible with the eventual introduction of a `#[random_handler]` lang item as part of core/alloc/std.

src/lib.rs
```diff
@@ -6,14 +6,15 @@
...
 
-//! Interface to the random number generator of the operating system.
+//! Interface to the operating system's random number generator.
 //!
-//! # Platform sources
+//! # Supported targets
 //!
-//! | OS               | interface
+//! | Target           | Implementation
 //! |------------------|---------------------------------------------------------
```
stevenrutherford, line 14: These aren't targets, but they are almost targets. Not sure what you should call it, but I think target isn't correct. You could change the column to be the taget triples. You could call it environment instead. 

This is a bit of a bikeshed and anyone reading this can probably read between the lines.

josephlr: We could make this table have three columns: Target, Target Triple, Implementation

would look like:

| Target | Target Triple | Implementation
| --- | ----- | --- | 
| Linux, Android | `*-linux-*` | [`getrandom`](1) system call if available, otherwise [`/dev/urandom`](2) after successfully polling `/dev/random` |
| Windows | `*-pc-windows-*` | [`RtlGenRandom`](3) |
| [Windows UWP](22) | `*-uwp-windows-*` | [`BCryptGenRandom`](23) | 

I'll need to see if this actually ends up looking good though

josephlr:
I added target names to the table and it actually looks pretty good:

table


This should also help people who are confused about macos vs Darwin, or what windows UWP is.

PR #138, opened 2020-04-16
Avoid require an expression (wasm-bindgen)

IcanDivideBy0:
When using getrandom with Webpack, there is a warning (which is a bit annoying) about using an expression in a `require` call
```
Critical dependency: the request of a dependency is an expression
```

It turns out this can be avoided by using the [`module` attribute](https://rustwasm.github.io/docs/wasm-bindgen/reference/attributes/on-js-imports/module.html) on the wasm_bindgen macro. Also the resulting code is a bit simpler and more correct (`randomFillSync` is not a method of some crypto class, just a function attached to the `module.exports` object, unlike the browser version, where the `getRandomValues` function is declared in the prototype of the crypto object)

### DIFF
--- a/custom/wasm-bindgen/src/lib.rs
+++ b/custom/wasm-bindgen/src/lib.rs
@@ -19,7 +19,7 @@ use getrandom::{register_custom_getrandom, Error};
 
 #[derive(Clone, Debug)]
 enum RngSource {
-    Node(NodeCrypto),
+    Node(),
     Browser(BrowserCrypto),
 }
 
@@ -41,7 +41,7 @@ fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
         }
 
         match source.as_ref().unwrap() {
-            RngSource::Node(n) => n.random_fill_sync(dest),
+            RngSource::Node() => random_fill_sync(dest),
             RngSource::Browser(n) => {
                 // see https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
                 //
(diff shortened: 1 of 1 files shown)


IcanDivideBy0:
Travis build failure seems unrelated:
```
$ cargo web test --target wasm32-unknown-emscripten --no-run
   Compiling libc v0.2.69
   Compiling getrandom v0.1.14 (/home/travis/build/rust-random/getrandom)
   Compiling cfg-if v0.1.10
    Finished test [unoptimized + debuginfo] target(s) in 1.35s
   Compiling getrandom v0.1.14 (/home/travis/build/rust-random/getrandom)
    Finished test [unoptimized + debuginfo] target(s) in 1m 01s
thread 'main' panicked at 'internal error: wasm doesn't exist where I expected it to be', src/cargo_shim/mod.rs:905:29
```

josephlr:
I prefer this approach (if it works) to the one in #137 

Once I fix the CI I'll try it out. As mentioned in that issue, this fix should be submitted against the 0.2 branch.

IcanDivideBy0:
Freshly rebased on 0.2 branch :)

IcanDivideBy0:
Freshly rebased on 0.2 branch :)

josephlr:
So I tried this out locally w/ Firefox, and unfortunately this isn't going to work. When running on Firefox, we get `TypeError: Error resolving module specifier: crypto`. So it looks like the NodeJS module is being looked up unconditionally (even when not running under NodeJS).

For context, this code should work in the browser and in Node by doing runtime detection. wasm-bindgen support is already in its own crate, so we could split it into `wasm-bindgen-web` and `wasm-bindgen-node`, but that seems unnecessary given that it's _possible_ to check the implementation at runtime.

IcanDivideBy0:
Ok feel free to close this then!

PR #139, opened 2020-04-24
wasm-bindgen: Added support for Internet Explorer 11

zer0x64:
It is a pretty obscure use case, but I did stumble upon it:

When using wasm2js to compile WASM to Javascript for use in older browsers [as recommended by the wasm-bindgen documentation](https://rustwasm.github.io/docs/wasm-bindgen/examples/wasm2js.html), we found out that `rand` crate did not work on Internet Explorer 11 since the `window.crypto` object is named `window.msCrypto`.

This PR adds a check to use `msCrypto` if `crypto` is not present.

### DIFF
--- a/src/wasm32_bindgen.rs
+++ b/src/wasm32_bindgen.rs
@@ -60,17 +60,17 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
 fn getrandom_init() -> Result<RngSource, Error> {
     if let Ok(self_) = Global::get_self() {
         // If `self` is defined then we're in a browser somehow (main window
-        // or web worker). Here we want to try to use
-        // `crypto.getRandomValues`, but if `crypto` isn't defined we assume
-        // we're in an older web browser and the OS RNG isn't available.
+        // or web worker). We get `self.crypto` (called `msCrypto` on IE), so we
+        // can call `crypto.getRandomValues`. If `crypto` isn't defined, we
+        // assume we're in an older web browser and the OS RNG isn't available.
 
-        let crypto = self_.crypto();
-        if crypto.is_undefined() {
-            return Err(BINDGEN_CRYPTO_UNDEF);
-        }
+        let crypto: BrowserCrypto = match (self_.crypto(), self_.ms_crypto()) {
+            (crypto, _) if !crypto.is_undefined() => crypto.into(),
+            (_, crypto) if !crypto.is_undefined() => crypto.into(),
+            _ => return Err(BINDGEN_CRYPTO_UNDEF),
+        };
 
         // Test if `crypto.getRandomValues` is undefined as well
-        let crypto: BrowserCrypto = crypto.into();
         if crypto.get_random_values_fn().is_undefined() {
             return Err(BINDGEN_GRV_UNDEF);
         }
@@ -88,6 +88,8 @@ extern "C" {
     fn get_self() -> Result<Self_, JsValue>;
 
     type Self_;
+    #[wasm_bindgen(method, getter, js_name = "msCrypto", structural)]
+    fn ms_crypto(me: &Self_) -> JsValue;
     #[wasm_bindgen(method, getter, structural)]
     fn crypto(me: &Self_) -> JsValue;
 


src/wasm32_bindgen.rs
```diff
@@ -88,6 +94,11 @@ extern "C" {
     fn get_self() -> Result<Self_, JsValue>;
 
     type Self_;
+
+    // Internet Explorer 11
+    #[wasm_bindgen(method, getter, js_name="msCrypto", structural)]
+    fn ms_crypto(me: &Self_) -> JsValue;
+
```
josephlr, line 101: Formatting nit here:

```rust
    type Self_;
    #[wasm_bindgen(method, getter, structural)]
    fn crypto(me: &Self_) -> JsValue;
    #[wasm_bindgen(method, getter, js_name = "msCrypto", structural)]
    fn ms_crypto(me: &Self_) -> JsValue;
```

We should explain the use of `msCrypto` above (when we use it)

src/wasm32_bindgen.rs
```diff
@@ -64,9 +64,15 @@ fn getrandom_init() -> Result<RngSource, Error> {
         // `crypto.getRandomValues`, but if `crypto` isn't defined we assume
         // we're in an older web browser and the OS RNG isn't available.
```
josephlr, line 65: We should document the crypto vs mscrypto distinction here, so it's all in one place. Something like
```rust
        // If `self` is defined then we're in a browser somehow (main window
        // or web worker). We get `self.crypto` (called `msCrypto` on IE), so we
        // can call `crypto.getRandomValues`. If `crypto` isn't defined, we
        // assume we're in an older web browser and the OS RNG isn't available.
```

src/wasm32_bindgen.rs
```diff
@@ -64,9 +64,15 @@ fn getrandom_init() -> Result<RngSource, Error> {
...
+        let mut crypto = self_.crypto();
         if crypto.is_undefined() {
-            return Err(BINDGEN_CRYPTO_UNDEF);
+            // If crypto is undefined, we can also check if msCrypto is defined, 
+            // in case it's Internet Explorer
+            crypto = self_.ms_crypto();
+            
+            if crypto.is_undefined() {
+                return Err(BINDGEN_CRYPTO_UNDEF);
+            }
```
josephlr, line 75: We can make this logic more straightforward using Rust's match guards:
```rust
        let crypto: BrowserCrypto = match (self_.crypto(), self_.ms_crypto()) {
            (crypto, _) if !crypto.is_undefined() => crypto.into(),
            (_, crypto) if !crypto.is_undefined() => crypto.into(),
            _ => return Err(BINDGEN_CRYPTO_UNDEF),
        };

        // Test if `crypto.getRandomValues` is undefined as well
        if crypto.get_random_values_fn().is_undefined() {
            return Err(BINDGEN_GRV_UNDEF);
        }
```

josephlr:
@zer0x64, also be sure to run `cargo fmt --all -- **/*.rs` before submitting. By default, `cargo fmt` won't format the target-specific files.

src/wasm32_bindgen.rs
zer0x64: I really like that!

zer0x64:
Great, I'll address the changes tomorrow! 

For the CI, unfortunately, wasm-bindgen-test doesn't support headless IE, so the only way I see to test this in a CI is to use a framework like Selenium. It is certainly doable, but not as straightforward as testing Firefox/Chrome. Also, I don't think this can be done without a Windows build server. However, I can help integrate it if you decide to add this check to the CI!

josephlr:
@zer0x64 would [`wasm-bindgen`'s `vendor_prefix`](https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-js-imports/vendor_prefix.html) do all of this for us?

Could we just do:
```
    type Self_;
    #[wasm_bindgen(method, getter, vendor_prefix = "ms", structural)]
    fn crypto(me: &Self_) -> JsValue;
```
and not need any additional code? I don't know if `msCrypto` and `mscrypto` are the same thing in JS.

> EDIT: Also forgot to mention that since IE does not support webassembly, we would need to use `wasm2js` to translate it to javascript and, from experience, that part is mostly manual work and does not integrates well in a CI for now. According to the `wasm-bindgen` doc, some day it should be supported out of the box without manual work.

This seems reasonable. We can always add more testing later once the tools are in a better state. Our coverage right now is fairly good (Node, Chrome, Firefox).

zer0x64:
I did some tests with the vendor_prefix and it simply doesn't seem to work in this context at all. I think it might be since the annotation is made for `type` and not for `fn`. Although even if it worked, we'd still have the problem that it exports `msCrypto` and not `mscrypto`.

zer0x64:
I've just pushed the fixes for your comments!

josephlr:
@zer0x64 can you rebase this onto the `0.2` branch (right now this PR is against the `master` branch), this will allow you to pickup #141 

The only change you'll need will be to turn `BINDGEN_CRYPTO_UNDEF` into `Error::BINDGEN_CRYPTO_UNDEF`

josephlr:
Actually, @zer0x64, having this against the `master` branch is fine. I'll handle putting the changes on the `0.2` branch. Just rebase this on #142 when that lands.

josephlr:
@zer0x64 #142 has been merged, can you rebase this onto master?

src/wasm32_bindgen.rs
```diff
@@ -60,14 +60,15 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
...
+        // assume we're in an older web browser and the OS RNG isn't available.
+
+        let crypto: BrowserCrypto = match (self_.crypto(), self_.ms_crypto()) {
+            (crypto, _) if !crypto.is_undefined() => crypto.into(),
+            (_, crypto) if !crypto.is_undefined() => crypto.into(),
+            _ => return Err(BINDGEN_CRYPTO_UNDEF),
+        };
 
         // Test if `crypto.getRandomValues` is undefined as well
         let crypto: BrowserCrypto = crypto.into();
```
josephlr, line 74: We no longer need this block (as we already convert the `JsValue` to a `BrowserCrypto` above.

zer0x64: No problem!

zer0x64:
Rebased and tested on my side, so when the CI pass it will be ready to merge!

PR #141, opened 2020-04-25
travis: Fix wasm32 CI

### LINKED ISSUE #140: wasm CI is broken
If we want to merge wasm related PRs (#137, #138, #139), we need to fix the CI. I'll look into this today.

CC: @Pauan, @zer0x64, @IcanDivideBy0

josephlr:
Split emscripten stuff into its own target and stop relying on `cargo web`
to download/manage then emscripten toolchain. We can just get it
ourselves.

We now use `stable` for all WASM32 tests and are now running the emscripten tests.

Fixes #140 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,15 +33,13 @@ jobs:
     - name: "Linux, beta"
       rust: beta
 
-    - name: "WASM via emscripten, stdweb, wasm-bindgen and WASI"
-      rust: nightly
+    - name: "WASM via stdweb, wasm-bindgen and WASI"
+      rust: stable
       addons:
         firefox: latest
         chrome: stable
       install:
         - rustup target add wasm32-unknown-unknown
-        - rustup target add wasm32-unknown-emscripten
-        - rustup target add asmjs-unknown-emscripten
         - rustup target add wasm32-wasi
         # Get latest geckodriver
         - export VERSION=$(curl -s https://api.github.com/repos/mozilla/geckodriver/releases/latest | jq -r ".tag_name")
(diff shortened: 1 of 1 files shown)


josephlr:
@newpavlov @dhardy dropping the use of `cargo web` for emscripten seems to have fixed the CI.

Other improvements:
  - We now actually run the tests (instead of just building them)
  - The CI is faster (one 6-7 minute stage became two 2-2.5 minute stages)

dhardy:
I don't think I can provide any useful feedback here. If someone else (e.g. @zer0x64) can provide feedback then I'll be happy to "approve" (if required).

josephlr:
> I don't think I can provide any useful feedback here. If someone else (e.g. @zer0x64) can provide feedback then I'll be happy to "approve" (if required).

@dhardy, I can give some context for the exact steps I added to the CI. I replaced `cargo web prepare-emscripten` with [the steps from the emscripten download page](https://emscripten.org/docs/getting_started/downloads.html). Getting the tests to run was then just a matter of [copying what `cross` does](https://github.com/rust-embedded/cross/blob/master/docker/Dockerfile.asmjs-unknown-emscripten).

Either you or @newpavlov have to approve (I can't approve my own requests).

zer0x64:
If I understand correctly, `cargo-web` is currently broken with `wasm32-unknown-emscripten` as per [this](https://github.com/koute/cargo-web/pull/244), and while the fixed is merge and released, this change aims compile and run directly this specific build without using `cargo web`? 
Also, the CI step is split into two, which is cool because they can now run concurrently and fail separately.
Makes sense to me!

PR #146, opened 2020-05-22
wasm32: Unify error codes and implementations

josephlr:
Right now, while our implementations for stdweb and wasm-bindgen use
similar APIs, they do not use similar implementations. This change:

  - Switches to using the same error codes for both implementations
  - Uses error codes that detail exactly what API is missing
  - Cleans up the implemetnations to make them much more readable
  - Consitantly use thread_local in both implementations

This makes issues easier to debug and the code easier to understand.


josephlr:
@dhardy I fixed the CI issue, this is ready for review.

dhardy:
Sorry, I'd rather not review this myself. Could I ask @zer0x64 or @Pauan to give this a review?

PR #153, opened 2020-07-19
Remove workaround for older Linux kernels

newpavlov:
Minimum supported kernel version got bumped to 2.6.32 in rust-lang/rust#74163

### DIFF
--- a/src/util_libc.rs
+++ b/src/util_libc.rs
@@ -111,9 +111,5 @@ pub unsafe fn open_readonly(path: &str) -> Result<libc::c_int, Error> {
     if fd < 0 {
         return Err(last_os_error());
     }
-    // O_CLOEXEC works on all Unix targets except for older Linux kernels (pre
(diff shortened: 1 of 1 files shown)


newpavlov:
CI failures look unrelated to changes in this PR.

dhardy:
Presumably that change will appear in Rust 1.46.0, which means that in theory we should not remove this workaround until the MSRV ≥ 1.46.

On the other hand, I doubt many people actually care about using the latest `getrandom` on such an ancient system, so it's likely unimportant. @josephlr?

newpavlov:
>Presumably that change will appear in Rust 1.46.0

I don't think so. Libraries requiring stricter OS requirements than `std` is not unusual, e.g. crates which utilize `io-uring`. We had two reasons for the workaround: possibility of `getrandom` being used by `std` and the fact that it's part of Firefox dependency tree. Since RHEL and CentOS no longer block the update, I think we can go forward with this change. And we are being extra careful by releasing it as part of v0.2. and not v0.1.

PR #154, opened 2020-07-21
stdweb: Remove CI runs with cargo-web

josephlr:
Right now "cargo web test" is broken due to upstream changes. We will
disable stdweb tests until the stdweb project fixes things.

See: https://github.com/koute/cargo-web/issues/243

Fixes CI issues noticed in #153

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -110,7 +110,6 @@ jobs:
         # remove cached documentation, otherwise files from previous PRs can get included
         - rm -rf target/doc
         - cargo doc --no-deps --features=std,custom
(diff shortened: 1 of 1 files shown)


josephlr:
CI is looking better, but there are still some CI flakes:
  - Adding the `rust-src` component is currently broken: https://github.com/rust-lang/rust/pull/74578
  - We are running into Github API limiting for certain CI operations ("API rate limit exceeded for <IP Addr>")

The solution to both these issues is to ignore them in the short term, and move away from Travis (to Github Actions) in the long term.

PR #156, opened 2020-08-14
bsd_arandom: Read only 256 bytes at a time

### LINKED ISSUE #151: bsd_arandom.rs should loop on 256-byte chunks
The assumption that it does a short read for long buffers isn't always true. After reviewing the code in the NetBSD kernel we decided to adopt this behaviour moving forward (because it seems safer after reviewing real-world code using it), but older NetBSD releases will simply return E2BIG if more than 256 bytes are requested.

https://github.com/NetBSD/src/commit/7161bc211696634c0d0bc55c41a341bf919206e0

The semantics should be the same as getentropy(), which does loop on 256 byte chunks. The only difference is that the output size should be checked (however, the NetBSD kernel at least will never return short reads when <=256 bytes are requested, and will otherwise always return 256 bytes).

josephlr:
Fixes #151 CC: @newpavlov @niacat 

Older NetBSD kernels cannot handle buffers bigger than 256 bytes, and
all FreeBSD and NetBSD kernels only return up to 256 bytes per call.

### DIFF
--- a/src/bsd_arandom.rs
+++ b/src/bsd_arandom.rs
@@ -43,5 +43,10 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
             return sys_fill_exact(dest, |buf| unsafe { func(buf.as_mut_ptr(), buf.len(), 0) });
         }
     }
-    sys_fill_exact(dest, kern_arnd)
+    // Both FreeBSD and NetBSD will only return up to 256 bytes at a time, and
+    // older NetBSD kernels will fail on longer buffers.
+    for chunk in dest.chunks_mut(256) {
+        sys_fill_exact(chunk, kern_arnd)?
+    }
+    Ok(())
 }


newpavlov:
I wonder if we can assume that both FreeBSD and NetBSD always completely fill buffers smaller than 256 byte. It would allow us to remove `sys_fill_exact`, thus removing one loop.

josephlr:
> I wonder if we can assume that both FreeBSD and NetBSD always completely fill buffers smaller than 256 byte. It would allow us to remove `sys_fill_exact`, thus removing one loop.

Looking at the code this seems true (at least with current versions), but unlike other interfaces that _do_ guarantee complete filling, the docs for both NetBSD and FreeBSD don't mention this anywhere, so I wouldn't want to depend on it. Having the extra loop is weird, but should have a negliglbe hurt perf or size impact (as the branch is always taken). Regardless we would still need to check for EINTR (if I'm understanding the API completely). 

> @dhardy
> If the above assumption does not hold in practice, shouldn't we replace `len as libc::ssize_t` with `ret` in the `kern_arnd` function?

Doing that would be incorrect as [the `ret` parameter of `sysctl` is 0 on success](https://www.freebsd.org/cgi/man.cgi?query=sysctl), so that wouldn't work with the expected sematics of `sys_fill_exact` which expects a length as the return value.

newpavlov:
>Doing that would be incorrect as the ret parameter of sysctl is 0 on success, so that wouldn't work with the expected sematics of sys_fill_exact which expects a length as the return value.

Doesn't "success" (i.e. non-negative `ret`) in that case mean that buffer is completely filled? IIRC for some other OS it was explicitly stated that `EINTR` can be only be returned for big buffers. It's surprising that FreeBSD and NetBSD documentation is not sufficiently clear in this regard. I guess to be extra safe we should go with the current code, though I think it smells a bit.

josephlr:
> Doesn't "success" (i.e. non-negative `ret`) in that case mean that buffer is completely filled? IIRC for some other OS it was explicitly stated that `EINTR` can be only be returned for big buffers. It's surprising that FreeBSD and NetBSD documentation is not sufficiently clear in this regard. I guess to be extra safe we should go with the current code, though I think it smells a bit.

You're correct, but `sys_fill_exact` expects the passed in function to return the number of bytes written. Agreed on everything else (I think Linux garuntees no EINTR for <= 256 bytes).

niacat:
Thanks!

PR #157, opened 2020-09-08
travis: Update pinned dependancies

josephlr:
Also, pin the version of rustc, so that nightly updates don't break
our build. This means that we can add the nightlies back into the
blocking set of tests.

Move cloudabi to `cargo xbuild` as it is now a Teir 3 target

Disable Firefox testing due to https://github.com/rustwasm/wasm-bindgen/issues/2261

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,7 +91,7 @@ matrix:
         - RUSTFLAGS='-C debuginfo=0' cargo test --target asmjs-unknown-emscripten
 
     - name: "Linux, nightly, docs"
-      rust: nightly-2020-09-08
+      rust: nightly
       os: linux
       install:
         - cargo --list | egrep "^\s*deadlinks$" -q || cargo install cargo-deadlinks
@@ -109,7 +109,7 @@ matrix:
         - cargo test
 
     - name: "OSX, nightly, docs"
-      rust: nightly-2020-09-08
+      rust: nightly
       os: osx
       install:
         - cargo --list | egrep "^\s*deadlinks$" -q || cargo install cargo-deadlinks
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -88,7 +91,7 @@ matrix:
         - RUSTFLAGS='-C debuginfo=0' cargo test --target asmjs-unknown-emscripten
 
     - name: "Linux, nightly, docs"
-      rust: nightly
+      rust: nightly-2020-09-08
```
newpavlov, line 94: I am not sure if it's a good idea to pin Nightly. Even if in future some change will break our build, I think it's better to find about it sooner than later.

josephlr: Would you want to mark all of the nightly builds as non-blocking then? We use nightly for:
  - a Linux build
  - a Mac build
  - cross compiling
  - All of our WASM targets

josephlr: I generally agree with this point though. I think we should move as much as we can out of the nightly tests into the stable tests. Then, mark all the stable tests as non-blocking.

newpavlov: Did we have problems with nightly builds previously? If I had to chose, I would prefer making those jobs non-blocking.

josephlr: We did, a nightly didn't have some component we needed (rustfmt maybe, but that's already non-blocking).

I agree that we should have them be blocking, should be fixed now.

newpavlov: BTW do we have cron jobs enabled in Travis?

cc @dhardy

newpavlov (approved):
Looks good to me, but note that I do not have a deep understanding of the WASM part of CI job.

.travis.yml
josephlr: Looking at https://travis-ci.org/github/rust-random/getrandom/builds/714877349 and https://travis-ci.org/github/rust-random/getrandom/builds/724230064 they are monthly cron jobs

PR #158, opened 2020-09-08
travis: Apply #157 to 0.2 branch

josephlr:
Note that Firefox testing had to be disabled due to https://github.com/rustwasm/wasm-bindgen/issues/2261, we need them to cut another release before we can start testing on Firefox again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,14 @@
 language: rust
 os: linux
+dist: focal
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


josephlr:
Given that this just changes CI and we can continue discussion in #157 (for the master branch), I'm going to merge this so the remaing `getrandom v0.2` PRs can be green (and easier to review.

PR #159, opened 2020-09-08
Update MSRV to 1.34

josephlr:
Per the discussion in #98, we can update the MSRV as far as 1.36 if we want (as that's `rand 0.8`'s MSRV). However, Debian Buster (aka Debian 10, aka Debian Stable) is only at 1.34. As we don't need any features released between 1.34 and 1.36, only bumping the MSRV to 1.34 seems fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,11 @@ env:
 
 jobs:
   include:
-    - name: "Linux, 1.33.0"
-      rust: 1.33.0
+    - name: "Linux, 1.34.0"
+      rust: 1.34.0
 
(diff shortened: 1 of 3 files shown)


newpavlov (approved):
I think we can keep the current MSRV, but I guess it could be worth to reserve an ability to use `TryInto`/`TryFrom` just in case.

josephlr:
> I think we can keep the current MSRV, but I guess it could be worth to reserve an ability to use `TryInto`/`TryFrom` just in case.

Ya, my thought was we might eventually want `TryFrom` for something with our Error types/codes.

PR #162, opened 2020-09-10
Release Version 0.2

### LINKED ISSUE #98: Tracking issue for 0.2.0

josephlr:
Fixes #98 

It's finally done! I can't believe this started almost a year ago.

To preview docs:
```bash
git clone --single-branch --branch 0.2 https://github.com/rust-random/getrandom.git
cd getrandom
cargo doc --no-deps --features=std,custom --open`
```

## Release plan
After merging, this version will be released as `getrandom 0.2`

The last pre-0.2 commit (https://github.com/rust-random/getrandom/commit/7dcb1676948fe3e8e5c7a0e5db9f2d1d609a01d4) will be released as `getrandom 0.1.15`


dhardy:
:tada: 

Maybe there should be a `0.1` branch now?

josephlr:
> Maybe there should be a `0.1` branch now?

That's the plan, I'll add it

josephlr:
> Maybe also add a changelog entry?

Done

CHANGELOG.md
```diff
@@ -4,6 +4,32 @@ All notable changes to this project will be documented in this file.
...
+### Features for using getrandom on unsupported targets
+
+The following (off by default) Cargo features have been added:
+- `"rdrand"` - use the RDRAND instruction on `no_std` `x86`/`x86_64` targets [#133]
+- `"js"` - use JavaScript calls on `wasm32-unknown-unknown` [#149]
+  - Replaces the `stdweb` and `wasm-bindgen` features (which are removed)
+- `"custom"` - allows a user to specify a custom implementation [#109]
+
+### Breaking Changes
+- Chanage/Add `Error` constants [#120]
```
newpavlov, line 17:
```suggestion
- Change/Add `Error` constants [#120]
```

josephlr: Fixed, I also added the 0.1.15 changelog in the 0.2 changelog (as those changes were applied to 0.2)

Issue #164, opened 2020-10-04
Exception on multithreaded wasm


chemicstry:
I'm building a project (`bevy`) on multithreaded wasm via web workers and this library crashes with exception:

`Uncaught (in promise) TypeError: Failed to execute 'getRandomValues' on 'Crypto': The provided ArrayBufferView value must not be shared.`

Indeed the spec does not allow generating random numbers into a memory backed by `SharedArrayBuffer`.

I managed to fix this issue by generating numbers into an intermediate buffer allocated by javascript:
```
for chunk in dest.chunks_mut(65536) {
    let arr = js_sys::Uint8Array::new_with_length(chunk.len() as u32);
    n.get_random_values(&arr);
    arr.copy_to(chunk);
}
```

This introduces some overhead, but I'm not sure if there is a standard way to check whether rust's memory is backed by `SharedArrayBuffer`.

I can submit a PR for this, so let me know how you want me to proceed.

newpavlov:
Some discussion on the matter: tc39/proposal-csprng#6

Allocating a buffer for each chunk via JS looks like a bit too much overhead to make this workaround a default option. But IIUC allocating a temporary buffer on the Rust side will not work either, since compiler in full right to eliminate "needless" copies.

My first reaction is to suggest not using a shared array like this and instead perform copy of generated random values in a higher level code, but I don't have a strong opinion on this matter. If it will be decided to leave code as-is, at the very least we should document this behavior.

gregkatz:
@chemicstry, would you mind sharing your fork? I'm not sure where the code you've shared above should go, and I'm hitting this as well. I don't care about the performance overhead for my personal project.

josephlr:
@chemicstry I would like some more info on how this is happening. Looking at the [`js_sys::SharedArrayBuffer` docs](https://docs.rs/js-sys/0.3.45/js_sys/struct.SharedArrayBuffer.html) there doesn't seem a way to turn that object into a Rust `&mut [u8]`. How is this being done?

Perhaps there just shouldn't be a way to turn  a `SharedArrayBuffer` into a `&mut [u8]` without copying. Our call to `getRandomValues` isn't going to be the only API that breaks, so maybe this should be fixed at a higher level. This seems reasonable, given that there are also a ton of security issues around `SharedArrayBuffer` as well.

[...]

PR #165, opened 2020-10-06
Fix multithreaded wasm crash (solves #164)

chemicstry:
This solves #164 by using a preallocated `Uint8Array` for calling `crypto.getRandomValues`.

The `stdweb` implementation seems fine, as it already uses `Uint8Array`.

Should this fix be backported to `0.15`? There are a lot of crates that still use that version.

### DIFF
--- a/src/wasm-bindgen.rs
+++ b/src/wasm-bindgen.rs
@@ -10,8 +10,8 @@ use crate::Error;
 extern crate std;
 use std::thread_local;
 
-use wasm_bindgen::prelude::*;
 use js_sys::Uint8Array;
+use wasm_bindgen::prelude::*;
 
 // Maximum is 65536 bytes see https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
 const BROWSER_CRYPTO_BUFFER_SIZE: usize = 256;
@@ -45,10 +45,14 @@ pub(crate) fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
             }
             RngSource::Browser(ctx) => {
                 for chunk in dest.chunks_mut(BROWSER_CRYPTO_BUFFER_SIZE) {
-                    if ctx.crypto.get_random_values(&ctx.buf).is_err() {
+                    // chunk can be smaller than buf length
+                    // this creates a smaller view to the buf memory without allocation
+                    let sub_buf = ctx.buf.subarray(0, chunk.len() as u32);
+
+                    if ctx.crypto.get_random_values(&sub_buf).is_err() {
                         return Err(Error::WEB_GET_RANDOM_VALUES);
                     }
-                    ctx.buf.copy_to(chunk);
+                    sub_buf.copy_to(chunk);
                 }
             }
         };
@@ -71,10 +75,7 @@ fn getrandom_init() -> Result<RngSource, Error> {
 
         let buf = Uint8Array::new_with_length(BROWSER_CRYPTO_BUFFER_SIZE as u32);
 
-        let ctx = BrowserCryptoContext {
-            crypto,
-            buf,
-        };
+        let ctx = BrowserCryptoContext { crypto, buf };
 
         return Ok(RngSource::Browser(ctx));
     }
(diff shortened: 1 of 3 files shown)


src/wasm-bindgen.rs
```diff
@@ -33,17 +43,12 @@ pub(crate) fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
...
-                // see https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
-                //
-                // where it says:
-                //
-                // > A QuotaExceededError DOMException is thrown if the
-                // > requested length is greater than 65536 bytes.
-                for chunk in dest.chunks_mut(65536) {
-                    if n.get_random_values(chunk).is_err() {
+            RngSource::Browser(ctx) => {
+                for chunk in dest.chunks_mut(BROWSER_CRYPTO_BUFFER_SIZE) {
```
chemicstry, line 47: Used a constant here, because calling length on `Uint8Array` involves external javascript API.

chemicstry:
The CI seems to have failed due to unrelated reason. Can someone rerun?

src/wasm-bindgen.rs
```diff
@@ -10,11 +10,21 @@ use crate::Error;
...
 
+// Maximum is 65536 bytes see https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
+const BROWSER_CRYPTO_BUFFER_SIZE: usize = 256;
+
+struct BrowserCryptoContext {
+    crypto: BrowserCrypto,
+    // A temporary buffer backed by browser memory to avoid multithreaded wasm exception. See issue #164
+    buf: Uint8Array,
+}
+
```
josephlr, line 24: Instead of having this large structure, we could just do:

```rust
enum RngSource {
    Node(NodeCrypto),
    Browser(BrowserCrypto, Uint8Array),
}
```

and then add the comment about using the temporary buffer  when we actually use the temp buffer:

```rust
RngSource::Browser(crypto, buf) => {
    // getRandomValues does not work with all types of WASM memory, so
    // we initially write to browser memory to avoid an exception.
    for chunk in dest.chunks_mut(BROWSER_CRYPTO_BUFFER_SIZE) {
        ...
    }
}
```

src/wasm-bindgen.rs
```diff
@@ -63,7 +72,12 @@ fn getrandom_init() -> Result<RngSource, Error> {
             (_, crypto) if !crypto.is_undefined() => crypto,
             _ => return Err(Error::WEB_CRYPTO),
         };
-        return Ok(RngSource::Browser(crypto));
+
+        let buf = Uint8Array::new_with_length(BROWSER_CRYPTO_BUFFER_SIZE as u32);
+
+        let ctx = BrowserCryptoContext { crypto, buf };
+
+        return Ok(RngSource::Browser(ctx));
```
josephlr, line 80: This then becomes:

```rust
let buf = Uint8Array::new_with_length(BROWSER_CRYPTO_BUFFER_SIZE as u32);
return Ok(RngSource::Browser(crypto, buf));
```

src/wasm-bindgen.rs
josephlr: Good idea. Thanks!

josephlr (approved):
Looks good, a few minor changes. Also, now that we are depending on `js-sys` can you add a TODO where we use `Global` and `Self_` to replace that functionality with [`js_sys::global()`](https://docs.rs/js-sys/0.3.45/js_sys/fn.global.html)

josephlr:
> Should this fix be backported to `0.15`? There are a lot of crates that still use that version.

Ya that should be fine, once we make sure this works, I'll backport it (maybe once we also cleanup the `Self` usage).

src/wasm-bindgen.rs
chemicstry: Sorry for the late response, but do you still want me to fix this? Since you force pushed to my branch and already approved it

josephlr: Yes, fixing this would be preferable (I can also do it when i have time).

josephlr (approved):

PR #166, opened 2020-10-12
Update to cfg-if 1.0

mbrubeck:


newpavlov (approved):
IIRC we also can use `cfg-if = "1"`, which is a bit shorter.

mbrubeck:
> IIRC we also can use `cfg-if = "1"`, which is a bit shorter.

Done.

lopopolo:
@newpavlov can this get backported to the v0.1 branch?

I'm holding off on upgrading `getrandom` to v0.2 until `rand` v0.8 is released.

I'm hoping to get all of my deps updated to `cfg-if` v1.0.

PR #173, opened 2020-11-19
Backport update to cfg-if v1.0 to the v0.1 branch

newpavlov:
Requested in #166.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ appveyor = { repository = "rust-random/getrandom" }
 
 [dependencies]
 log = { version = "0.4", optional = true }
-cfg-if = "0.1.2"
+cfg-if = "1"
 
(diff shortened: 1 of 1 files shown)


Dragonrun1:
So this is now the only part of rand crates that is still using the old version of cfg-if. Do we have a timeline when this is getting merged?

newpavlov:
@Dragonrun1
It's blocked on @josephlr approving this PR.

Dragonrun1:
Actually I just updated to rand 0.8 and was looking with `cargo tree` and noticed it showing this package still using the old version. If you look at the Cargo.toml on getrandom it still shows `cfg-if = "0.1.2"` so that kind of why I decided to ask 😉

josephlr (approved):
Sorry for letting this slip by. Thanks for the backport!

PR #177, opened 2020-12-07
Remove Windows XP support

### LINKED ISSUE #65: Windows: RtlGenRandom vs BCryptGenRandom
IIUC both functions use the same algorithm [under the hood](https://news.ycombinator.com/item?id=12444228) (also see [this](https://stackoverflow.com/questions/48875929/rtlgenrandom-cryptgenrandom-or-other-winapi-to-generate-cryptographically-secure)). The main difference is that `BCryptGenRandom` will not work on Windows XP, but otherwise it looks like `BCryptGenRandom` should be preferred. IIRC Rust explicitly does not support Windows XP anymore, so it should not matter.

@chouquette
Can you help here? In rust-lang/rust#60260 you have used `BCryptGenRandom` for UWP targets and left `RtlGenRandom` for non-UWP targets. Is there a reason for that?

cc @GabrielMajeri

newpavlov:
Closes #65.


josephlr:
I'm totally down with removing XP support, but I'm not sure about switching over the implementation for Windows 7,8,10

Initially, I thought it would be fine, but the fact that Go and BoringSSL didnt switch means this probably should have a second look. I'll review the logic in those threads and try to see if I can find any (external or internal) reasoning for not switching.

I suspect that it doesn't matter, in which case we should switch implementations, but I want to be sure.

josephlr:
@newpavlov I kept the content of this PR the same, but redid the commit ordering. That way, when this PR is merged, `windows.rs` will keep the `windows_uwp.rs` git blame (instead having this PR appear like a change to `windows.rs`).

josephlr (approved):
Approved, see [my comment in the bug](https://github.com/rust-random/getrandom/issues/65#issuecomment-753634074) for why I think switching is fine.

PR #178, opened 2020-12-07
Remove direct stdweb support

### LINKED ISSUE #175: Remove stdweb support (RUSTSEC-2020-0056)
Stdweb seems to be abandoned, making the duplicate wasm feature options redundant.
https://github.com/RustSec/advisory-db/blob/master/crates/stdweb/RUSTSEC-2020-0056.md

newpavlov:
Closes #175.

IIUC `getrandom` still can be used with the `stdweb` toolchain, since its recent versions are compatible with `wasm-bindgen`.

cc @koute


josephlr:
Thanks for opening this, I'll merge this one last (after a point release), so there's at least some easy way for users stuck on `cargo web` to pin a version.

newpavlov:
I [think](https://github.com/rust-random/rand/issues/889#issuecomment-737179205) similarly regarding the version pinning escape hatch. But we already have released v0.2.0, haven't we?

josephlr:
> I [think](https://github.com/rust-random/rand/issues/889#issuecomment-737179205) similarly regarding the version pinning escape hatch. But we already have released v0.2.0, haven't we?

That's correct. I was thinking to release any minor fixes as 0.2.1 and then drop XP and stdweb support in 0.2.2.

I wouldn't consider dropping already broken support to be a major version change (but I'm not 100% sure). Does that sounds reasonable?

newpavlov:
>I was thinking to release any minor fixes as 0.2.1 and then drop XP and stdweb support in 0.2.2.

Honestly, I don't think it worth the trouble. But if you are willing to cut those releases, it's fine by me.

>Does that sounds reasonable?

Strictly speaking, I think it can be considered a breaking change. But practically speaking, I think potential WinXP/`stdweb` users would prefer to have the version pinning escape hatch, compared to a v0.3 release used by `rand v0.8` without any support of those use-cases whatsoever. So I am in favor of dropping their support in v0.2.

newpavlov:
@josephlr 
Can you review this and the other two PRs? It would be nice to release these changes before `rand v0.8`. Or do you plan to do the v0.2.1 release with minor fixes?

josephlr (approved):

josephlr (approved):
Not that this means we won't be able to use pure `cargo web` builds anymore, but I don't think that solution is maintained anymore, so that seems fine.

PR #179, opened 2020-12-07
Fix CI

newpavlov:


newpavlov:
Weird... For some reason  `curl -s https://api.github.com/repos/mozilla/geckodriver/releases/latest | jq -r ".tag_name"` returns `null` when executed by Travis, while locally it returns `v0.28.0` as expected.

newpavlov:
Will close it, since we have to migrate to GA anyway.

PR #180, opened 2020-12-14
Migrate CI to GitHub Actions

### LINKED ISSUE #174: Migrate CI to GitHub Actions?
It would mean that project will use a single CI service instead of two and CI will have a nicer integration with GitHub UI.

Relevant issue: rust-random/rand#1066

newpavlov:
Closes #174 

Converts most existing tests except related to WASM targets.


josephlr:
@newpavlov Do you know why AppVeyor Is still running? Is there some way to turn it off?

newpavlov:
I am not sure. Probably because we still have its config in the master branch. Hopefully it will be solved after merge.

dhardy:
Yup, I can turn it off: (1) there's a webhook still registered and (2) AppVeyor was set up through my account. I'll clear it for you.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,185 @@
...
+        ]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          target: ${{ matrix.target }}
+          toolchain: nightly
+          components: rust-src
```
josephlr, line 139: Do we need rust-src for these targets?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,185 @@
...
+          x86_64-fortanix-unknown-sgx,
+        ]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          target: ${{ matrix.target }}
+          toolchain: nightly
```
josephlr, line 138: Should we be using the stable toolchain for these targets?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,185 @@
...
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.toolchain }}
+          override: true
+      - run: cargo test
+      - run: cargo test --features std
+      - if: ${{ matrix.toolchain == 'nightly' }}
+        run: cargo build --benches
```
josephlr, line 46: Do you think we should have a job that actually runs the benchmarks? This could be done in a followup CL.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,185 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          components: rustfmt, clippy
+      - name: clippy
+        run: cargo clippy --all -- -D warnings
```
josephlr, line 183: Do we need `-D warnings` here if it's already set at the top?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,185 @@
+name: Tests
+
+on:
+  push:
+    branches: master
+  pull_request:
+    branches: master
```
josephlr, line 7: Should we also run these for the `0.1` branch?

josephlr (approved):
Thanks for doing this. I just have a few nits/questions

.github/workflows/tests.yml
newpavlov: The `0.1` branch will require a slightly different workflow (e.g. due to the differences in features), which I think can live separately in that branch.

.github/workflows/tests.yml
newpavlov: I think it will be a waste of CI time.

josephlr: Seems reasonable

.github/workflows/tests.yml
josephlr: Sounds good to just have a separate config in that branch.

.github/workflows/tests.yml
newpavlov: Travis config was using nightly, so I haven't changed it. I can try `stable` instead.

.github/workflows/tests.yml
newpavlov: According to the Travis config, yes.

.github/workflows/tests.yml
newpavlov: Hm, not sure. I will check it.

newpavlov: Yes, it was redundant. Removed.

.github/workflows/tests.yml
newpavlov: It was only needed for the `xbuild` job. Removed.

.github/workflows/tests.yml
newpavlov: On a second thought, I think it's worth to keep it `nightly` to allow early compiler bug detection if there will be any affecting us. I also will enable the cron job option to run CI periodically.

PR #183, opened 2020-12-18
Add Hermit support

newpavlov:
I am not sure if it's worth to add compile-time switch between `secure_rand32` and `secure_rand64`. Currently hermit supports only 64-bit targets.

cc @stlankes


newpavlov:
Ah, I forgot that direct Hermit support was intentionally removed in #61. It looks like the RDRAND implementation was not fixed.

PR #188, opened 2021-01-02
Release v0.2.1

newpavlov:
Assumes that #177 will be merged.

I wonder if it could be worth to move the badges to the header.


josephlr (approved):
Rebase this on #177 (and optionally #192) before merging

PR #190, opened 2021-01-03
Update 0.1 branch to use Github actions

josephlr:
Also update the badges in the README


newpavlov (approved):
Maybe it's worth to simply remove the commented out lines in the workflow config?

josephlr:
> Maybe it's worth to simply remove the commented out lines in the workflow config?

Ya, I'm just going to copy over the config from #192 and see if I can just use that.

josephlr:
Ok @newpavlov, this is now just a slightly modified version of the config in #192, and all 34 of the jobs are now passing. So once you approve that one, I'll merge this one.

newpavlov (approved):
I left some minor comments in #192 which apply here as well, but otherwise looks good!

josephlr:
> I left some minor comments in #192 which apply here as well, but otherwise looks good!

Applied the fixed from #192 to this branch

Issue #194, opened 2021-01-04
Add remaining CI improvements


josephlr:
After #192, we still have a few things we should add to our CI. These include:
- [x] Running the iOS binaries we build (ideally on both x86_64 and aarch64) #316 
- [x] Build and run `aarch64-apple-darwin` binaries (aka "Apple Silicon") #398 
- [x] Add emscripten tests (#196)
- [x] Add no-std test for a `custom` hook (#197)
- [x] ~~Cache toolchain downloads (see actions-rs/meta#21 and actions-rs/toolchain#54)~~ Not useful with move to https://github.com/dtolnay/rust-toolchain
- [x] Cache crates.io downloads/index (#198)

josephlr:
With #316 the only think we are waiting on is AArch64 macOS runners. This is tracked in https://github.com/github/roadmap/issues/528

PR #196, opened 2021-01-05
Add Emscripten tests

josephlr:
See #194 

This installs (and caches) the emsdk toolchain at the last version compatible w/ stable rust. It also tests on both asmjs and wasm32, uses node by default, and works around an asm.js bug.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,6 +11,7 @@ on:
 env:
   CARGO_INCREMENTAL: 0
   RUSTFLAGS: "-Dwarnings"
+  EMSDK_VERSION: 1.39.20 # Last emsdk compatible with Rust's LLVM 11
 
 jobs:
   check-doc:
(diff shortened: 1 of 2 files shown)


josephlr:
@newpavlov This is ready now.

I also took your advice from the last CI PR and make all the emscripten stuff one job.

.github/workflows/tests.yml
```diff
@@ -11,6 +11,7 @@ on:
 env:
   CARGO_INCREMENTAL: 0
   RUSTFLAGS: "-Dwarnings"
+  EMSDK_VERSION: 1.39.20 # Last emsdk compatible with Rust's LLVM 11
```
newpavlov, line 14: I think it would be better to move this variable to the `emscripten-tests` job, since it's the only affected job.

newpavlov (approved):

.github/workflows/tests.yml
josephlr: I forgot jobs can have `env`s, good catch

PR #197, opened 2021-01-05
Add test for custom RNG handler

josephlr:
See #194

This uses the fact that `wasm32-unknown-unknown` is an "unsupported" target. This means we can use the `"custom"` feature to define a custom handler, and then write tests to make sure that function is called.


newpavlov (approved):
Great idea!

PR #198, opened 2021-01-14
Add fixes for Caching and rustc-dep-of-std feature

josephlr:
We now cache the Cargo index (among other things) in our CI. This prevents some spurious test failures on Windows.

Also, Properly forward `rustc-dep-of-std` feature

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,6 +23,7 @@ jobs:
           profile: minimal
           toolchain: nightly # Needed for -Z minimal-versions
           override: true
+      - uses: Swatinem/rust-cache@v1
       - name: Install precompiled cargo-deadlinks
         run: |
           export URL=$(curl -s https://api.github.com/repos/deadlinks/cargo-deadlinks/releases/latest | jq -r '.assets[] | select(.name | contains("cargo-deadlinks-linux")) | .browser_download_url')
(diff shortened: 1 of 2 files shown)


newpavlov:
IIRC on aarch64 Hermit simply always returns an error, did this situation change? In other words, any use of `HashMap` with such code would result in a panic, so the existing exception in the `std` integration PR looks like a better choice for now. Personally, I would prefer for hermitcore/libhermit-rs#143 to be solved first and for `getrandom` to use the new API.

The CI changes are probably better done in a separate PR.

newpavlov:
I like all changes, except the Hermit addition. :) How about removing the Hermit part for now, renaming this PR, and opening a tracking issue for the Hermit part?

@stlankes
Any ETA on adding a Linux-like syscall to Hermit as described in the linked issue?

josephlr:
> I like all changes, except the Hermit addition. :) How about removing the Hermit part for now, renaming this PR, and opening a tracking issue for the Hermit part?

Done. This PR now just contains the caching and `rustc-dep-of-std` fixes.

josephlr:
https://github.com/rust-random/getrandom/issues/199 is now open to track all hermit <-> getrandom issues. Further discussion should continue there.

stlankes:
A small comment to the ARM port, it comes, but the current version isn't stable. It takes some time.

newpavlov (approved):

PR #200, opened 2021-01-17
Use doc_cfg to improve register_custom_getrandom docs

newpavlov:
The resulting docs will look like this:

### DIFF
--- a/src/error_impls.rs
+++ b/src/error_impls.rs
@@ -5,6 +5,7 @@
 // <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
+#![cfg_attr(docsrs, doc(cfg(feature = "std")))]
 extern crate std;
 
 use crate::Error;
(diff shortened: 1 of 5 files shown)


josephlr (approved):
Confirmed that this works by running:
```
RUSTDOCFLAGS="--cfg docsrs" cargo deadlinks -- --open --features=custom,std
```

src/error_impls.rs
```diff
@@ -5,6 +5,7 @@
 // <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
+#![cfg_attr(docsrs, doc(cfg(feature = "std")))]
```
newpavlov, line 8: This line is not needed.

josephlr:
@newpavlov I updated the docs tests to set `docsrs`, and also added a `doc(cfg(feature = "std"))` to our `std` trait implementations, this produces the following nice note:

rustdoc

newpavlov:
Ah, good to know!

src/error_impls.rs
josephlr: This is also what I thought, but apparently impl support got added.

PR #205, opened 2021-02-12
Replace build.rs with link attributes

### LINKED ISSUE #204: use link attributes on extern declarations rather than build.rs
The current build.rs is unnecessary, the link attribute can be used on the extern declaration blocks.

newpavlov:
Closes #204.

iOS code already had link attributes. To be extra safe I would prefer for someone to verify that this change does not break iOS targets.


aspenluxxxy:
I'm having issues related to ` note: ld: framework not found Security` with iOS targets and LTO, so this seems like it'd actually fix things.

josephlr (approved):
This looks correct. After reading the documentation for [the `cargo:rustc-link-lib=[KIND=]NAME` instruction in build scripts](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-link-libkindname), [the `#[link]` attribute](https://doc.rust-lang.org/reference/items/external-blocks.html#the-link-attribute), and [the `-l` rustc flag](https://doc.rust-lang.org/rustc/command-line-arguments.html#option-l-link-lib), I can confidently say that:
  - Adding `"cargo:rustc-link-lib=framework=Security"` to `build.rs`
  - Adding `#[link(name = "Security", kind = "framework")]` to an `extern "C"` block

do the exact same thing, namely passing `-l framework=Security` to rustc.

josephlr:
> Can't test as my dep tree pulls in 0.1 _and_ 0.2 of getrandom, though, so `[patches.crates-io]` doesn't work

Does this mean you would also like to see this change backported to `0.1`?

Lokathor:
if 0.1 is getting new releases there's no reason to not also do this in 0.1

PR #213, opened 2021-05-07
Remove extern crate std from source files

markrtuttle:
This pull request removes 'extern crate std' from source files for cleaner name spaces.

It is no longer necessary to import crates: https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html

This has been tested with 'cargo build' and 'cargo test'.

### DIFF
--- a/src/error_impls.rs
+++ b/src/error_impls.rs
@@ -6,7 +6,6 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 #![cfg_attr(docsrs, doc(cfg(feature = "std")))]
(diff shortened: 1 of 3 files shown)


dhardy:
`extern crate std` is not redundant when using `no_std`.

dhardy:
In all three cases `std` components get used after the `extern crate std`: it is required. The crate is marked as `no_std` which implies that `core` is imported and `std` is not, yet potentially (platform dependant) `std` is still available. This statement is importing the crate in those cases.

PR #215, opened 2021-05-12
Improve detection for Node JS

### LINKED ISSUE #214: nodejs thinks it is in a browser environment
I'm not sure if this is something particular to my setup but I was running into:
```    panicked at 'from_entropy failed: Web API self.crypto is unavailable', /Users/dev/.cargo/registry/src/github.com-1ecc6299db9ec823/rand_core-0.6.2/src/lib.rs:381:13```

when using a module created via `wasm-pack build --target nodejs ...`

Based on my experimentation, it appears https://github.com/rust-random/getrandom/blob/24716865bd70eb2128222667e05f5e2c15a6f929/src/js.rs#L61 global/global.self can exist in a nodejs environment so it follows the browser branch instead of the nodejs branch. I haven't set up a minimal reproduction but verified that `delete global.self` in my nodejs entry point resolved my issue. Thoughts?

josephlr:
Fixes #214 (CC: @devdoshi can you verify that this fixes your issue)

This uses `process.versions.node` to see if we are in Node.js rather
than using `global.self` to check if we are in a Browser.

This change also makes some minor `wasm_bindgen` improvements:
  - Use `js_sys::global()` to get the global object
  - Bind the Node's `require()` as `module.require`
  - Improving error messages
  - Being more strict about JsValue types (i.e. check that they are objects rather than just making sure they are not undefined)

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -162,7 +162,7 @@ fn internal_desc(error: Error) -> Option<&'static str> {
         Error::WINDOWS_RTL_GEN_RANDOM => Some("RtlGenRandom: Windows system function failure"),
         Error::FAILED_RDRAND => Some("RDRAND: failed multiple times: CPU issue likely"),
         Error::NO_RDRAND => Some("RDRAND: instruction not supported"),
-        Error::WEB_CRYPTO => Some("Web API self.crypto is unavailable"),
+        Error::WEB_CRYPTO => Some("Web Crypto API is unavailable"),
         Error::WEB_GET_RANDOM_VALUES => Some("Web API crypto.getRandomValues is unavailable"),
         Error::VXWORKS_RAND_SECURE => Some("randSecure: VxWorks RNG module is not initialized"),
         Error::NODE_CRYPTO => Some("Node.js crypto module is unavailable"),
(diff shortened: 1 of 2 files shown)


newpavlov (approved):
Looks good, but let's wait for confirmation from @devdoshi before merging.

BTW you have a typo in the PR title.

devdoshi:
Thanks @josephlr I'm checking right now via https://github.com/opticdev/optic/pull/786/checks?check_run_id=2611731376#step:6:25 but I still see the issues. Perhaps I'm not setting the dependency correctly? let me know if you can't see the logs but I believe you should be able to.

devdoshi:
Now that I look at the issue with fresh eyes, I'm using Jest on a React UI project. It's set up to run in a jsdom environment. However, the jsdom environment does not provide an implementation for global.crypto. It appears I have to do that manually in the Jest setup, but I can definitely do that. So, I think the issue still remains that detection should be improved, but in my particular situation I think I should be making my jsdom environment include a crypto implementation, since jsdom is intended to make it look like it's in a browser, so your detection would probably continue to be fooled.

PR #222, opened 2021-08-14
redox: Switch to /dev/urandom

josephlr:
From https://github.com/briansmith/ring/pull/1341 apparently on Redox the `/dev/urandom` device now exists. Changing the implementation for consistency between RNG crates.

@jackpot51 does this look good to you? Also, where is this device documented?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,7 +22,7 @@
 //! | Dragonfly BSD     | `*‑dragonfly`      | [`getrandom()`][22] if available, otherwise [`/dev/random`][8]
 //! | Solaris, illumos  | `*‑solaris`, `*‑illumos` | [`getrandom()`][9] if available, otherwise [`/dev/random`][10]
 //! | Fuchsia OS        | `*‑fuchsia`        | [`cprng_draw`][11]
-//! | Redox             | `*‑redox`          | [`rand:`][12]
+//! | Redox             | `*‑redox`          | [`/dev/urandom`][12]
 //! | Haiku             | `*‑haiku`          | `/dev/random` (identical to `/dev/urandom`)
 //! | SGX               | `x86_64‑*‑sgx`     | [RDRAND][18]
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -22,7 +22,7 @@
 //! | Dragonfly BSD     | `*‑dragonfly`      | [`getrandom()`][22] if available, otherwise [`/dev/random`][8]
 //! | Solaris, illumos  | `*‑solaris`, `*‑illumos` | [`getrandom()`][9] if available, otherwise [`/dev/random`][10]
 //! | Fuchsia OS        | `*‑fuchsia`        | [`cprng_draw`][11]
-//! | Redox             | `*‑redox`          | [`rand:`][12]
+//! | Redox             | `*‑redox`          | [`/dev/urandom`][12]
```
josephlr, line 25: We may need to update the documentation link here

jackpot51 (approved):

jackpot51:
/dev/urandom is now a symlink to rand:, and rand: is always a non-blocking CPRNG

src/lib.rs
newpavlov: The link simply points to relevant Redox source code. It would be nice to replace it with a proper documentation link, but AFAIK it currently does not exist.

newpavlov (approved):
IIUC `/dev/urandom` was added mostly for compatibility purposes and reading `rand:` is more idiomatic. From practical point of view both options are equivalent to each other, so I am fine with this change.

@jackpot51 
BTW how Redox handles entropy pool initialization on system startup? The preferred behavior for us would be for it to block `rand:`/`dev/urandom` reading until the pool gets properly initialized.

src/lib.rs
josephlr: @jackpot51 does Redox have any docs describing either `rand:` or `/dev/urandom`?

jackpot51: @josephlr not really

jackpot51:
@newpavlov rand: (and /dev/urandom with it being a symlink) will not return any data until it is initialized. It will block on read.

PR #233, opened 2021-10-14
Use `NonZeroU32::new_unchecked` to convert wasi error

Cyborus04:
The check in `NonZeroU32::new` isn't needed, as `wasi::Error` is `NonZeroU16` internally, so `wasi::Error::raw_error` will never return zero

### DIFF
--- a/src/wasi.rs
+++ b/src/wasi.rs
@@ -12,10 +12,12 @@ use core::num::NonZeroU32;
 use wasi::random_get;
 
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
-    unsafe { random_get(dest.as_mut_ptr(), dest.len()) }.map_err(|e: wasi::Error| {
-        // convert wasi's Error into getrandom's NonZeroU32 error
-        // SAFETY: `wasi::Error` is `NonZeroU16` internally, so `e.raw_error()`
-        // will never return 0
-        unsafe { NonZeroU32::new_unchecked(e.raw_error() as u32) }.into()
-    })
+    unsafe {
+        random_get(dest.as_mut_ptr(), dest.len()).map_err(|e: wasi::Error| {
+            // convert wasi's Error into getrandom's NonZeroU32 error
+            // SAFETY: `wasi::Error` is `NonZeroU16` internally, so `e.raw_error()`
+            // will never return 0
+            NonZeroU32::new_unchecked(e.raw_error() as u32).into()
+        })
+    }
 }


josephlr:
Thanks for opening this.

I think it's fine to leave this as a call to `unwrap`. If what you say is accurate, it won't ever trigger and doesn't cost anything. It also has the advantage of avoiding unnecessary `unsafe` code. It's not like the single call to unwrap is going to cost anything.

It might be worth adding/chaining a comment explaining why it is OK to call `unwrap()` though, as we do wish to avoid panics in this library.

newpavlov (approved):
At the time I thought that having `unwrap` in the error conversion path would be fine since compiler should be able to remove the panic completely, but according to [godbolt](https://rust.godbolt.org/z/nhE8j8b6c) surprisingly it's not the case today. Ideally we would simply use the `Into` impl, but it would require bumping MSRV to 1.41. Can you please add a respective TODO comment?

Also I think it could be worth to join the `unsafe`s into one:
```rust
unsafe {
    random_get(dest.as_mut_ptr(), dest.len())
        .map_err(|e| NonZeroU32::new_unchecked(e.raw_error() as u32).into())
}
```

josephlr:
@newpavlov do you thing the unsafe code is worth the "cost" of a not-taken jump instruction? It's always going to be predicted correctly regardless. There's also not a guarantee in the [`wasi::Error::raw_error()`](https://docs.rs/wasi/0.10.2+wasi-snapshot-preview1/wasi/struct.Error.html#method.raw_error) docs that the function cannot return zero.

I think the better solution is for `wasi::Error` to implement a conversion to `NonZeroU16`, and for us to just use that.

newpavlov:
>do you thing the unsafe code is worth the "cost" of a not-taken jump instruction?

Considering that the conversion happens only for errors, which are already extremely rare, I am not concerned about runtime efficiency in this case at all. But I am not fond of having clearly redundant and easily removable panic paths in compiled results.

>There's also not a guarantee in the wasi::Error::raw_error() docs that the function cannot return zero.

IIRC it's guaranteed by WASI docs, i.e. it specifies that zero return code always represents "success". I agree that ideally we would use a method returning `NonZeroU16`, but unfortunately it does not exist currently. We could use the raw [`random_get`](https://docs.rs/wasi/0.10.2+wasi-snapshot-preview1/wasi/wasi_snapshot_preview1/fn.random_get.html) directly, but I don't think it's worth the trouble and it may result in additional headache when hypothetical `wasi_snapshot_preview2` is released.

josephlr (approved):
> IIRC it's guaranteed by WASI docs, i.e. it specifies that zero return code always represents "success". I agree that ideally we would use a method returning `NonZeroU16`, but unfortunately it does not exist currently. We could use the raw [`random_get`](https://docs.rs/wasi/0.10.2+wasi-snapshot-preview1/wasi/wasi_snapshot_preview1/fn.random_get.html) directly, but I don't think it's worth the trouble and it may result in additional headache when hypothetical `wasi_snapshot_preview2` is released.

Sounds reasonable, I agree that using the underlying raw call is not worth it. Approved modulo @newpavlov's suggested changes.

src/wasi.rs
```diff
@@ -12,10 +12,12 @@ use core::num::NonZeroU32;
 use wasi::random_get;
 
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
-    unsafe { random_get(dest.as_mut_ptr(), dest.len()) }.map_err(|e: wasi::Error| {
-        // convert wasi's Error into getrandom's NonZeroU32 error
-        // SAFETY: `wasi::Error` is `NonZeroU16` internally, so `e.raw_error()`
-        // will never return 0
-        unsafe { NonZeroU32::new_unchecked(e.raw_error() as u32) }.into()
-    })
+    unsafe {
```
Cyborus04, line 15: I had no idea `unsafe` blocks propagated into closures

Cyborus04:
> I think it would be a good addition to wasi, but I think we can merge this PR with the raw_error code and update it later after release with the new impl (or method) will land.

I'll go make a PR for that, then make another one here for if/when that one gets merged

Issue #224, opened 2021-09-01
This crate hits a buggy Webpack 5 warning when building for wasm32-unknown


josephlr:
See this comment for context: https://github.com/rust-random/getrandom/pull/137#issuecomment-909874327

Previous toil on this issue:
Fix in this crate for Webpack 4: https://github.com/rust-random/getrandom/pull/137
Rustwasm bug: https://github.com/rustwasm/wasm-pack/issues/822
Webpack bug closed as WONTFIX: https://github.com/webpack/webpack/issues/8826
Advice on how to fix this issue for Webpack: https://github.com/webpack/webpack/issues/8826#issuecomment-671402668

Basically, our [`#[wasm_bindgen]`  declaration](https://github.com/rust-random/getrandom/blob/30308ae845b0bf3839e5a92120559eaf56048c28/src/js.rs#L105-L107) generates Javascript that looks like:
```js
module.require(getStringFromWasm0(arg0, arg1));
```
Despite this being valid JavaScript, Webpack doesn't like it for some reason.

It seems like our options to fix this are:
  - Tell users how to ignore the warning (it is just a warning after all)
  - Tell users how to configure Webpack to not get this error (involves some `package.json` stuff I don't understand)
  - File an upstream bug with Webpack to properly support wasm
  - File an upstream bug with the Rust `wasm_bindgen` tool to annotate `module.require` with some sort of "ignore warning" requirement
  - Use a [Node.js dynamic `import()`](https://nodejs.org/api/esm.html#esm_import_expressions) call
  - Use `wasm_bindgen`'s [`module` attribute](https://rustwasm.github.io/docs/wasm-bindgen/reference/attributes/on-js-imports/module.html`).
  - Use `wasm_bindgen`'s [`inline_js` attribute](https://rustwasm.github.io/docs/wasm-bindgen/reference/js-snippets.html#using-inline_js)
  - Get upstream Rust to finally have `wasm32-unknown-browser` and `wasm32-unknown-node` targets, so that we don't have to do browser vs Node detection at runtime.

Things we shouldn't try to do:
  - Replace this with another hack to trick the bundler (I don't want to keep dealing with this issue)
  - `eval("require(crypto)")`

However, **I don't have the expertise to solve this problem**. If anyone (@Pauan, @huacnlee  , @Herohtar) knows how to fix this issue, I would happily accept a PR provided the solution is documented and our CI keeps passing (we don't use the Webpack bundler in our CI, hence why we missed this).

[...]

PR #234, opened 2021-10-15
Fix Webpack warning caused by dynamic require

mjhanninen:
This PR reverts the changes in commit 120a1d7  **partially** . In particular, it changes the wasm-bindgen binding of `module.requirement` back to the older form. As a result the generated Javascript binding code looks now like:

```
 getObject(arg0).require(getStringFromWasm0(arg1, arg2))
```

As a consequence Webpack does not detect this as a package import and, thus, does not attempt to bundle it which solves problem. It is safe skip the bundling of the `crypto` package because (1) it is imported only on Node and (2) it comes with Node.

Relevant issues:

- https://github.com/rust-random/getrandom/issues/224
- https://github.com/rustwasm/wasm-pack/issues/822

### DIFF
--- a/src/js.rs
+++ b/src/js.rs
@@ -59,7 +59,9 @@ pub(crate) fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
 fn getrandom_init() -> Result<RngSource, Error> {
     let global: Global = global().unchecked_into();
     if is_node(&global) {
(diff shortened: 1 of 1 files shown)


josephlr (approved):

josephlr:
Sorry for the delay. I was hoping there would eventually be a better solution to #224 but nobody has chimed in, so I'll just merge this (as it does seem to work).

Issue #199, opened 2021-01-17
Hermit is not currently supported by getrandom


josephlr:
[Hermitcore](https://hermitcore.org/) is a unikernel OS, written in Rust, with [Tier 3 `libstd` support](https://doc.rust-lang.org/beta/rustc/platform-support.html#tier-3) for `aarch64-unknown-hermit` and `x86_64-unknown-hermit`.

See #198 and #183 for previous attempts at this.

Other PRs/issues of note:
  - My initial question to the Hermit folks about their RNG: https://github.com/hermitcore/rusty-hermit/issues/63
  - Change that adds Cryptographic RNG support to Hermit (x86 only): https://github.com/hermitcore/libhermit-rs/pull/98
  - Change that exposes RNG APIs as part of `hermit-abi`: https://github.com/hermitcore/rusty-hermit/pull/64
  - Making `hermit-abi` always `no_std`: hermitcore/rusty-hermit#103
  - Fix for the RDRAND implementation in `libhermit-rs`: https://github.com/hermitcore/libhermit-rs/pull/154
  - Feature request to have hermit use a buffer-based API: hermitcore/libhermit-rs#143

We should figure out:
  - How/when we want to add support to `getrandom`
  - Is the existing [`secure_rand64` API](https://hermitcore.github.io/rusty-hermit/hermit_abi/fn.secure_rand64.html) sufficient?
  - Hermit's RNG apis current don't work on `aarch64`, is that an issue?

newpavlov:
>Hermit's RNG apis current don't work on aarch64, is that an issue?

Yes, it's an issue, since it causes workarounds like [this](https://github.com/rust-lang/rust/blob/582566d513ee4e1e33c9f6f98d913e6a025b6283/library/std/Cargo.toml#L26-L28). In user code it can be worked around slightly more elegantly using the `custom` feature, but we can not rely on it in `std`. Note that currently Hermit's `std` simply [uses](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/hermit/mod.rs#L86-L89) a constant seed on all targets.

josephlr:
I think the most reasonable approach here is the following:

1. Simply add support for hermit using their `secure_rand64` API and the `hermit-abi` crate. Given that hermit is a unikernel, I don't see any reason to push them for a buffer-based API (so hermitcore/libhermit-rs#143 isn't too important).
[...]

PR #236, opened 2021-11-27
Added x86_64-unknown-hermit support

sjmackenzie:
[Later](https://github.com/hermitcore/rusty-hermit/pull/64#issuecomment-980668685) once hermit supports `aarch64` rng I'll make a patch to support it. 

For more context see #199

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -306,10 +306,10 @@ jobs:
           components: rust-src
           override: true
       - uses: Swatinem/rust-cache@v1
+      - name: Hermit (x86-64 only)
+        run: cargo build -Z build-std=core --target=x86_64-unknown-hermit
       - name: UEFI (RDRAND)
         run: cargo build -Z build-std=core --features=rdrand --target=x86_64-unknown-uefi
-      - name: Hermit (RDRAND)
-        run: cargo build -Z build-std=core --features=rdrand --target=x86_64-unknown-hermit
       - name: L4Re (RDRAND)
         run: cargo build -Z build-std=core --features=rdrand --target=x86_64-unknown-l4re-uclibc
       - name: VxWorks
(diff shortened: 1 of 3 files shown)


newpavlov (approved):
Hm, I would prefer a proper buffer-based API, but I guess this code should be a fine stop-gap solution.

I will leave this PR open for some time in the case if @josephlr has any concerns.

Meanwhile can you please add an `[UNRELEASED]` section to the changelog with an entry containing description of this PR? Also it would be nice to add a Hermit build test to the CI workflow.

sjmackenzie:
Agreed on the buffer-based API, I suggest this is good feedback for @stlankes & @josephlr!

I believe `getrandom` CI already has hermit workflow support, correct me if I'm wrong please: https://github.com/rust-random/getrandom/blob/f5e33009edc2ac5ea59f7dde68709e9572b94458/.github/workflows/tests.yml#L312

Ideally I'd love to try get this stop-gap merged relatively soon as there are a number of transitive dependencies I need to lobby to bump to a hermit-supported-getrandom. I'm choosing hermit as the sole target platform for an existing project. 

@josephlr I'd love your feedback on this stop-gap solution.

@newpavlov I'd pushed the requested changes, if you could be so kind and review them.

Thanks!

sjmackenzie:
By buffer-based API do you mean you can pass in an arbitrary sized `u8` buffer and it'd be filled accordingly? If so, this is what is happening in this PR.

newpavlov:
>By buffer-based API do you mean you can pass in an arbitrary sized u8 buffer and it'd be filled accordingly? If so, this is what is happening in this PR.

I meant a "syscall" on the Hermit side, like requested in https://github.com/hermitcore/libhermit-rs/issues/143 

>I believe getrandom CI already has hermit workflow support, correct me if I'm wrong please:

Ah, I forgot about that. You can simply remove the `--features=rdrand` part then since it's not longer needed.

sjmackenzie:
Thanks for removing the `--features=rdrand`, it fell on the wayside. Appreciate the merge!

PR #245, opened 2022-01-29
Enable support for espidf (esp32 family of microcontrollers)

thorhs:
Shamelessly forked from  https://github.com/esp-rs-compat/getrandom.

Has been tested on my local MCU.

### DIFF
--- a/src/espidf.rs
+++ b/src/espidf.rs
@@ -20,9 +20,6 @@ pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
     // https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/random.html
     //
     // However tracking if some of these entropy sources is enabled is way too difficult to implement here
-    //
-    // Using esp_fill_random since it has some optimizations regarding filling a byte array from an
-    // u32 source.  See https://github.com/espressif/esp-idf/blob/master/components/esp_hw_support/hw_random.c
     unsafe { esp_fill_random(dest.as_mut_ptr().cast(), dest.len()) };
 
     Ok(())
--- a/src/espidf.rs
+++ b/src/espidf.rs
@@ -1,0 +1,24 @@
+// Copyright 2021 Developers of the Rand project.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Implementation for ESP-IDF
+use crate::Error;
+
+pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
+    // ESP-IDF fails and returns -1 only when the passed buffer is NULL, which cannot happen in our case:
+    // https://github.com/espressif/esp-idf/blob/master/components/newlib/random.c#L33
+    //
+    // Not that NOT enabling WiFi, BT, or the voltage noise entropy source (via `bootloader_random_enable`)
+    // will cause ESP-IDF to return pseudo-random numbers based on the voltage noise entropy, after the initial boot process:
+    // https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/random.html
+    //
+    // However tracking if some of these entropy sources is enabled is way too difficult to implement here
+    unsafe { libc::getrandom(dest.as_mut_ptr().cast(), dest.len(), 0) };
+
+    Ok(())
+}
--- a/src/espidf.rs
+++ b/src/espidf.rs
@@ -8,17 +8,22 @@
 
 //! Implementation for ESP-IDF
 use crate::Error;
+use core::ffi::c_void;
+
+extern "C" {
+    fn esp_fill_random(buf: *mut c_void, len: usize) -> u32;
+}
 
 pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
-    // ESP-IDF fails and returns -1 only when the passed buffer is NULL, which cannot happen in our case:
-    // https://github.com/espressif/esp-idf/blob/master/components/newlib/random.c#L33
-    //
     // Not that NOT enabling WiFi, BT, or the voltage noise entropy source (via `bootloader_random_enable`)
     // will cause ESP-IDF to return pseudo-random numbers based on the voltage noise entropy, after the initial boot process:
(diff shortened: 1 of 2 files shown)


newpavlov:
Shouldn't we use [`esp_fill_random`](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/random.html#_CPPv415esp_fill_randomPv6size_t)?

thorhs:
That was my initial implementation, but then I found the esp-rs-compat repo and figured it was a better solution. Using esp_fill_random requires including a dependency on esp-idf-sys, which includes the burden on updating that dependency as time goes by. 

I have a version standing by using esp_fill_random, if that is preferred. 

As for the build test, I think this requires a patched rust compiler from espressif, so I'm not sure how to write that in a portable manner.

src/lib.rs
```diff
@@ -142,6 +143,7 @@
 //! [20]: https://www.unix.com/man-page/mojave/4/random/
 //! [21]: https://www.freebsd.org/cgi/man.cgi?query=getrandom&manpath=FreeBSD+12.0-stable
 //! [22]: https://leaf.dragonflybsd.org/cgi/web-man?command=getrandom
+//! [23]: https://cygwin.com/git/?p=newlib-cygwin.git;a=blob;f=winsup/cygwin/include/sys/random.h;hb=HEAD#l22
```
josephlr, line 146: If we are depending on the esp targets, we should change the docs (and doc link) to point to:

https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/random.html

josephlr:
@thorhs if you resync onto master you should be able to fix the CI issues.

> That was my initial implementation, but then I found the esp-rs-compat repo and figured it was a better solution. Using esp_fill_random requires including a dependency on esp-idf-sys, which includes the burden on updating that dependency as time goes by.

This is a good point (about avoiding dependencies), which is why we generally just manually include the `extern "C"` declarations for various platforms. For example, our implementations for [Windows](https://github.com/rust-random/getrandom/blob/e1a9c9e74068d29bb4158d7282e398591da69811/src/windows.rs#L14-L22), [iOS](https://github.com/rust-random/getrandom/blob/e1a9c9e74068d29bb4158d7282e398591da69811/src/ios.rs#L13-L16), or [SOLID](https://github.com/rust-random/getrandom/blob/e1a9c9e74068d29bb4158d7282e398591da69811/src/solid.rs#L13-L15) do this. So we would probably just have something like
```rust
extern "C" {
  fn esp_random() -> u32;
}
```
In general, we try to use the lowest level RNG API available on a given platform (provided that it is safe and stable). So depending directly on the `esp_random` functions is fine. 

> We even may prefer `esp_random` since it can allow Rust compiler to optimize filling loop in some cases.

@newpavlov's point is a good one, although using `esp_random` avoids being able to inline the [`esp_random` implementation](https://github.com/espressif/esp-idf/blob/95c02402bed275aa6d0d5e3ed309ca0dc3b6694b/components/esp_hw_support/hw_random.c#L24-L55) into `esp_fill_random`, so it might not really matter. This stuff also isn't fast to begin with.

However, it seems like `esp_fill_random` wasn't added until [v3.2 of esp-idf](https://github.com/espressif/esp-idf/commit/83a179abb0d2bd0bfd50da71edaa63ea1add0071), so if supporting older versions matters, we should use `esp_random`. If not, I think we should use `esp_fill_random` for its simple implementation.

> As for the build test, I think this requires a patched rust compiler from espressif, so I'm not sure how to write that in a portable manner.

If support for `espidf` hasn't been merged into upstream Rust yet, that's fine. Just add a TODO to the `build-std` section of [`tests.yml`](https://github.com/rust-random/getrandom/blob/master/.github/workflows/tests.yml). However, it seems like https://github.com/rust-lang/rust/pull/87666 added the necessary support to rust, what else is left to do upstream?

src/espidf.rs
```diff
@@ -0,0 +1,29 @@
...
+
+pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
+    // Not that NOT enabling WiFi, BT, or the voltage noise entropy source (via `bootloader_random_enable`)
+    // will cause ESP-IDF to return pseudo-random numbers based on the voltage noise entropy, after the initial boot process:
+    // https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/random.html
+    //
+    // However tracking if some of these entropy sources is enabled is way too difficult to implement here
+    //
+    // Using esp_fill_random since it has some optimizations regarding filling a byte array from an
+    // u32 source.  See https://github.com/espressif/esp-idf/blob/master/components/esp_hw_support/hw_random.c
```
newpavlov, line 25: These two lines are misleading in my opinion. There are not optimizations in the code itself, compiler is simply able to inline implementation of `esp_random` during compilation of the library. Also this optimization is still possible to perform with enabled LTO. So I think it's better to simply remove these lines.

src/lib.rs
```diff
@@ -142,6 +143,7 @@
 //! [20]: https://www.unix.com/man-page/mojave/4/random/
 //! [21]: https://www.freebsd.org/cgi/man.cgi?query=getrandom&manpath=FreeBSD+12.0-stable
 //! [22]: https://leaf.dragonflybsd.org/cgi/web-man?command=getrandom
+//! [23]: https://github.com/espressif/esp-idf/blob/master/components/esp_hw_support/hw_random.c
```
newpavlov, line 146: As noted earlier by @josephlr it's better to link the relevant page from https://docs.espressif.com/, not the source code.

newpavlov (commented):
>using esp_random avoids being able to inline the esp_random implementation into esp_fill_random, so it might not really matter

The most common use-case for `getrandom` is seeding of user-space PRNGs, which often use `u32` or `u64` as seeds. Using `esp_random` would allow generated data to stay in registers without round-tripping through stack. But agree that it's a minuscule difference, so I am fine with `esp_fill_random`.

src/lib.rs
thorhs: I have updated the documentation link to reference the API documentation.

src/espidf.rs
thorhs: Ok, so just remove lines 22-25?

newpavlov: Only 24-25.

thorhs: Done

newpavlov (approved):

thorhs:
Could this also be merged into 0.1 branch?  Should I submit a new PR for that?

newpavlov:
If you **really** need it in practice, then we could do it. But otherwise I don't think it's worth the trouble, since it would be better for your dependencies to use up-to-date `rand` crates.

PR #247, opened 2022-02-05
Release v0.2.5

newpavlov:
Also includes minor doc changes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,7 +12,7 @@
 //!
 //! | Target            | Target Triple      | Implementation
 //! | ----------------- | ------------------ | --------------
-//! | Linux, Android    | `*‑linux‑*`        | [`getrandom`][getrandom_linux] system call if available, otherwise [`/dev/urandom`][2] after successfully polling `/dev/random`
+//! | Linux, Android    | `*‑linux‑*`        | [`getrandom`][1] system call if available, otherwise [`/dev/urandom`][2] after successfully polling `/dev/random`
 //! | Windows           | `*‑windows‑*`      | [`BCryptGenRandom`]
 //! | macOS             | `*‑apple‑darwin`   | [`getentropy`][3] if available, otherwise [`/dev/random`][4] (identical to `/dev/urandom`)
 //! | iOS               | `*‑apple‑ios`      | [`SecRandomCopyBytes`]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,26 +12,26 @@
 //!
 //! | Target            | Target Triple      | Implementation
 //! | ----------------- | ------------------ | --------------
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -12,26 +12,26 @@
...
-//! | macOS             | `*‑apple‑darwin`   | [`getentropy()`][19] if available, otherwise [`/dev/random`][20] (identical to `/dev/urandom`)
-//! | iOS               | `*‑apple‑ios`      | [`SecRandomCopyBytes`][4]
-//! | FreeBSD           | `*‑freebsd`        | [`getrandom()`][21] if available, otherwise [`kern.arandom`][5]
-//! | OpenBSD           | `*‑openbsd`        | [`getentropy`][6]
-//! | NetBSD            | `*‑netbsd`         | [`kern.arandom`][7]
-//! | Dragonfly BSD     | `*‑dragonfly`      | [`getrandom()`][22] if available, otherwise [`/dev/random`][8]
-//! | Solaris, illumos  | `*‑solaris`, `*‑illumos` | [`getrandom()`][9] if available, otherwise [`/dev/random`][10]
-//! | Fuchsia OS        | `*‑fuchsia`        | [`cprng_draw`][11]
-//! | Redox             | `*‑redox`          | [`/dev/urandom`][12]
+//! | Linux, Android    | `*‑linux‑*`        | [`getrandom`][getrandom_linux] system call if available, otherwise [`/dev/urandom`][2] after successfully polling `/dev/random`
```
josephlr, line 15: Shouldn't this be `[1]` instead of `[getrandom_linux]`

src/lib.rs
```diff
@@ -12,26 +12,26 @@
...
+//! | Windows           | `*‑windows‑*`      | [`BCryptGenRandom`]
+//! | macOS             | `*‑apple‑darwin`   | [`getentropy`][3] if available, otherwise [`/dev/random`][4] (identical to `/dev/urandom`)
+//! | iOS               | `*‑apple‑ios`      | [`SecRandomCopyBytes`]
+//! | FreeBSD           | `*‑freebsd`        | [`getrandom`][5] if available, otherwise [`kern.arandom`][6]
+//! | OpenBSD           | `*‑openbsd`        | [`getentropy`][7]
+//! | NetBSD            | `*‑netbsd`         | [`kern.arandom`][8]
```
josephlr, line 21: Can this also not be made into a named link (or does the dot mess things up)?

josephlr (approved):
Thanks for handling this (and the doc fixes).

src/lib.rs
newpavlov: We have two different `arandom` links for FreeBSD and NetBSD. I guess we could name them as `arandom_freebsd` and `arandom_netbsd`.

josephlr: Makes sense, keeping them as numbers is fine

newpavlov:
The CI failure is not relevant to changes in this PR. See #249.

PR #252, opened 2022-03-14
libc: retry open when interrupted

stoeckmann:
The open call can be interrupted. Since sys_fill_exact covers this for
read operation already, it should be done here as well.

Signed-off-by: Tobias Stoeckmann <tobias@stoeckmann.org>

Shoutout to [c3h2-ctf](https://twitter.com/c3h2_ctf).

### DIFF
--- a/src/util_libc.rs
+++ b/src/util_libc.rs
@@ -107,9 +107,15 @@ cfg_if! {
 // SAFETY: path must be null terminated, FD must be manually closed.
 pub unsafe fn open_readonly(path: &str) -> Result<libc::c_int, Error> {
     debug_assert_eq!(path.as_bytes().last(), Some(&0));
-    let fd = open(path.as_ptr() as *const _, libc::O_RDONLY | libc::O_CLOEXEC);
-    if fd < 0 {
-        return Err(last_os_error());
+    loop {
+        let fd = open(path.as_ptr() as *const _, libc::O_RDONLY | libc::O_CLOEXEC);
+        if fd >= 0 {
+            return Ok(fd);
+        }
+        let err = last_os_error();
+        // We should try again if open() was interrupted.
+        if err.raw_os_error() != Some(libc::EINTR) {
+            return Err(err);
+        }
     }
-    Ok(fd)
 }


src/util_libc.rs
```diff
@@ -107,9 +107,16 @@ cfg_if! {
...
+        let fd = open(path.as_ptr() as *const _, libc::O_RDONLY | libc::O_CLOEXEC);
+        if fd < 0 {
+            let err = last_os_error();
+            // We should try again if the call was interrupted.
+            if err.raw_os_error() != Some(libc::EINTR) {
+                return Err(err);
+            }
+        } else {
+            return Ok(fd);
+        }
```
josephlr, line 120: I'm wondering if it's more idiomatic to do:
```rust
let mut fd;
while { fd = open(...); fd < 0 } {
    let err = last_os_error();
    // We should try again if open() was interrupted.
    if err.raw_os_error() != Some(libc::EINTR) {
        return Err(err);
    }
}
Ok(fd)
```

josephlr (changes requested):
Nice catch here!!

src/util_libc.rs
stoeckmann: I have taken the second version, which received a thumbs up. :)

josephlr (approved):

josephlr:
I'll give @newpavlov a little time to take a look, and then I'll merge this.

newpavlov (approved):
I wonder if it could be worth to limit number of retries to be extra safe by replacing `loop { .. }` with something like this:
```rust
for _ in 0..256 {
    // ..
}
Err(Error::TOO_MANY_ITERATIONS) // find a better name
```

josephlr:
> I wonder if it could be worth to limit number of retries to be extra safe by replacing `loop { .. }` with something like this:
> 
> ```rust
> for _ in 0..256 {
>     // ..
> }
> Err(Error::TOO_MANY_ITERATIONS) // find a better name
> ```

It looks like [Rust's `open` implementation on Linux calls `cvt_r`](https://github.com/rust-lang/rust/blob/d6f3a4ecb48ead838638e902f2fa4e5f3059779b/library/std/src/sys/unix/fs.rs#L829-L845) and [`cvt_r` just retries indefinitely](https://github.com/rust-lang/rust/blob/f19851069efd6ee1fe899a469f08ad2d66e76050/library/std/src/sys/unix/mod.rs#L206-L217). So we should probably mirror that behavior here.

PR #253, opened 2022-03-19
Update to wasi 0.11

Thomasdezeeuw:
The main breaking change between v0.10 and v0.11 is that Error is
removed in favour of Errno. Unfortunately we can't create an Errno from
outside the wasi create so we're loosing some debug information for
errors.

I've opened an issue to add back such a constructor, see
<https://github.com/bytecodealliance/wasi/issues/64>.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ cfg-if = "1"
 compiler_builtins = { version = "0.1", optional = true }
 core = { version = "1.0", optional = true, package = "rustc-std-workspace-core" }
 
-[target.'cfg(any(unix, target_os = "wasi"))'.dependencies]
+[target.'cfg(unix)'.dependencies]
 libc = { version = "0.2.120", default-features = false }
 
 [target.'cfg(target_os = "wasi")'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ core = { version = "1.0", optional = true, package = "rustc-std-workspace-core" 
 libc = { version = "0.2.120", default-features = false }
 
 [target.'cfg(target_os = "wasi")'.dependencies]
-wasi = "0.10"
+wasi = "0.11"
 
 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
 wasm-bindgen = { version = "0.2.62", default-features = false, optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ cfg-if = "1"
 compiler_builtins = { version = "0.1", optional = true }
 core = { version = "1.0", optional = true, package = "rustc-std-workspace-core" }
 
-[target.'cfg(unix)'.dependencies]
+[target.'cfg(any(unix, target_os = "wasi"))'.dependencies]
 libc = { version = "0.2.120", default-features = false }
 
 [target.'cfg(target_os = "wasi")'.dependencies]
--- a/src/error.rs
+++ b/src/error.rs
@@ -110,8 +110,10 @@ cfg_if! {
             core::str::from_utf8(&buf[..idx]).ok()
         }
     } else if #[cfg(target_os = "wasi")] {
-        fn os_err(errno: i32, _buf: &mut [u8]) -> Option<wasi::Error> {
-            wasi::Error::from_raw_error(errno as _)
+        fn os_err(errno: i32, _buf: &mut [u8]) -> Option<wasi::Errno> {
+            // Can't create `Errno` at the moment, see
+            // <https://github.com/bytecodealliance/wasi/issues/64>.
+            None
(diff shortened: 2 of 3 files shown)


newpavlov:
Personally, I would prefer for `wasi` developers to fix the Error issue first before moving from v0.10.

josephlr:
While I'd also prefer the `wasi` types to have the error information, it looks like wasi's `libc` [provides `strerror_r`](https://github.com/WebAssembly/wasi-libc/blob/main/libc-top-half/musl/src/string/strerror_r.c) which [contains the errno information](https://github.com/WebAssembly/wasi-libc/blob/main/libc-top-half/musl/src/errno/__strerror.h). So we might be able to get the debug info just from the call to libc, and handle things on `wasi` like we do on `unix`.

This also seems to be [what the standard library does](https://github.com/rust-lang/rust/blob/d53246fedde4c193eae8a003546a8f0f9f85d223/library/std/src/sys/wasi/os.rs#L41-L51).

Cargo.toml
```diff
@@ -21,7 +21,7 @@ core = { version = "1.0", optional = true, package = "rustc-std-workspace-core"
 libc = { version = "0.2.64", default-features = false }
```
josephlr, line 21: Can you update this to:
```toml
[target.'any(cfg(unix), target_os = "wasi")'.dependencies]
libc = { ... }
```
We shouldn't need to increase the crate version as https://github.com/rust-lang/libc/commit/58ba8129c4a46b7e61ec61cd61c32d2082fefcd8 was added in `libc` 0.2.53

src/error.rs
```diff
@@ -110,8 +110,10 @@ cfg_if! {
             core::str::from_utf8(&buf[..idx]).ok()
         }
     } else if #[cfg(target_os = "wasi")] {
-        fn os_err(errno: i32, _buf: &mut [u8]) -> Option<wasi::Error> {
-            wasi::Error::from_raw_error(errno as _)
+        fn os_err(errno: i32, _buf: &mut [u8]) -> Option<wasi::Errno> {
```
josephlr, line 113: Can you just change the 
```rust
if #[cfg(unix)] {
```
above to
```rust
if #[cfg(any(unix, target_os = "wasi"))] {
```

and remove the `wasi` specific section here.

josephlr (changes requested):

Thomasdezeeuw:
I've changed over to use `libc::stderror`and needed to force push to resolve the merge conflict.

src/wasi.rs
```diff
@@ -9,15 +9,11 @@
 //! Implementation for WASI
 use crate::Error;
 use core::num::NonZeroU32;
-use wasi::random_get;
+use wasi::wasi_snapshot_preview1::random_get;
```
josephlr, line 12: Why not just use `wasi::random_get`?

newpavlov:
Adding `libc` simply for error messages is not great... Especially considering that `random_get` errors are extremely unlikely and there are no `random_get`-specific error codes.

src/wasi.rs
Thomasdezeeuw: Because the new [`Errno`](https://docs.rs/wasi/latest/wasi/struct.Errno.html) from wasi, which replaced `Error`, can actually hold 0. `Errno::as_raw` returns `u16`, which may be zero, thus invalidating the previous code calling `NonZeroU32::new_unchecked`. However handling zero ourselves ensures we can call `NonZeroU32::new_unchecked` again.

josephlr:
> Adding `libc` simply for error messages is not great... Especially considering that `random_get` errors are extremely unlikely and there are no `random_get`-specific error codes.

So do you think we should just not have error messages for `wasi` errors then? I would be fine with that.

src/wasi.rs
josephlr: But it won't actually be zero if the function returns `Err`, so it seems fine.

newpavlov:
Yes, before the `wasi` crate improves handling of errors, I think it's fine to simply do not have any error messages for WASI.

src/wasi.rs
newpavlov: It means that soundness of our code relies on an implementation detail of `wasi`. I am not sure we can rely on it, even though it's a very reasonable assumption.

Thomasdezeeuw: > It would mean that soundness of our code relies on an implementation detail of `wasi`. I am not sure we can rely on it, even though it's a very reasonable assumption.

This is the exact reason I decided to switch.

Thomasdezeeuw:
So I should drop the libc dependency and remove the error message?

josephlr:
> remove

That sounds great!

Thomasdezeeuw:
@josephlr done.

Thomasdezeeuw:
Any update on this?

josephlr (approved):

PR #259, opened 2022-04-21
Auto-emit feature js on wasm32-unknown-unknown target

hanabi1224:
Currently, feature `js` has to be enabled explicitly to build on `wasm32-unknown-unknown` target.  This PR tries to automatically emit feature `js` on `wasm32-unknown-unknown` target with cargo build script, to make `cargo build --target wasm32-unknown-unknown` work.  Feature `js` is still kept in `Cargo.toml` for compatibility, it can be removed in future release.

All CI tests pass on my fork: https://github.com/hanabi1224/getrandom/actions/runs/2203432125

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -205,7 +205,9 @@ jobs:
           tar -C /tmp -xzf /tmp/binaries.tar.gz --strip-components=1
           mv /tmp/wasm-bindgen-test-runner ~/.cargo/bin
       - name: Test (Node)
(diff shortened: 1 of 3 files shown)


josephlr:
Unfortunately something like this is not going to work. We have this feature because not all `wasm32-unknown-unonown` platforms support JS.

We want it to be possible for users in this situation to use the `custom` feature of this crate to provide their own implementation.

hanabi1224:
Thx for the explanation

PR #263, opened 2022-06-07
Use AtomicPtr instead of AtomicUsize for Weak

### LINKED ISSUE #262: Weak Bindings Violate Strict Provenance in MIRI
Because the [`Weak` struct](https://github.com/rust-random/getrandom/blob/v0.2.6/src/util_libc.rs#L77-L84) is defined as having `LazyUsize` for the address, it does not maintain pointer provenance and thus fails when run under MIRI with `-Zmiri-strict-provenance`.

The solution would be to define a `LazyPtr` type that utilizes `AtomicPtr` instead of `AtomicUsize`.

josephlr:
Fixes #262

This allows Strict Provenance to work properly. It also now matches what `libstd` does:
https://github.com/rust-lang/rust/blob/9f7e997c8bc3cacd2ab4eb75e63cb5fa9279c7b0/library/std/src/sys/unix/weak.rs#L85-L141

Also, while reading the `libstd` code, I noticed that they use an `Acquire` fence and `Release` store as the returned pointer should have "consume" semantics. I changed our code to do something slightly stronger (Acquire load and Release store) as it's simpler and definitely correct.

### DIFF
--- a/src/util_libc.rs
+++ b/src/util_libc.rs
@@ -108,7 +108,7 @@ impl Weak {
     }
 
     // Return the address of a function if present at runtime. Otherwise,
-    // return null. Multiple callers can call ptr() concurrently. It will
+    // return None. Multiple callers can call ptr() concurrently. It will
     // always return _some_ value returned by libc::dlsym. However, the
     // dlsym function may be called multiple times.
     pub fn ptr(&self) -> Option<NonNull<c_void>> {
--- a/src/util_libc.rs
+++ b/src/util_libc.rs
@@ -81,8 +81,9 @@ pub fn sys_fill_exact(
 
 // A "weak" binding to a C function that may or may not be present at runtime.
 // Used for supporting newer OS features while still building on older systems.
-// Based off of the DlsymWeak struct in libstd (src/sys/unix/weak.rs), except
-// that the caller must cast self.ptr() to a function pointer.
+// Based off of the DlsymWeak struct in libstd:
+// https://github.com/rust-lang/rust/blob/1.61.0/library/std/src/sys/unix/weak.rs#L84
+// except that the caller must manually cast self.ptr() to a function pointer.
 pub struct Weak {
     name: &'static str,
     addr: AtomicPtr<c_void>,
--- a/src/util_libc.rs
+++ b/src/util_libc.rs
@@ -6,8 +6,13 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 #![allow(dead_code)]
-use crate::{util::LazyUsize, Error};
-use core::{num::NonZeroU32, ptr::NonNull};
+use crate::Error;
+use core::{
+    num::NonZeroU32,
+    ptr::NonNull,
+    sync::atomic::{AtomicPtr, Ordering},
+};
+use libc::c_void;
 
 cfg_if! {
     if #[cfg(any(target_os = "netbsd", target_os = "openbsd", target_os = "android"))] {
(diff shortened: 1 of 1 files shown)


josephlr:
CC @nvzqz if you want to take a look.

src/util_libc.rs
```diff
@@ -76,29 +81,48 @@ pub fn sys_fill_exact(
 
 // A "weak" binding to a C function that may or may not be present at runtime.
 // Used for supporting newer OS features while still building on older systems.
-// F must be a function pointer of type `unsafe extern "C" fn`. Based off of the
-// weak! macro in libstd.
+// Based off of the DlsymWeak struct in libstd (src/sys/unix/weak.rs), except
+// that the caller must cast self.ptr() to a function pointer.
```
nvzqz, line 85: Perhaps link directly to the struct?

```suggestion
// Based off of the DlsymWeak struct in libstd (https://github.com/rust-lang/rust/blob/1.61.0/library/std/src/sys/unix/weak.rs#L84),
// except that the caller must cast self.ptr() to a function pointer.
```

nvzqz (approved):
Thanks for this! Using `[patch.crates-io]` in my `Cargo.toml`, I was able to verify that this makes my code's tests now pass under strict provenance.

src/util_libc.rs
```diff
@@ -76,29 +81,49 @@ pub fn sys_fill_exact(
...
     }
 
-    // Return a function pointer if present at runtime. Otherwise, return null.
-    pub fn ptr(&self) -> Option<NonNull<libc::c_void>> {
-        let addr = self.addr.unsync_init(|| unsafe {
-            libc::dlsym(libc::RTLD_DEFAULT, self.name.as_ptr() as *const _) as usize
-        });
-        NonNull::new(addr as *mut _)
+    // Return the address of a function if present at runtime. Otherwise,
+    // return null. Multiple callers can call ptr() concurrently. It will
```
fbstj, line 111:
```suggestion
    // return `None`. Multiple callers can call ptr() concurrently. It will
```
should this be a doc comment?

src/util_libc.rs
```diff
@@ -76,29 +81,49 @@ pub fn sys_fill_exact(
...
+        // Despite having only a single atomic variable (self.addr), we still
+        // need a "consume" ordering as we will generally be reading though
+        // this value (by calling the function we dlsym'ed). Rust lacks this
+        // ordering, so we have to go with the next strongest: Acquire/Release.
+        // As noted in libstd, this might be unnecessary.
+        let mut addr = self.addr.load(Ordering::Acquire);
+        if addr == Self::UNINIT {
+            let symbol = self.name.as_ptr() as *const _;
+            addr = unsafe { libc::dlsym(libc::RTLD_DEFAULT, symbol) };
+            self.addr.store(addr, Ordering::Release);
```
newpavlov, line 124: Since the initialization code can run several times without impacting correctness and these load/stores are not related to any other memory operations, I think using `Relaxed` ordering should be correct.

newpavlov (approved):

src/util_libc.rs
nvzqz: These load/stores relate to memory operations within `dlsym`. Usually I'd err on the side of caution per [this comment in libstd's `weak.rs`](https://github.com/rust-lang/rust/blob/1.61.0/library/std/src/sys/unix/weak.rs#L105-L121) from https://github.com/rust-lang/rust/pull/79039.

However, I may be convinced that relaxed is fine here. @thomcc, can you please weigh in?

newpavlov: >These load/stores relate to memory operations within dlsym.

In this case `Release` store should be sufficient, while the load can be `Relaxed`.

josephlr: I updated the implementation here to exactly match [the implementation in `libstd`](https://github.com/rust-lang/rust/blob/282445a288b3c409a998eb72e51827537e81abd3/library/std/src/sys/unix/weak.rs#L101-L126) and I clarified some comments.

Note that we need a `Release` op and an `Acquire` op to get the "ordered before" relationship. A `Relaxed` and `Release` op don't give us that.

thomcc: Yes, this needs acquire. It’s not true that they don’t relate to other memory operations — you read through the pointer (by calling it), which counts.

PR #266, opened 2022-06-14
Move to using #[wasm_bindgen(module)]

josephlr:


josephlr:
This is infeasible, we should prefer a solution like: #284

PR #269, opened 2022-06-29
Add support for forced custom implementation

### LINKED ISSUE #230: It should be possible to use a custom implementation when RDRAND is available
See https://github.com/rust-random/getrandom/blob/30308ae845b0bf3839e5a92120559eaf56048c28/src/lib.rs#L208-L217:

Presently, the RDRAND implementation supersedes the custom implementation, and there's no way to prevent that for SGX or when the `rdrand` feature is explicitly set. Further, there's no way to use RDRAND through this crate from a custom implementation.

More generally, in any target we should be able to avoid using RDRAND output directly. For example, in an SGX and/or UEFI or operating system kernel, one may want to implement their own RNG, perhaps one that conforms to external security standards (FIPS, NIST, etc.), perhaps using RDRAND as one entropy source. Or, the implementation may want to mix the RDRAND output into a DRBG (using NIST terminology) to protect against unknown issues with RDRAND.

And, I think we `getrandom` even avoid defaulting to RDRAND on any target. 

In UEFI and SGX, I would like to force the use of my custom implementation, and I'd like to implement mine partially in terms of RDRAND.

Concretely, I suggest:
1. Stop defaulting to the RDRAND implementation on SGX targets.
2. When the `rdrand` feature is requested, expose a new API to get the (raw, unbiased) results of RDRAND, independent of the current `getrandom()` API.
3. When the `rdrand` and `custom` features are requested, provide an API for registering the RDRAND implementation as the custom implementation, and document that environments such as UEFI and SGX must do this registration before using `getrandom`, just like any custom implementation.

Protryon:
This adds a feature `force-custom` that will supersede `target_os` based rng selection.

This fixes #230, and supports my particular use case, which is building for `wasm32-unknown-unknown` in theory, but using `wasm32-unknown-emscripten` in practice, with a custom rand impl.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,8 @@ rdrand = []
 js = ["wasm-bindgen", "js-sys"]
 # Feature to enable custom RNG implementations
 custom = []
+# Feature to force usage of custom RNG implementation
+force-custom = []
 # Unstable feature to support being a libstd dependency
 rustc-dep-of-std = [
   "compiler_builtins",
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -162,7 +162,7 @@ mod error;
 mod util;
 // To prevent a breaking change when targets are added, we always export the
 // register_custom_getrandom macro, so old Custom RNG crates continue to build.
-#[cfg(feature = "custom")]
+#[cfg(any(feature = "custom", feature = "force-custom"))]
 mod custom;
 #[cfg(feature = "std")]
 mod error_impls;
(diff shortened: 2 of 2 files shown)


newpavlov:
We intentionally did not want an easy way for users to overwrite the "default" sources on supported platforms, since it may cause security issues. Imagine a crate which silently enables the `force-custom` feature and uses a backdoored RNG.

If users absolutely need to use a different entropy source on supported platforms, then they always can [override](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html) `getrandom` with their custom crate.

josephlr:
I agree with @newpavlov here. We don't want to allow unilateral overriding of the RNG, and there are other ways to achieve your goal here. Note that when we add #275, it may help your use-case, as you could then link against a different getentropy.

I'm going to leave #230 open, as it deals with out desired preference for `rdrand` vs `custom`.

PR #270, opened 2022-07-07
use $crate in macro

leoyvens:
This is the recommended way of referring to the current crate in a macro.


josephlr (approved):
Per https://doc.rust-lang.org/reference/macros-by-example.html#hygiene `$crate` was added in 1.30 and our MSRV is 1.34, so LGTM.

PR #271, opened 2022-07-08
Added possibility to pass uninit arrays to random generator

AngelicosPhosphoros:
This can be useful to skip extra work on caller side and it is more logical because we don't really need to have initialized values to fill them with new values.

### DIFF
--- a/src/3ds.rs
+++ b/src/3ds.rs
@@ -7,10 +7,12 @@
 // except according to those terms.
 
 //! Implementation for Nintendo 3DS
+use core::mem::MaybeUninit;
+
 use crate::util_libc::sys_fill_exact;
 use crate::Error;
 
-pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
+pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
     sys_fill_exact(dest, |buf| unsafe {
         libc::getrandom(buf.as_mut_ptr() as *mut libc::c_void, buf.len(), 0)
     })
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -155,6 +155,8 @@
 #![warn(rust_2018_idioms, unused_lifetimes, missing_docs)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
+use core::mem::MaybeUninit;
+
 #[macro_use]
 extern crate cfg_if;
 
@@ -259,5 +261,41 @@ pub fn getrandom(dest: &mut [u8]) -> Result<(), Error> {
     if dest.is_empty() {
         return Ok(());
     }
-    imp::getrandom_inner(dest)
+    imp::getrandom_inner(
+        // SAFETY: `MaybeUninit` have same size and alignment as wrapped value
+        // and it is safe to convert init value to MaybeUninit.
+        // See https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#layout
+        unsafe {
+            core::slice::from_raw_parts_mut(dest.as_mut_ptr() as *mut MaybeUninit<u8>, dest.len())
+        },
+    )
+}
+
+/// This version uses rdbuf-impl
+/// It is intended to replace function that accepts `&mut [u8]`
+/// once MSRV is bumped so you may want to use this to make easier switch
+///
+/// Caller must clear dest before use because function only append data to buffer
+/// until fill it to capacity.
+///
+/// Blocking is possible, at least during early boot; see module documentation.
+///
+/// In general, `getrandom` will be fast enough for interactive usage, though
+/// significantly slower than a user-space CSPRNG; for the latter consider
+/// [`rand::thread_rng`](https://docs.rs/rand/*/rand/fn.thread_rng.html).
+#[cfg(any(doc, feature = "rdbuf-impl"))]
+pub fn getrandom_buf(dest: &mut std::io::ReadBuf) -> Result<(), Error> {
+    if dest.capacity() == dest.filled_len() {
+        // Already filled
+        return Ok(());
+    }
+    // SAFETY:
+    // 1. getrandom_inner must fill all bytes provided if succeedes.
+    // 2. getrandom_inner must not read from provided buffer because it can be uninit.
+    unsafe {
+        imp::getrandom_inner(dest.unfilled_mut())?;
+        dest.assume_init(dest.capacity());
+        dest.set_filled(dest.capacity());
+    }
+    Ok(())
 }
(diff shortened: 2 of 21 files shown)


src/windows.rs
```diff
@@ -6,23 +6,29 @@
...
-use core::{ffi::c_void, num::NonZeroU32, ptr};
 
 const BCRYPT_USE_SYSTEM_PREFERRED_RNG: u32 = 0x00000002;
 
 #[link(name = "bcrypt")]
 extern "system" {
     fn BCryptGenRandom(
         hAlgorithm: *mut c_void,
-        pBuffer: *mut u8,
+        pBuffer: *mut MaybeUninit<u8>,
```
AngelicosPhosphoros, line 22: T and MaybeUninit<T> is equal for FFI purposes: https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#layout

src/lib.rs
```diff
@@ -259,5 +261,42 @@ pub fn getrandom(dest: &mut [u8]) -> Result<(), Error> {
...
+    }
+    let res = imp::getrandom_inner(dest);
+
+    // Help miri to catch mistakes.
+    // If user calls `assume_init()` on elements,
+    // miri would show error.
+    #[cfg(miri)]
+    if res.is_err() {
+        dest.fill(MaybeUninit::uninit());
+    }
```
AngelicosPhosphoros, line 300: I am thinking that maybe this is better to be put into `rand` crate and let `rand` use `getrandom_to_uninit` even for initialized slices.

AngelicosPhosphoros:
Also related PR: https://github.com/rust-random/rand/pull/1241

notgull:
I like this idea. Not only because of the potential utility, but because I think having methods like this to deal with uninit slices should be a pattern throughout the ecosystem.

My only thing is that I wonder if using `ReadBuf` would be a better fit.

josephlr:
Personally, I would want to wait until there's a standard interface in the standard library (ReadBuf), instead of implementing our own non-standard interface to deal with uninitialized buffers.

josephlr:
Also, as this is explicitly a performance improvement, could you also add some benchmarks so that we can see the perf impact of this change (vs just zeroing the uninit buffer and calling getrandom normally).

AngelicosPhosphoros:
> Personally, I would want to wait until there's a standard interface in the standard library (ReadBuf), instead of implementing our own non-standard interface to deal with uninitialized buffers.

It looks good but it is defined in std library and not in core and getrandom can be used in environments without std.

AngelicosPhosphoros:
@josephlr @notgull What you think about this updated API?

notgull:
I like this!

Is there any particular reason why `ReadBuf` isn't available on `core`?

src/3ds.rs
```diff
@@ -7,10 +7,12 @@
 // except according to those terms.
 
 //! Implementation for Nintendo 3DS
+use core::mem::MaybeUninit;
```
josephlr, line 10: Note: This increases our MSRV to 1.36.0 so is a breaking change. However, if we gate everything behind the `read_buf` feature, this shouldn't be an issue.

josephlr:
> > Personally, I would want to wait until there's a standard interface in the standard library (ReadBuf), instead of implementing our own non-standard interface to deal with uninitialized buffers.
> 
> It looks good but it is defined in std library and not in core and getrandom can be used in environments without std.

Then I think for now we should add an off-by-default `"read_buf"` feature that:
  - Depends on the `"std"` features 
  - Enables the `read_buf` nightly feature
  - Documents that the feature requires `libstd` and that it requires a recent-ish nightly

Then we could have code like this in our `Cargo.toml`:
```toml
# Nightly-only implementation with `std::io::ReadBuf` struct.
read_buf = ["std"]
```
and in `lib.rs`
```rust
#![cfg_attr(feature = "read_buf", feature(read_buf))]

#[cfg(feature = "std")]
extern crate std;
#[cfg(feature = "read_buf")]
use std::io::ReadBuf;
```

Then everywhere else we can just refer to `ReadBuf` or `crate::ReadBuf`. That way if `ReadBuf` moves to core, we only have to change 1-2 lines.

josephlr:
@AngelicosPhosphoros before we get too far into this, could you add some benchmarks to show that this is actually faster when dealing with uninitialized buffers? I don't want you to waste your time if it turns out there isn't a notable difference.

Specifically, I would want to compare:
  - Directly reading random data into an uninitialized buffer
  - Initializing a buffer (say with zero), then calling the normal `getrandom` method

I would recommend focusing on the `rdrand` implementation or the `linux` implementation for benchmarking, as those are the two fastest IIRC, so it's where the cost of writing the buffer twice would be the most noticeable. Feel free to add the benchmarks to this PR using [`#[bench]`](https://doc.rust-lang.org/unstable-book/library-features/test.html).

If it turns out there's a statistically significant improvement, I think there is a path to incorporate such functionality into `getrandom`. However, it might be the case that this optimization doesn't matter for `getrandom` (as it calls the underlying OS RNG, which can be slow compared to `memset`/`bzero`). In that case, this optimization might only matter for dealing with a userspace CSPRNG, so such uninitialized buffer code would be best to live in the `rand` (and similar) creates, rather than this crate.

josephlr:
See https://godbolt.org/z/9Waeqdrh5 for some example implementations using the existing API and the proposed new API.

The generated code confirms that regardless of how we use the existing API, it basically becomes a call to `memset`+`getrandom_impl`, while using the new API lets us only have a call to `getrandom_impl`.

Now it's just a question if `getrandom_impl` is fast enough for the initial call to `memset` to matter.

josephlr:
@AngelicosPhosphoros actually I just realized that we don't need this PR to benchmark the cost of initialization + calling getrandom. #272 adds benchmarks that should be able to measure the cost.

newpavlov:
On Linux and "common"  hardware memset is approximately 2 orders of magnitude faster than calling `getrandom` (~10 GB/s vs ~80 MB/s), so I highly doubt we will be able to see the performance difference on common platforms. The situation may change on constrained targets (e.g. embedded devices).

Either way, I don't think it's worth to use `ReadBuf` and `MaybeUninit` in our case. A better solution could be something like this:
```rust
// It's safe to assume that the buffer gets fully filled if the function returns `Ok(())`
pub unsafe fn getrandom_raw(buf_ptr: *mut u8, buf_len: usize) -> Result<(), Error> { .. }

#[inline(always)]
pub getrandom(buf: &mut [u8]) ->  Result<(), Error> {
    unsafe { getrandom_raw(buf.as_mut_ptr(), buf.len()) }
}
```

josephlr:
I think @newpavlov's suggestion in https://github.com/rust-random/getrandom/pull/271#issuecomment-1179703647 is the way we will want to go with this. Closing in favor of #279

PR #279, opened 2022-08-18
Implement raw API

newpavlov:
A simpler alternative to #271.

It introduces `getrandom_raw` function with the following signature:
```rust
pub unsafe fn getrandom_raw(dst: *mut u8, len: usize) -> Result<(), Error> { ... }
```

Note that `getrandom` is built on top of this function and all `getrandom_impl`s also now implement the pointer-based API. It fits quite well for all backends except the `js` one.

It allows to fill uninitialized arrays with random data.

Additionally, as a safe wrapper around `getrandom_raw` this PR adds `getrandom_uninit`:
```rust
pub fn getrandom_uninit(dst: &mut [MaybeUninit<u8>]) -> Result<&mut [u8], Error> { ... }
```

Because of `MaybeUninit` MSRV gets bumped to 1.36 (the same MSRV as `rand`).

I have reworked the benchmark and I get the following results:
```
running 6 tests
test bench_large     ... bench:  26,509,202 ns/iter (+/- 131,913) = 79 MB/s
test bench_large_raw ... bench:  26,478,315 ns/iter (+/- 96,640) = 79 MB/s
test bench_page      ... bench:      52,642 ns/iter (+/- 376) = 77 MB/s
test bench_page_raw  ... bench:      52,592 ns/iter (+/- 328) = 77 MB/s
test bench_seed      ... bench:         888 ns/iter (+/- 4) = 36 MB/s
test bench_seed_raw  ... bench:         887 ns/iter (+/- 5) = 36 MB/s
```
On my (Linux-based) PC the difference between `getrandom` and `getrandom_raw` is less than 1% and barely noticeable. The former zeroizes stack-based buffer before calling `getrandom`, while the latter uses `MaybeUninit` to skip the unnecessary initialization step.

### DIFF
--- a/benches/mod.rs
+++ b/benches/mod.rs
@@ -1,59 +1,21 @@
 #![feature(test)]
 extern crate test;
 
-use std::{
-    alloc::{alloc_zeroed, dealloc, Layout},
-    ptr::NonNull,
-};
-
-// AlignedBuffer is like a Box<[u8; N]> except that it is always N-byte aligned
-struct AlignedBuffer<const N: usize>(NonNull<[u8; N]>);
-
-impl<const N: usize> AlignedBuffer<N> {
-    fn layout() -> Layout {
-        Layout::from_size_align(N, N).unwrap()
-    }
-
-    fn new() -> Self {
-        let p = unsafe { alloc_zeroed(Self::layout()) } as *mut [u8; N];
-        Self(NonNull::new(p).unwrap())
-    }
-
-    fn buf(&mut self) -> &mut [u8; N] {
-        unsafe { self.0.as_mut() }
-    }
-}
-
-impl<const N: usize> Drop for AlignedBuffer<N> {
-    fn drop(&mut self) {
-        unsafe { dealloc(self.0.as_ptr() as *mut u8, Self::layout()) }
-    }
-}
-
-// Used to benchmark the throughput of getrandom in an optimal scenario.
-// The buffer is hot, and does not require initialization.
 #[inline(always)]
 fn bench<const N: usize>(b: &mut test::Bencher) {
-    let mut ab = AlignedBuffer::<N>::new();
-    let buf = ab.buf();
     b.iter(|| {
+        let mut buf = [0u8; N];
         getrandom::getrandom(&mut buf[..]).unwrap();
         test::black_box(&buf);
     });
     b.bytes = N as u64;
 }
 
-// Used to benchmark the throughput of getrandom is a slightly less optimal
-// scenario. The buffer is still hot, but requires initialization.
 #[inline(always)]
-fn bench_with_init<const N: usize>(b: &mut test::Bencher) {
-    let mut ab = AlignedBuffer::<N>::new();
-    let buf = ab.buf();
+fn bench_raw<const N: usize>(b: &mut test::Bencher) {
     b.iter(|| {
-        for byte in buf.iter_mut() {
-            *byte = 0;
-        }
-        getrandom::getrandom(&mut buf[..]).unwrap();
+        let mut buf = core::mem::MaybeUninit::<[u8; N]>::uninit();
+        unsafe { getrandom::getrandom_raw(buf.as_mut_ptr().cast(), N).unwrap() };
         test::black_box(&buf);
     });
     b.bytes = N as u64;
@@ -71,8 +33,8 @@ fn bench_seed(b: &mut test::Bencher) {
     bench::<SEED>(b);
 }
 #[bench]
-fn bench_seed_init(b: &mut test::Bencher) {
-    bench_with_init::<SEED>(b);
+fn bench_seed_raw(b: &mut test::Bencher) {
+    bench_raw::<SEED>(b);
 }
 
 #[bench]
@@ -80,8 +42,8 @@ fn bench_page(b: &mut test::Bencher) {
     bench::<PAGE>(b);
 }
 #[bench]
-fn bench_page_init(b: &mut test::Bencher) {
-    bench_with_init::<PAGE>(b);
+fn bench_page_raw(b: &mut test::Bencher) {
+    bench_raw::<PAGE>(b);
 }
 
 #[bench]
@@ -89,6 +51,6 @@ fn bench_large(b: &mut test::Bencher) {
     bench::<LARGE>(b);
 }
 #[bench]
-fn bench_large_init(b: &mut test::Bencher) {
-    bench_with_init::<LARGE>(b);
+fn bench_large_raw(b: &mut test::Bencher) {
+    bench_raw::<LARGE>(b);
 }
--- a/src/custom.rs
+++ b/src/custom.rs
@@ -78,9 +78,9 @@ macro_rules! register_custom_getrandom {
     ($path:path) => {
         // We use an extern "C" function to get the guarantees of a stable ABI.
         #[no_mangle]
-        extern "C" fn __getrandom_custom(dest: *mut u8, len: usize) -> u32 {
+        extern "C" fn __getrandom_custom(dst: *mut u8, len: usize) -> u32 {
             let f: fn(&mut [u8]) -> Result<(), $crate::Error> = $path;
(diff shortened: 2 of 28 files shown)


src/js.rs
```diff
@@ -27,28 +27,33 @@ thread_local!(
...
 
-pub(crate) fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
+pub(crate) unsafe fn getrandom_inner(mut dst: *mut u8, mut len: usize) -> Result<(), Error> {
     RNG_SOURCE.with(|result| {
         let source = result.as_ref().map_err(|&e| e)?;
 
         match source {
             RngSource::Node(n) => {
-                if n.random_fill_sync(dest).is_err() {
+                let dst = core::slice::from_raw_parts_mut(dst, len);
```
newpavlov, line 36: Note that strictly speaking this line is unsound, since we can create slice which points to uninitialized memory. Practically speaking, it's fine since we know that no one will read from the created slice, but we still break the Rust rules.

Ideally we would use a pointer-based analog of `random_fill_sync`, but I am not sure whether it's possible to create it. At the very least it could require changes in `js-sys`.

As a temporary workaround we could use a similar approach to the browser branch, i.e. create a temporary buffer initialized with zeros, fill it with random data, and finally copy the random data to the target buffer. Obviously, it will introduce unnecessary overhead which we try to remove with the raw API in the first place.

newpavlov: As a better solution, I added initialization of the input buffer with zeros. It's not ideal, but should be tolerable.

benches/mod.rs
```diff
@@ -1,59 +1,22 @@
...
-    let mut ab = AlignedBuffer::<N>::new();
-    let buf = ab.buf();
+fn bench_raw<const N: usize>(b: &mut test::Bencher) {
     b.iter(|| {
-        for byte in buf.iter_mut() {
-            *byte = 0;
-        }
-        getrandom::getrandom(&mut buf[..]).unwrap();
+        let mut buf = core::mem::MaybeUninit::<[u8; N]>::uninit();
+        // TODO: use `cast` on MSRV bump to 1.38
```
josephlr, line 18: This code can only be built on nightly anyway as it requires `#![feature(test)]`, so I'm fine using Cast.

newpavlov: Done.

newpavlov:
@josephlr
Could you approve it? Unless, of course, you plan to review it more thoroughly later or have some concerns about the approach.

josephlr (commented):
The overall approach here seems fine, but there would be two things I would like to see:

First, is a better way to deal us not having `chunks` anymore. A bunch of our code now has repeated sections where we do the same loop manipulations, I'm wondering if we could but something in `util.rs` to make this nicer.  Maybe something similar to `sys_fill_exact`.

Second, I think we should add `#![deny(unsafe_op_in_unsafe_fn)]` so that we can better see which operations are unsafe inside our (sometimes large) unsafe functions.

newpavlov:
>A bunch of our code now has repeated sections where we do the same loop manipulations, I'm wondering if we could but something in util.rs to make this nicer.

Yeah, it could worthwhile. I will look into it a bit later.

>I think we should add `#![deny(unsafe_op_in_unsafe_fn)]` so that we can better see which operations are unsafe inside our (sometimes large) unsafe functions.

Personally, I am not a huge fan of `unsafe_op_in_unsafe_fn`. Most `unsafe` functions (in `getrandom` and my other crates) are quite straightforward and additional `unsafe` blocks would only add noise. Also adding it would require `#![allow(unknown_lints)]`, if we don't want warnings on older toolchains. We can revisit it after the lint will be enabled by default.

briansmith:
> ```rust
> pub unsafe fn getrandom_raw(dst: *mut u8, len: usize) -> Result<(), Error> { ... }
> ```
> It allows to fill uninitialized arrays with random data.

If the goal is to fill uninitialized arrays then the API can and should be should be using `&mut [MaybeUninit<u8>]` for the output parameter, shouldn't it? Then this function wouldn't need to be declared `unsafe` either.

briansmith:
> I don't think there is anything wrong with exposing `unsafe` API, if there are legitimate uses for it.

I think it would be better to remove it from the API, as in general we should prefer to expose safe APIs instead of unsafe ones. I don't think the `_raw` version is useful if we have the `_uninit` variant; if a caller only has a pointer and a length, the caller can construct the `MaybeUninit` slice from that, and then use the `_uninit` function.

briansmith:
@newpavlov If you'd like, I could share a PR that implements the `_uninit` version using the suggestions above that attempt to minimize the amount of `unsafe` used. LMK.

>  if a caller only has a pointer and a length, the caller can construct the MaybeUninit slice from that, and then use the _uninit function.

Maybe this is only true of the length is small enough to be a valid slice length. But, again, the current implementation is already making such assumptions.

notgull:
In my opinion, we should go for the best of both worlds here. Have a raw pointer API *and* a `MaybeUninit<u8>` API. The raw pointer API would serve the MSRV use-case, among other cases that need raw pointers. The `MaybeUninit` API could be enabled on a feature to preserve the MSRV and serve the other use case. The latter can be implemented as a thin wrapper around the former.

josephlr:
@briansmith @notgull I've opened #286 to figure out the consensus for increasing the MSRV of `getrandom 0.2`. I think raising to 1.36 wouldn't be much of a problem, but I want to be sure.

briansmith:
> In my opinion, we should go for the best of both worlds here. Have a raw pointer API _and_ a `MaybeUninit<u8>` API. The raw pointer API would serve the MSRV use-case, among other cases that need raw pointers. The `MaybeUninit` API could be enabled on a feature to preserve the MSRV and serve the other use case. The latter can be implemented as a thin wrapper around the former.

My position is based on the following guideline: It is important to minimize the amount of `unsafe` in the implementation of the crate, to better leverage the language to ensure the crate's implementation is memory-safe. It is good to minimize the amount of `unsafe` APIs exposed to encourage users to write memory-safe code. These ideas are a large part of what makes this whole endeavor of writing Rust libraries meaningful, IMO, and the principle of maximizing memory safety is much more important than maximizing backward compatibility with years-old versions of the Rust toolchain, IMO.

josephlr:
I agree with @briansmith here, I think if we make more use of `MaybeUninit` inside of the crate, we can make the implementation easier to follow and reduce the amount of `unsafe` code.

src/custom.rs
```diff
@@ -78,9 +78,9 @@ macro_rules! register_custom_getrandom {
     ($path:path) => {
         // We use an extern "C" function to get the guarantees of a stable ABI.
         #[no_mangle]
-        extern "C" fn __getrandom_custom(dest: *mut u8, len: usize) -> u32 {
+        extern "C" fn __getrandom_custom(dst: *mut u8, len: usize) -> u32 {
             let f: fn(&mut [u8]) -> Result<(), $crate::Error> = $path;
-            let slice = unsafe { ::core::slice::from_raw_parts_mut(dest, len) };
+            let slice = unsafe { ::core::slice::from_raw_parts_mut(dst, len) };
```
briansmith, line 83: I think this is incorrect and fixing it would break backward compatibility.

`getrandom_raw` can be called with a pointer to uninitialized memory. So, we cannot legally construct a `&mut [u8]` slice from that pointer since we're only allowed to construct such a slice if the pointer points to initialized memory.

The function `$path` has a parameter of type `&mut [u8]` but to support the `_raw` API and/or the `_uninit` API it must take `&mut [MaybeUninit<u8>]` or a raw pointer and length instead.

briansmith:
> @newpavlov If you'd like, I could share a PR that implements the _uninit version using the suggestions above that attempt to minimize the amount of unsafe used. LMK.

I put up a draft PR #291 that does that. Again, my main goal there is to minimize the amount of `unsafe` added and to make it clear that the new uses of `unsafe` are safe. I hope that if a `_raw` API is really needed then it could be implemented on top of the `_uninit` variant, instead of vice-versa.

newpavlov:
Closing it in favor of #291

Issue #267, opened 2022-06-20
Where to enable "js" for wasm-bindgen


marvellous987:
Hi, I am trying to compile my program into WebAssembly using wasm-bindgen, but I got the following errors:

```
error: the wasm32-unknown-unknown target is not supported by default, you may need to enable the "js" feature. For more information see: https://docs.rs/getrandom/#webassembly-support
   --> /Users/xxx/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.7/src/lib.rs:235:9
    |
235 | /         compile_error!("the wasm32-unknown-unknown target is not supported by \
236 | |                         default, you may need to enable the \"js\" feature. \
237 | |                         For more information see: \
238 | |                         https://docs.rs/getrandom/#webassembly-support");
    | |________________________________________________________________________^

error[E0433]: failed to resolve: use of undeclared crate or module `imp`
   --> /Users/xxx/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.7/src/lib.rs:262:5
    |
262 |     imp::getrandom_inner(dest)
    |     ^^^ use of undeclared crate or module `imp`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `getrandom` due to 2 previous errors
```

May I ask where I can enable the "js" feature, as shown in https://docs.rs/getrandom/latest/getrandom/#webassembly-support? Or are there any other solutions?

Thank you!

vdods:
You want to enable the "js" feature on the getrandom crate itself.  Even if your crate doesn't directly depend on it, you'll want to add a line like
`getrandom = { version = "0.2", features = ["js"] }`

josephlr:
@vdods is correct. We explain this in our docs for the `custom` feature, but we should also explain how to do it for the `js` feature, as that is more common.

newpavlov:
I think we close this issue considering #280 is merged.

lucidBrot:
Leaving this here for the next dummy who thought he applied both changes in the docs and goes searching for more magical incantations, but actually did these two changes wrong :)

* I depend on a crate that depends on `rand`  that depends on `getrandom`.   
* I am now trying to build for `wasm32-unknown-unknown` instead of my usual linux build
* Initially I ran into the error message that points to the docs. After [reading the docs](https://docs.rs/getrandom/latest/getrandom/#webassembly-support) I have  
  * modified my `Cargo.toml` to set the feature
[...]

PR #280, opened 2022-08-20
Add description about Cargo js feature for WebAssembly section

Kazhuu:
Documentation improvement regarding https://github.com/rust-random/getrandom/issues/267.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -102,16 +102,6 @@
 //! using `rdrand` and `js` Cargo features) continue using their normal
 //! implementations even if a function is registered.
 //!
-//! ### Indirect Dependencies
-//!
-//! If `getrandom` is not a direct dependency of your crate, you can still
-//! enable any of the above fallback behaviors by enabling the relevant
-//! feature in your root crate's `Cargo.toml`:
-//! ```toml
-//! [dependencies]
-//! getrandom = { version = "0.2", features = ["js"] }
-//! ```
-//!
 //! ## Early boot
 //!
 //! Sometimes, early in the boot process, the OS has not collected enough
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -82,7 +82,9 @@
 //! getrandom = { version = "0.2", features = ["js"] }
 //! ```
 //!
-//! This feature has no effect on targets other than `wasm32-unknown-unknown`.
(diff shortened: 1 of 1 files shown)


newpavlov:
You should add a clarification that the `js` feature SHOULD be enabled only in "application" crates. For library tests and benchmarks the feature should be enabled only for tests and benchmarks, not for the whole library.

Kazhuu:
Okay I updated the doc specifically stating what you mentioned. I'm quite new to Rust in general so had to do some digging with this. Is it possible to enable specific feature when the crate type is binary for example? If yes, then example of this could be added to the docs.

For one of my own projects I've enabled js feature like this:
```TOML
[dependencies.getrandom]
features = ["js"]
```

josephlr:
@Kazhuu I've made some updates to the wording here. Hopefully this will help emphasize that  this feature should generally be enabled on non-library crates (binary, benchmark, or test crates). I also removed the "Indirect Dependencies" Section as its now redundant.

I'll give @newpavlov a chance to review this. Note that the CI failure is caused by an unrelated nightly bug, see:
  - https://github.com/rust-osdev/uefi-rs/issues/494
  - https://github.com/rust-lang/rust/issues/101071
  - https://github.com/rust-lang/rust/pull/101088

josephlr (approved):

Kazhuu:
Thanks @josephlr! Now the wording is a lot better and conveys more clearly what is needed. Even the original text I wrote is not there anymore I learned from this experience :+1:

src/lib.rs
```diff
@@ -74,6 +74,20 @@
...
+//! [dependencies]
+//! getrandom = { version = "0.2", features = ["js"] }
+//! ```
+//!
+//! This can be done even if `getrandom` is not a direct dependency. Cargo
+//! allows crates to enable features for indirect dependencies.
+//!
+//! This feature should only be enabled for binary, test, or benchmark crates.
+//! Library crates should generally not enable this feature, leaving such a
+//! decision to *users* of their library.
```
newpavlov, line 89: It may be worth to also add that library crates should not introduce their own `js` features which simply enable `getrandom`'s one.

newpavlov (approved):

src/lib.rs
josephlr: Good idea, done!

PR #282, opened 2022-09-05
Support random file of AIX

ecnelises:


josephlr:
Could we add some IBM docs to justify the implementation choice? I think this is correct, but I've never worked with AIX

ecnelises:
The official doc can be found at https://www.ibm.com/docs/en/aix/7.2?topic=files-random-urandom-devices , [this discussion in GnuPG mailing list](https://lists.gnupg.org/pipermail/gnupg-devel/2003-April/019954.html) is also good reference.

In short, the mechanism of `/dev/urandom` (and `/dev/random`) on AIX is the same as [Linux](https://unix.stackexchange.com/questions/324209/when-to-use-dev-random-vs-dev-urandom): the former is more reliable (it never blocks).

josephlr:
@ecnelises I'm just wondering if we will have to `poll` `/dev/random` like we do on Linux. I don't think we have to, but I'd like to see what OpenSSL does.

ecnelises:
> @ecnelises I'm just wondering if we will have to `poll` `/dev/random` like we do on Linux. I don't think we have to, but I'd like to see what OpenSSL does.

According to [rand_unix.c](https://github.com/openssl/openssl/blob/master/providers/implementations/rands/seeding/rand_unix.c) in OpenSSL, only Linux defines `DEVRANDOM_WAIT`, and `/dev/urandom` is in the highest priority by default.

josephlr:
Can you add a row to the documentation in `lib.rs` about AIX (and link to the official IBM docs).

I think we should also add AIX to the Tier 3 section of our CI. That may require waiting on Rust Nightly to actually understand the target.

josephlr:
Note: we are blocked on https://github.com/rust-lang/rust/pull/102293 being merged.

ecnelises:
That PR has been merged :)

newpavlov (approved):

PR #283, opened 2022-09-12
Prevent false detection of Node.js in Electron

maccesch:


josephlr:
Ideally, we would avoid a bunch of platform specific workarounds. Would something like https://github.com/rust-random/getrandom/pull/284 fix your issue? Also, how did you test that this change fixed the problem.

maccesch:
I think #284 might be a way better solution for this. I only tested my solution that it works in Electron.

PR #284, opened 2022-09-13
Rework JS feature detection

### LINKED ISSUE #256: Javascript code does not work in ESM context
Currently wasm-pack has two common build options:

- `nodejs` produces a commonjs module for node
- `web` produces an es6 module for the web

I'd like to use the same wasm module (published in npm) isomorphically for both nodejs and the web. It almost works - but unfortunately getrandom (run from nodejs) doesn't work at all in an es6 context. It produces this code in a esm file:

```
    imports.wbg.__wbg_modulerequire_0a83c0c31d12d2c7 = function() { return handleError(function (arg0, arg1) {
        var ret = module.require(getStringFromWasm0(arg0, arg1));
        return addHeapObject(ret);
    }, arguments) };
```

... Which errors, because `module` is not defined in this context.

I'm not sure how getrandom should work around this. Any thoughts?

josephlr:
This PR tries to incorporate some of the suggestions from #256

The first commit changes our detection logic to _first_  look for the Web Cryptography API _before_ attempting to check for Node.js support. This allows Node.js ES6 module users to add a polyfill like:
```js
import {webcrypto} from 'crypto'
globalThis.crypto = webcrypto
```
as described in https://github.com/rust-random/getrandom/issues/256#issuecomment-1161028902

We also rename `BroswerCrypto` to `WebCrypto` and update some of the outdated error messages and package comments.

The Second commit adds better detection for _why_ we failed to call `module.require("crypto")`, returning a different error code if `module.require` is not a function,

I think this fixes #256 about as well as we can.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -54,7 +54,7 @@ impl Error {
     /// Calling Node.js function `crypto.randomFillSync` failed.
     pub const NODE_RANDOM_FILL_SYNC: Error = internal_error(13);
     /// Called from an ES module on Node.js. This is unsupported, see:
-    /// https://docs.rs/getrandom#nodejs-es6-module-support.
+    /// <https://docs.rs/getrandom#nodejs-es-module-support>.
     pub const NODE_ES_MODULE: Error = internal_error(14);
 
     /// Codes below this point represent OS Errors (i.e. positive i32 values).
@@ -173,7 +173,7 @@ fn internal_desc(error: Error) -> Option<&'static str> {
         Error::VXWORKS_RAND_SECURE => Some("randSecure: VxWorks RNG module is not initialized"),
         Error::NODE_CRYPTO => Some("Node.js crypto CommonJS module is unavailable"),
         Error::NODE_RANDOM_FILL_SYNC => Some("Calling Node.js API crypto.randomFillSync failed"),
-        Error::NODE_ES_MODULE => Some("Node.js ES modules are not directly supported, see https://docs.rs/getrandom#nodejs-es6-module-support"),
+        Error::NODE_ES_MODULE => Some("Node.js ES modules are not directly supported, see https://docs.rs/getrandom#nodejs-es-module-support"),
         _ => None,
     }
 }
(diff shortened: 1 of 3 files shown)


newpavlov (approved):
Looks fine to me, but note that I am not familiar enough with the JS targets to judge quality of this solution.

josephlr:
> Looks fine to me, but note that I am not familiar enough with the JS targets to judge quality of this solution.

I'm not so sure about the implementation quality myself, I did try changing around some of the names of functions to make sure errors occurred when they "should".

The main feedback I was looking to get was mostly on the documentation, and error messages.

maccesch (approved):
I'm not an expert for this either but from what I can see this looks like a nice solution.

josephg:
I left some comments on the [related issue](https://github.com/rust-random/getrandom/issues/256). In short, there's a lingering problem in wasm-pack's `bundler` mode which we could fix or ignore. Either way it looks good to me!

newpavlov:
@josephlr
>The main feedback I was looking to get was mostly on the documentation, and error messages.

Both look good to me.

maccesch:
Is this going to be merged soon? I could really use this in my project 🙂

josephlr:
> Is this going to be merged soon? I could really use this in my project slightly_smiling_face

Yes, I just need to run a few more checks that we are correctly handling errors when in a Node.js ES module.

josephlr:
I added https://github.com/rust-random/getrandom/pull/284/commits/e0c93b10d5fa11ad9983e7da3d2ca973f5554c48 which makes the error @josephg pointed out in https://github.com/rust-random/getrandom/issues/256#issuecomment-1246277724 much nicer. Now, if we try to use an ES module from Node.js, we get an error like:
```
panicked at 'getrandom failed: Error { internal_code: 2147483662, description: "Node.js ES modules are not directly supported, see https://docs.rs/getrandom#nodejs-es-module-support" }', src/lib.rs:9:31
```
Then if we apply the workaround in our docs, everything works!!

This should hopefully allow users to (a) figure out what's going wrong, and (b) figure out how to fix the error.

maccesch:
Awesome! Is this going to be published as 2.8.0 ?

josephlr:
> Awesome! Is this going to be published as 0.2.8 ?

See #294

PR #301, opened 2022-10-22
Fix Solaris/OpenBSD/Dragonfly build and re-enable CI

josephlr:
Fixes #216 

This also adds two minor CI improvements:
  - Do a full link on `freebsd`
  - Build, Link, but don't run on `illumos`
  - Use the stable toolchain for our Tier 2 Build-only targets
  - Build (via `build-std`) for Tier 3 targets: `openbsd`, `dragonfly`, `haiku`

### DIFF
--- a/src/openbsd.rs
+++ b/src/openbsd.rs
@@ -8,6 +8,7 @@
 
 //! Implementation for OpenBSD
 use crate::{util_libc::last_os_error, Error};
+use core::mem::MaybeUninit;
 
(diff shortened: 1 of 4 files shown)


src/openbsd.rs
```diff
@@ -8,6 +8,7 @@
 
 //! Implementation for OpenBSD
 use crate::{util_libc::last_os_error, Error};
+use core::mem::MaybeUninit;
```
newpavlov, line 11: Heh, good catch.

newpavlov (approved):

PR #304, opened 2022-10-22
Add additional, sharper tests for checking RNG quality

### LINKED ISSUE #290: All test cases use buffers that are multiples of the word size
Consider this change to `rdrand_exact()`:

```diff
@@ -91,7 +91,7 @@ unsafe fn rdrand_exact(dest: &mut [u8]) -> Result<(), Error> {
     let tail = chunks.into_remainder();
     let n = tail.len();
     if n > 0 {
-        tail.copy_from_slice(&rdrand()?[..n]);
+        unreachable!()
     }
     Ok(())
 }
```

I would expect at least one test to fail with this modification, but they all pass, because every test buffer is a multiple of the word size. We should have tests for this branch.

josephlr:
This PR adds some tests which make sure calls to getrandom (for both small and large buffers) "look" random.

While we could certainly add more complicated randomness tests, these simple tests are:
  - Very easy to understand
  - Don't require any external crates
  - Makes sure we aren't doing something obviously stupid like
    - forgetting [these lines](https://github.com/rust-random/getrandom/blob/bd0654fe70980583e51573e755bafa3b2f8342d9/src/rdrand.rs#L91-L95)
    - failing to initialize every other byte
    - initializing some significant fraction of bytes with a constant

As this tests all buffer sizes from 1 to 64, it also fixes #290.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -12,6 +12,18 @@ fn test_zero() {
     getrandom_impl(&mut [0u8; 0]).unwrap();
 }
 
+// Return the number of bits in which s1 and s2 differ
+#[cfg(not(feature = "custom"))]
+fn num_diff_bits(s1: &[u8], s2: &[u8]) -> usize {
+    assert_eq!(s1.len(), s2.len());
+    let mut n = 0;
+    for i in 0..s1.len() {
+        n += (s1[i] ^ s2[i]).count_ones() as usize;
+    }
+    n
+}
+
+// Tests the quality of calling getrandom on two large buffers
 #[test]
 #[cfg(not(feature = "custom"))]
 fn test_diff() {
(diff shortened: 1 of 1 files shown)


tests/common/mod.rs
```diff
@@ -12,6 +12,17 @@ fn test_zero() {
...
 }
 
+// Return the number of bits in which s1 and s2 differ
+fn num_diff_bits(s1: &[u8], s2: &[u8]) -> usize {
+    assert_eq!(s1.len(), s2.len());
+    let mut n = 0;
+    for i in 0..s1.len() {
+        n += (s1[i] ^ s2[i]).count_ones() as usize;
+    }
+    n
```
newpavlov, line 22: You could use a fancier approach here:
```rust
pub fn num_diff_bits(s1: &[u8], s2: &[u8]) -> u32 {
    assert_eq!(s1.len(), s2.len());
    s1.iter()
        .zip(s2.iter())
        .map(|(a, b)| (a ^ b).count_ones())
        .sum()
}

newpavlov (approved):

tests/common/mod.rs
josephlr: Nice, I'm going to keep it as returning `usize` though to reduce the casting boilerplate in the tests.

PR #306, opened 2022-10-23
Cleanup wasm32-wasi target

josephlr:
This change ensures that we only compile our WASI implementation for 32-bit targets. The interaction between the WASI proposal and the memory64 proposal is not yet clear, [wasmtime does not yet support]( https://github.com/bytecodealliance/wasmtime/issues/3594#issuecomment-992590383) using WASI with memory64, and many of the interfaces use 32-bit values for pointers.

This change also reduces the use of `unsafe` from the wasi implementation. As noted in #253, changes to `Errno` mean that we can't get the error message from the raw error code, but we can avoid using unsafe when converting this code to a NonZeroU32. This handling also makes WASI behave more like our other targets, which also manually check that errno is non-zero.

### DIFF
--- a/src/wasi.rs
+++ b/src/wasi.rs
@@ -8,12 +8,18 @@
 
 //! Implementation for WASI
 use crate::Error;
-use core::{mem::MaybeUninit, num::NonZeroU32};
-use wasi::wasi_snapshot_preview1::random_get;
+use core::{
+    mem::MaybeUninit,
+    num::{NonZeroU16, NonZeroU32},
+};
+use wasi::random_get;
 
 pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
-    match unsafe { random_get(dest.as_mut_ptr() as i32, dest.len() as i32) } {
-        0 => Ok(()),
-        err => Err(unsafe { NonZeroU32::new_unchecked(err as u32) }.into()),
-    }
+    unsafe { random_get(dest.as_mut_ptr() as *mut u8, dest.len()) }.map_err(|e| {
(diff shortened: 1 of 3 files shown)


briansmith (commented):
LGTM.

src/wasi.rs
```diff
@@ -8,12 +8,18 @@
...
 pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
-    match unsafe { random_get(dest.as_mut_ptr() as i32, dest.len() as i32) } {
-        0 => Ok(()),
-        err => Err(unsafe { NonZeroU32::new_unchecked(err as u32) }.into()),
-    }
+    unsafe { random_get(dest.as_mut_ptr() as *mut u8, dest.len()) }.map_err(|e| {
+        // The WASI errno will always be non-zero, but we check just in case.
+        match NonZeroU16::new(e.raw()) {
+            Some(r) => Error::from(NonZeroU32::from(r)),
+            None => Error::ERRNO_NOT_POSITIVE,
```
newpavlov, line 22: One minor problem with this code is that compiler will not be able to remove the `None` branch without LTO. [`random_get`](https://github.com/bytecodealliance/wasi/blob/main/src/lib_generated.rs#L2056) is not marked as `#[inline]` so compiler only knows that `wasi::Errno` has type `u16`. But I guess it's fine, since the error should not happen in the first place.

newpavlov (approved):

src/wasi.rs
josephlr: This is correct, but I think it's fine because failure is rare, and the branch will be extremely predictable.

PR #307, opened 2022-10-23
Use getentropy on Emscripten

### LINKED ISSUE #275: Use getentropy() libc function on Emscripten
[Release 2.0.5: 09/28/2020](https://emscripten.org/docs/introducing_emscripten/release_notes.html) of Emscripten introduced a `getentropy()` call that can be used instead of opening `/dev/urandom`. We should use this as the preferred (or sole) method to get random data, as it avoids needing to open a file and hold onto a file descriptor.

NOTE: this probably requires getting the emscipten CI working with a recent-ish version

josephlr:
Fixes #275, note that `libc::getentropy` isn't in the `libc` crate yet, so we have to manually declare the bindings.

The last version of the EMSDK that didn't have support for this was 2.0.4 released over two years ago, so I don't think we have to worry about users using a super old SDK version, especially considering this is an less-used target.

### DIFF
--- a/src/emscripten.rs
+++ b/src/emscripten.rs
@@ -1,0 +1,18 @@
+//! Implementation for Emscripten
+use crate::{util_libc::last_os_error, Error};
+use core::mem::MaybeUninit;
+
+// Not yet defined in libc crate.
+extern "C" {
+    fn getentropy(buffer: *mut libc::c_void, length: usize) -> libc::c_int;
+}
+
+pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
+    // Emscripten 2.0.5 added getentropy, so we can use it unconditionally.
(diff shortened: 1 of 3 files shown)


newpavlov (approved):
Maybe it's worth to submit a patch which adds `getentropy` to `libc`? It looks like a small enough change for them to accept. But for now, I am fine with the manual declaration.

josephlr:
> Maybe it's worth to submit a patch which adds `getentropy` to `libc`? It looks like a small enough change for them to accept. But for now, I am fine with the manual declaration.

Sounds reasonable. I'm working on a few patches to send their way to make things easier on us w.r.t. declaring our own bindings.

kleisauke:
FWIW, `getentropy` was added to `libc`  for Emscripten via PR https://github.com/rust-lang/libc/pull/3087.

PR #310, opened 2022-10-23
Solaris: consistantly use /dev/random source

josephlr:
On Solaris, we opt to use `/dev/random` instead of `/dev/urandom` due to reasons explained in the comments and [in this Solaris blog post](https://blogs.oracle.com/solaris/post/solaris-new-system-calls-getentropy2-and-getrandom2).

However, we haven't been making the same choice when getting randomness via the `getrandom(2)` function, as we just pass `0` for the flags. We [used to](https://github.com/rust-random/rand/pull/730/files#diff-694d4302a3ff2a976f2fbd34bc05ada22ee61a4e21d2d985beab27f7a809268fR151) always set `GRND_RANDOM`, but that was removed in the move from `OsRng` to this crate.

For context, https://github.com/rust-random/rand/pull/730, https://github.com/rust-random/getrandom/pull/9, and https://github.com/rust-random/getrandom/pull/51 are the major changes to the Solaris/Illumos implementation over the years.

See the solaris documentation for:
- [`getrandom(2)`](https://docs.oracle.com/cd/E88353_01/html/E37841/getrandom-2.html)
- [`urandom(4)`](https://docs.oracle.com/cd/E88353_01/html/E37851/urandom-4d.html)

I also updated the doucmentation to better reflect when [Illumos added the `getrandom(2)` function](https://www.illumos.org/issues/9971#change-23483).

Finally, a question.  https://github.com/rust-random/getrandom/pull/51 removed chunking for the fallback implementation that reads from `/dev/random`, should we add it back?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ compiler_builtins = { version = "0.1", optional = true }
 core = { version = "1.0", optional = true, package = "rustc-std-workspace-core" }
 
 [target.'cfg(unix)'.dependencies]
-libc = { version = "0.2.120", default-features = false }
+libc = { version = "0.2.128", default-features = false }
 
 [target.'cfg(target_os = "wasi")'.dependencies]
 wasi = "0.11"
--- a/src/solaris_illumos.rs
+++ b/src/solaris_illumos.rs
@@ -8,12 +8,11 @@
 
 //! Implementation for the Solaris family
 //!
-//! Read from `/dev/random`, with chunks of limited size (256 bytes).
 //! `/dev/random` uses the Hash_DRBG with SHA512 algorithm from NIST SP 800-90A.
 //! `/dev/urandom` uses the FIPS 186-2 algorithm, which is considered less
-//! secure. We choose to read from `/dev/random`.
+//! secure. We choose to read from `/dev/random` (and use GRND_RANDOM).
 //!
-//! Since Solaris 11.3 and mid-2015 illumos, the `getrandom` syscall is available.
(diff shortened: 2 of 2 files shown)


josephlr:
As a data point, @briansmith's _ring_ [reads from `/dev/urandom` without chunking](https://github.com/briansmith/ring/blob/0d18da865afee5698590d566df8606d29d51d842/src/rand.rs#L343).

newpavlov (approved):

newpavlov:
>Finally, a question. https://github.com/rust-random/getrandom/pull/51 removed chunking for the fallback implementation that reads from /dev/random, should we add it back?

Isn't "chunking" handled automatically by `sys_fill_exact`? We have the exception for Emscripten, but I don't think we need a similar code for Solaris.

josephlr:
> > Finally, a question. #51 removed chunking for the fallback implementation that reads from /dev/random, should we add it back?
> 
> Isn't "chunking" handled automatically by `sys_fill_exact`? We have the exception for Emscripten, but I don't think we need a similar code for Solaris.

You're correct. Looking at the Solaris documentation again, it seems like `read()`s from /dev/random and calls to `getrandom()` behave differently when they are passed buffers which are "too long". The `read()` call will [succeed and just return fewer bytes](https://github.com/kofemann/opensolaris/blob/80192cd83bf665e708269dae856f9145f7190f74/usr/src/uts/common/io/random.c#L262) (so using `sys_fill_exact` is fine), while calls to `getrandom()` will just fail with `EINVAL`.

So the implementation in this PR seems correct.

PR #313, opened 2022-10-23
Update to 2021 edition

josephlr:
No changes to the code were needed.


josephlr:
Whoops, we obviously can't do this and have the MSRV be 1.36

PR #315, opened 2022-10-23
Remove buffer zeroing from Node.js implementation

josephlr:
We can use [`Uint8Array::view_mut_raw`](https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.Uint8Array.html#method.view_mut_raw) which was [added](https://github.com/rustwasm/wasm-bindgen/pull/1850) in version `0.2.54` of `wasm-bindgen`. This method was introduced to deal with uninitialized memory, see the `Safety` comment for more info.

### DIFF
--- a/src/js.rs
+++ b/src/js.rs
@@ -5,7 +5,7 @@
 // <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
-use crate::{util::uninit_slice_fill_zero, Error};
+use crate::Error;
 
 extern crate std;
 use std::{mem::MaybeUninit, thread_local};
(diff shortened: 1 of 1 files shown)


src/js.rs
```diff
@@ -36,12 +36,19 @@ pub(crate) fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error>
...
+                    // and crypto.randomFillSync does not resize the buffer. We
+                    // know the length is less than u32::MAX because of the
+                    // chunking above.
+                    // Note that this uses the fact that JavaScript doesn't
+                    // have a notion of "uninitialized memory", this is purely
+                    // a Rust/C/C++ concept.
+                    let buf = unsafe {
+                        Uint8Array::view_mut_raw(chunk.as_mut_ptr() as *mut u8, chunk.len())
+                    };
+                    if n.random_fill_sync(buf).is_err() {
```
newpavlov, line 51: Shouldn't you extend the `unsafe` block to include `random_fill_sync`? I don't it's quite legal to have `buf` around in safe code since using it we can look into uninitialized memory. I mean something like this:
```rust
let res = unsafe {
    let buf = Uint8Array::view_mut_raw(chunk.as_mut_ptr() as *mut u8, chunk.len());
    n.random_fill_sync(buf)
};
if res.is_err() { .. }

File: src/js.rs [hunk 2]
```diff
@@ -5,7 +5,7 @@
 // <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
-use crate::{util::uninit_slice_fill_zero, Error};
```
newpavlov, line 8: After this change, the only other place which uses `uninit_slice_fill_zero` is the custom backend. It could be worth to inline this function there.

newpavlov (approved):
Good find!

src/js.rs
josephlr: Makes sense to me. Do you think that:
```rust
let res = n.random_fill_sync(unsafe {
    Uint8Array::view_mut_raw(chunk.as_mut_ptr() as *mut u8, chunk.len())
});
```
would be ok?

src/js.rs
josephlr: Ya I was going to clean this up when I cleaned up the `custom` implementation.

src/js.rs
newpavlov: Yeah, it should be fine.

PR #317, opened 2022-10-24
Add tvOS and watchOS support, identical to iOS support

thomcc:
This adds tvOS and watchOS support. Support in the Rust stdlib in watchOS is [recent](https://github.com/rust-lang/rust/pull/98101), and tvOS is [in progress](https://github.com/rust-lang/rust/pull/103503).

These OSes all can use the same interface from Security.framework. Ideally they'd use `getrandom`, but [sadly](https://github.com/rust-lang/rust/issues/102643) they currently cannot. I've renamed `ios.rs` to `apple-other.rs` as a result of this, but there are no other changes.


josephlr:
Given that these targets are not super well supported currently, it's fine that we don't run the tests in CI. However, we should do something similar to our build and link tests for Apple Arm64. You should just be able to add the targets to the tests.yml

thomcc:
> Given that these targets are not super well supported currently, it's fine that we don't run the tests in CI. However, we should do something similar to our build and link tests for Apple Arm64. You should just be able to add the targets to the tests.yml

Hm, `cargo test --no-run --target=arm64_32-apple-watchos -Zbuild-std --features=std` fails with a pretty obscure linker error. I guess we don't pass things to the linker correctly for arm64_32 (I mostly am making this PR for tvOS, but figured I'd add watchOS while I was here).

For tvos it's not currently supported in libstd, so the best we could do is `cargo check`. I'll follow up here later.

josephlr:
> Hm, `cargo test --no-run --target=arm64_32-apple-watchos -Zbuild-std --features=std` fails with a pretty obscure linker error. I guess we don't pass things to the linker correctly for arm64_32.

Does this issue occur of the other watchOS targets (i.e. `{aarch64,x86_64}-apple-watchos-sim` or `armv7k-apple-watchos`)? We really only need to check one of them in our CI. I would check myself, but I don't have a mac.

If you can get one of them working, add it [here]().

> (I mostly am making this PR for tvOS, but figured I'd add watchOS while I was here).

Makes sense if linking doesn't work, just add a comment and put the watchOS target with the tvOS target for now.

> For tvos it's not currently supported in libstd, so the best we could do is `cargo check`. I'll follow up here later.

This makes sense, can you add `aarch64-apple-tvos` to the list of `no_std` targets [here](https://github.com/rust-random/getrandom/blob/2aabb63d2c0be1507dd6ebae74007ee641ba8fa8/.github/workflows/tests.yml#L270).

thomcc:
> Does this issue occur of the other watchOS targets (i.e. `{aarch64,x86_64}-apple-watchos-sim` or `armv7k-apple-watchos`)? We really only need to check one of them in our CI. I would check myself, but I don't have a mac.

Good call, the simulator targets work, which actually helps point out what this issue probably is. Thanks.

Some notes:

- The watchos-sim target is not distributed via rustup, so I had to just -Zbuild-std it. Let me know if you want me to move it elsewhere or something.
- I did have to bump your libc version to 0.2.136 for tvOS, to get https://github.com/rust-lang/libc/pull/2958.

.github/workflows/tests.yml
```diff
@@ -287,6 +289,10 @@ jobs:
             features: ["rdrand"]
           - target: x86_64-unknown-l4re-uclibc
             features: ["rdrand"]
+          # `std` support still in progress. Can be moved up with the other
+          # apple targets after https://github.com/rust-lang/rust/pull/103503
+          - target: aarch64-apple-tvos
```
josephlr, line 294: Nit: how this rule works is a little confusing, but you don't need to have it under `include:`. No features is the default for the targets in the matrix.

If you move this comment to above where you have `aarch64-apple-tvos` in the target list, that should be fine.

.github/workflows/tests.yml
```diff
@@ -158,6 +158,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo test --no-run --target=aarch64-apple-darwin --features=std
       - run: cargo test --no-run --target=aarch64-apple-ios --features=std
+      - run: cargo test --no-run --target=aarch64-apple-watchos-sim -Zbuild-std --features=std
```
josephlr, line 161: You need to update the `dtolnay/rust-toolchain@stable` to `dtolnay/rust-toolchain@nightly` and add `components: rust-src` to the `with:`, so something like:
```yaml
      - uses: dtolnay/rust-toolchain@nightly
        with:
          targets: aarch64-apple-darwin, aarch64-apple-ios
          components: rust-src
```

josephlr (commented):
Other than some nits for the CI, this looks reasonable.

Bumping the `libc` version is fine.

thomcc:
Thanks, is this more what you meant?

thomcc:
Ah, -Zbuild-std requires nightly, do you want me to move it to another job, or would you rather make that one use nightly?

josephlr:
> requires nightly, do you want me to move it to another job, or would you r

Having that one use nightly is fine, ideally it would be in its own job, but macOS runners are slow and constrained on Github Actions, so reducing the jobs generally makes things better.

thomcc:
I've updated that job to nightly.

josephlr (approved):

PR #318, opened 2022-10-25
windows: move BCryptGenRandom to helper function

josephlr:
This _is not_ a functional change. It just makes adding additional fallback implementations easier in the future.

Note that the try_into().unwrap() is optimized away: https://godbolt.org/z/nbThsqsTE

I was originally going to try to use `BCRYPT_RNG_ALG_HANDLE` and fall back to our currently implementation, but that will cause segfaults on older versions.

### DIFF
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -7,7 +7,7 @@
 // except according to those terms.
 
 use crate::Error;
-use core::{ffi::c_void, mem::MaybeUninit, num::NonZeroU32, ptr};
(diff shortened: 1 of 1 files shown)


newpavlov:
I think it will be better to have one PR with fallback implementation, instead of splitting it into several.

josephlr:
> I think it will be better to have one PR with fallback implementation, instead of splitting it into several.

Sounds reasonable, I'll move this to draft then.

josephlr:
Closing, we went with a different approach in #337

PR #326, opened 2022-12-29
Do not use open64

kraj:
glibc is providing open64 and other lfs64 functions but musl aliases them to normal equivalents since off_t is always 64-bit on musl, therefore check for target env along when target OS is linux before using open64, this is more available. Latest Musl has made these namespace changes [1]

[1] https://git.musl-libc.org/cgit/musl/commit/?id=246f1c811448f37a44b41cd8df8d0ef9736d95f4

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/util_libc.rs
+++ b/src/util_libc.rs
@@ -140,19 +140,11 @@ impl Weak {
     }
 }
 
-cfg_if! {
-    if #[cfg(any(target_os = "linux", target_os = "emscripten"))] {
-        use libc::open64 as open;
-    } else {
-        use libc::open;
-    }
-}
-
 // SAFETY: path must be null terminated, FD must be manually closed.
 pub unsafe fn open_readonly(path: &str) -> Result<libc::c_int, Error> {
     debug_assert_eq!(path.as_bytes().last(), Some(&0));
     loop {
-        let fd = open(path.as_ptr() as *const _, libc::O_RDONLY | libc::O_CLOEXEC);
+        let fd = libc::open(path.as_ptr() as *const _, libc::O_RDONLY | libc::O_CLOEXEC);
         if fd >= 0 {
             return Ok(fd);
         }


briansmith:
Why do we ever need to use open64 in the first place? Since we never seek it seems like we should just avoid the conditional logic.

kraj:
> Why do we ever need to use open64 in the first place? Since we never seek it seems like we should just avoid the conditional logic.

I am fine by using `open`, I was trying to fix just musl case, but if we think its better to use just `open` I can update the patch

newpavlov:
>Why do we ever need to use open64 in the first place?

It's likely we have inherited it from `std`, but I agree that we probably do not need it. If @josephlr agrees, we should simply use `open`.

briansmith:
What does libstd do for musl targets?

kraj:
> What does libstd do for musl targets?

The change to drop/deprecate LFS64 functions in musl is recent and has not yet made it to a GA release.
I will be fixing it in `libstd-rs` as well see https://github.com/rust-lang/rust/pull/106246

josephlr:
> > Why do we ever need to use open64 in the first place?
> 
> It's likely we have inherited it from `std`, but I agree that we probably do not need it. If @josephlr agrees, we should simply use `open`.

Looking through the various documentation, the only difference (if any) is the presence of `O_LARGEFILE`. That flag isn't relevant to us, as we are only opening a device (not a real file), so we never care about overflowing `off_t`.

Let's change this patch to just unconditionally use `open`.

josephlr:
If you rebase this PR onto the latest master, the tests should be passing.

kraj:
> > > Why do we ever need to use open64 in the first place?
> > 
> > 
> > It's likely we have inherited it from `std`, but I agree that we probably do not need it. If @josephlr agrees, we should simply use `open`.
> 
> Looking through the various documentation, the only difference (if any) is the presence of `O_LARGEFILE`. That flag isn't relevant to us, as we are only opening a device (not a real file), so we never care about overflowing `off_t`.
> 
> Let's change this patch to just unconditionally use `open`.

done

kraj:
> If you rebase this PR onto the latest master, the tests should be passing.

cool. Uploaded a new version of PR. Lets see

src/util_libc.rs
```diff
@@ -141,11 +141,7 @@ impl Weak {
 }
 
 cfg_if! {
-    if #[cfg(any(target_os = "linux", target_os = "emscripten"))] {
-        use libc::open64 as open;
-    } else {
-        use libc::open;
-    }
+    use libc::open;
```
newpavlov, line 144: You can simply remove `cfg_if` and use `libc::open` at call sites (I think we have only one such place).

newpavlov (approved):

PR #328, opened 2023-01-04
Disable cross build/link tests

newpavlov:


josephlr (approved):
Good catch, I think this is a cashing issue with the cross build stuff, the solution is probably just to not cache the build when doing cross linking.

Issue #314, opened 2022-10-23
Windows: should we fall back to RtlGenRandom if BCryptGenRandom fails


josephlr:
In #65 we decided to use `BCryptGenRandom` instead of `RtlGenRandom`. I still think this is the correct decision.

However, Rust has seen issues (https://github.com/rust-lang/rust/issues/94098 and https://github.com/rust-lang/rust/issues/99341) around such use. To fix these issues they decided to use `RtlGenRandom` again, but only as a fallback mechanism (https://github.com/rust-lang/rust/pull/96917). 

Should we do the same?

newpavlov:
Unfortunately, it does not look like anyone has uncovered root of the issue. So the most practical choice (though I personally dislike it) will be to follow `std` and introduce the `BCryptGenRandom` fallback, but with the clear indication that it's a hack for the OS issue.

briansmith:
* The Rust stdlib uses the RNG just for the HashMap anti-DoS mechanism, right? That requires a lower level of security than what other uses of `getrandom` require.
* I think `RtlGenRandom` isn't supported on ARM and the problems Firefox reports with `BCryptGenRandom` probably don't apply to ARM Windows machines. So perhaps only do the fallback on x86/x86_64.
* https://bugs.chromium.org/p/boringssl/issues/detail?id=307 has more discussion, including some good notes.
* Chromium also uses `RtlGenRandom`, according to https://source.chromium.org/search?q=RtlGenRandom&sq=.

newpavlov:
>The Rust stdlib uses the RNG just for the HashMap anti-DoS mechanism, right? That requires a lower level of security than what other uses of getrandom require.

AFAIK there are no security issues with `RtlGenRandom`. The reason why we avoid it is because it was deprecated by Microsoft and not present on newer targets (such as UWP and ARM).

josephlr:
> > The Rust stdlib uses the RNG just for the HashMap anti-DoS mechanism, right? That requires a lower level of security than what other uses of getrandom require.
> 
> AFAIK there are no security issues with `RtlGenRandom`. The reason why we avoid it is because it was deprecated by Microsoft and not present on newer targets (such as UWP and ARM).

Given `RtlGenRandom` is still used in security critical software (Chrome, Edge, etc...) I'm not concerned with the security implications of falling back to this method. I think the main point here is about usability and stability. Given that I think our default should be to:
  - Always try `BCryptGenRandom`
[...]

PR #330, opened 2023-01-11
Make the RNG fall back to using an algorithm handle if BCryptGenRandom fails

badboy:
Based on https://github.com/rust-lang/rust/pull/102044

---

Adresses #314, replaces #318 

This doesn't have a fast-path to avoid `BCryptGenRandom` if it is known to fail.
Thus this is probably gonna get expensive for some users, I can add some fast-path though.
Right now I don't have a Windows machine to actually test it on (but CI should do that, I also have acess to a machine again next week).
I'm not 100% sure yet that it fixes the crashes we see on Firefox, but given the stdlib switched to a similar solution I have hopes.

### DIFF
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -5,45 +5,144 @@
 // <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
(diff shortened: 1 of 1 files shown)


josephlr:
Closing, we went with a different approach in #337

PR #331, opened 2023-01-25
Add support for getrandom syscall on NetBSD ≥ 10.0

tklauser:
NetBSD supports the getrandom system call since version 10.0 [1]. Use it if available, otherwise fall back to kern.arandom like on FreeBSD.

[1] https://man.netbsd.org/getrandom.2


josephlr (approved):

src/lib.rs
```diff
@@ -18,7 +18,7 @@
 //! | iOS, tvOS, watchOS | `*‑apple‑ios`, `*-apple-tvos`, `*-apple-watchos` | [`SecRandomCopyBytes`]
 //! | FreeBSD           | `*‑freebsd`        | [`getrandom`][5] if available, otherwise [`kern.arandom`][6]
 //! | OpenBSD           | `*‑openbsd`        | [`getentropy`][7]
-//! | NetBSD            | `*‑netbsd`         | [`kern.arandom`][8]
+//! | NetBSD            | `*‑netbsd`         | [`getrandom`][16] if available, [`kern.arandom`][8]
```
newpavlov, line 21: Add "otherwise" after the comma.

PR #333, opened 2023-01-26
Hermit: use sys_read_entropy syscall

newpavlov:
The syscall was implemented in https://github.com/hermitcore/libhermit-rs/pull/640

On non-x86 targets it will return `-ENOSYS` (which, frankly, IMO is quite bad), but I would say it's outside of our area of responsibility.

cc @joboet @stlankes

### DIFF
--- a/src/hermit.rs
+++ b/src/hermit.rs
@@ -1,0 +1,21 @@
+use crate::Error;
+use core::{cmp::min, mem::MaybeUninit, num::NonZeroU32};
+
+extern "C" {
+    fn sys_read_entropy(buffer: *mut u8, length: usize, flags: u32) -> isize;
+}
+
+pub fn getrandom_inner(mut dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
+    while !dest.is_empty() {
+        let res = unsafe { sys_read_entropy(dest.as_mut_ptr() as *mut u8, dest.len(), 0) };
+        if res < 0 {
+            // SAFETY: all Hermit error codes use i32 under the hood:
+            // https://github.com/hermitcore/libhermit-rs/blob/master/src/errno.rs
+            let code = unsafe { NonZeroU32::new_unchecked((-res) as u32) };
+            return Err(code.into());
+        }
+        let len = min(res as usize, dest.len());
+        dest = &mut dest[len..];
+    }
+    Ok(())
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,7 +24,7 @@
 //! | Fuchsia OS        | `*‑fuchsia`        | [`cprng_draw`]
 //! | Redox             | `*‑redox`          | `/dev/urandom`
 //! | Haiku             | `*‑haiku`          | `/dev/urandom` (identical to `/dev/random`)
-//! | Hermit            | `*-hermit`         | `sys_read_entropy`
+//! | Hermit            | `*-hermit`         | [`sys_read_entropy`]
 //! | SGX               | `x86_64‑*‑sgx`     | [`RDRAND`]
 //! | VxWorks           | `*‑wrs‑vxworks‑*`  | `randABytes` after checking entropy pool initialization with `randSecure`
 //! | ESP-IDF           | `*‑espidf`         | [`esp_fill_random`]
@@ -179,6 +179,7 @@
 //! [`module`]: https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-js-imports/module.html
 //! [CommonJS modules]: https://nodejs.org/api/modules.html
 //! [ES modules]: https://nodejs.org/api/esm.html
+//! [`sys_read_entropy`]: https://hermitcore.github.io/libhermit-rs/hermit/fn.sys_read_entropy.html
 
 #![doc(
     html_logo_url = "https://www.rust-lang.org/logos/rust-logo-128x128-blk.png",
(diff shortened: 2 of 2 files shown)


josephlr:
I agree that handing which platform Hermit supports is not our job, so this implementation should be fine.

Could we add a documentation link?

newpavlov:
I couldn't find a documentation to which we could link, so the only option for now is to link the source code. I am not sure such link would be useful.

josephlr:
> I couldn't find a documentation to which we could link, so the only option for now is to link the source code. I am not sure such link would be useful.

I think linking to the system crate (which is the preferred interface to Hermit) would be appropriate: https://hermitcore.github.io/libhermit-rs/hermit/fn.sys_read_entropy.html

newpavlov:
Ah, good find!

src/hermit.rs
```diff
@@ -0,0 +1,21 @@
...
+    fn sys_read_entropy(buffer: *mut u8, length: usize, flags: u32) -> isize;
+}
+
+pub fn getrandom_inner(mut dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
+    while !dest.is_empty() {
+        let res = unsafe { sys_read_entropy(dest.as_mut_ptr() as *mut u8, dest.len(), 0) };
+        if res < 0 {
+            // SAFETY: all Hermit error codes use i32 under the hood:
+            // https://github.com/hermitcore/libhermit-rs/blob/master/src/errno.rs
+            let code = unsafe { NonZeroU32::new_unchecked((-res) as u32) };
```
josephlr, line 14: For this, could we avoid unsafe code by doing:
```rust
if res < 0 {
    return Err(match NonZeroU32::new((-res) as u32) {
        Some(code) => code.into(),
        None => Error::ERRNO_NOT_POSITIVE,
    });
}
```
which is sort of similar to code we use elsewhere, and avoids and unsafe or panicking code.

josephlr (approved):

src/hermit.rs
```diff
@@ -0,0 +1,21 @@
...
+            return Err(code.into());
+        }
+        let len = min(res as usize, dest.len());
+        dest = &mut dest[len..];
```
joboet, line 18: The call returns the number of bytes written when successful, which should never be larger than the length of the buffer passed in.
```suggestion
        dest = &mut dest[res as usize..];
```

src/hermit.rs
```diff
@@ -0,0 +1,21 @@
+use crate::Error;
+use core::{cmp::min, mem::MaybeUninit, num::NonZeroU32};
+
+extern "C" {
+    fn sys_read_entropy(buffer: *mut u8, length: usize, flags: u32) -> isize;
+}
```
joboet, line 6: The newest version of [`hermit-abi`](https://crates.io/crates/hermit-abi) (docs are currently broken) already includes this definition.

joboet (approved):

src/hermit.rs
newpavlov: See #334. Compiler does not know this semantic information and thus generates unreachable panic branch.

src/hermit.rs
newpavlov: In such simple cases we prefer to link directly instead of adding a dependency on a "system" crate.

src/hermit.rs
newpavlov: I think `res < 0` followed by `ERRNO_NOT_POSITIVE` would be... a bit confusing. The reason why we can't simply `.unwrap()` here is because the syscall returns `isize`, thus on 64-bit platforms it's theoretically  possible to get a negative number with lower 32 bits equal to zero.

josephlr: That's very fair. Maybe we can just have an unwrap then instead of an unsafe block. The panic branch would only be in the error handling, so it doesn't seem like an issue.

newpavlov: I think that using `ERRNO_NOT_POSITIVE` as an unreachable fallback constant is better than introduction of an unreachable panic branch. Maybe we should introduce `UNSPECIFIED_ERROR` as `internal_error(2)`?

josephlr: What if we just added the following method to `Error`:
```rust
pub fn from_raw_os_error(errno: i32) -> Self {
    if errno > 0 {
        Error::from(NonZeroU32::new(errno as u32).unwrap())
    } else {
        Error::ERRNO_NOT_POSITIVE
    }
}
```

And in `hermit.rs` we just had:
```
if res < 0 {
    return Err(Error::from_raw_os_error(-res as i32));
}
```

josephlr: Actually, I'm fine leaving this as-is for now. I've got a followup PR that might make this sort of thing easier going forward.

hugusmaximus:
> > I couldn't find a documentation to which we could link, so the only option for now is to link the source code. I am not sure such link would be useful.
> 
> I think linking to the system crate (which is the preferred interface to Hermit) would be appropriate: https://hermitcore.github.io/libhermit-rs/hermit/fn.sys_read_entropy.html

This link is broken...

newpavlov:
@hugusmaximus 
The [documentation](https://docs.rs/getrandom) currently uses this [link](https://github.com/hermit-os/kernel/blob/315f58ff5efc81d9bf0618af85a59963ff55f8b1/src/syscalls/entropy.rs#L47-L55).

hugusmaximus:
Yes, I found this info too... I was s bit confused. Thanks.

PR #335, opened 2023-02-02
rdrand: Remove checking for 0 and !0 and instead check CPU family and do a self-test

### LINKED ISSUE #228: RDRAND-based output is (too) biased
See https://github.com/rust-random/getrandom/blob/30308ae845b0bf3839e5a92120559eaf56048c28/src/rdrand.rs#L33-L42, merged from PR #48:
```
 if rdrand_step(&mut el) == 1 {
            // AMD CPUs from families 14h to 16h (pre Ryzen) sometimes fail to
            // set CF on bogus random data, so we check these values explicitly.
            // See https://github.com/systemd/systemd/issues/11810#issuecomment-489727505
            // We perform this check regardless of target to guard against
            // any implementation that incorrectly fails to set CF.
            if el != 0 && el != !0 {
                return Ok(el.to_ne_bytes());
            }
            // Keep looping in case this was a false positive.
        }
```

The condition `if el != 0 && el != !0` is testing that the value returned by RDRAND, after it reports success, is not zero or all-one bits, i.e. never `usize::MIN` or `usize::MAX`. Consequently, `getrandom::getrandom` will never return a result where there a single word is zero or all-one bits, where as word is a 4-byte chunk on 32-bit x86 or a 8-byte chunk on 64-bit x86, when the RDRAND implementation is being used. Such values are expected to occur every 2/2^N words on average. As a result, any use of `getrandom::getrandom` returns results that are wrongly biased; 2/2^N values are rejected on an N-bit platform. 

32-bit x86 support for the RDRAND feature was added in PR #134, after PR #48. The analysis on the bias when the code was implemented was based on the probability 2/2^64 which is correct when this code runs on a 64-bit CPU, but not for when it runs on a 32-bit CPU. I suspect that when PR #134 was under consideration, the difference in the bias was perhaps overlooked. The bias is notably worse on 32-bit platforms since 2/2^32 is much more likely than 2/2^64.

Outside of cryptography, I find the present solution particularly unfortunate because if getrandom's output is used for randomized testing of a 64-bit/32-bit function on x86_64/x86 on a target for which the RDRAND implementation is used, the important boundary conditions of the values `usize::MIN` and `usize::MAX` will never be reached, ever!

In the context of implementing cryptographic functions, and especially keygen and nonce generation, I suspect any user of this crate is likely to get pestered about this bias and will have to address it in some way.

In any case, I think the present solution is weird enough that it is worth having more eyes on it and more discussion, and also I hope we could find a better solution.

Incidentally, BoringSSL's code has this to say:
```
      // Also disable for family 0x17, models 0x70–0x7f, due to possible RDRAND
      // failures there too.
```
It seems like Google has reason to believe that some family 0x17 models may also be bad. So the comment indicating the problem is only with families 14h-16h may be worth updating, at least.

josephlr:
Fixes #228 

This is essentially the plan discussed in https://github.com/rust-random/getrandom/issues/228#issuecomment-1407391965

I changed the `rdrand` function to return `Option<u64>` as that makes reusing it in the self-test much nicer.

The main complexity here comes from:
  - Checking if the CPU is AMD with family < 0x17
  - Performing the self test. See the [Linux implementation](https://github.com/torvalds/linux/blob/9f266ccaa2f5228bfe67ad58a94ca4e0109b954a/arch/x86/kernel/cpu/rdrand.c)

There are also some minor commits which:
  - Refer to the the correct flag name now that [`+rdrand` is supported by rustc](https://github.com/rust-lang/rust/pull/84991)
  - Update old comments

### DIFF
--- a/src/rdrand.rs
+++ b/src/rdrand.rs
@@ -45,10 +45,10 @@ unsafe fn rdrand() -> Result<[u8; WORD_SIZE], Error> {
     Err(Error::FAILED_RDRAND)
 }
 
-// "rdrand" target feature requires "+rdrnd" flag, see https://github.com/rust-lang/rust/issues/49653.
+// "rdrand" target feature requires "+rdrand" flag, see https://github.com/rust-lang/rust/issues/49653.
 #[cfg(all(target_env = "sgx", not(target_feature = "rdrand")))]
 compile_error!(
-    "SGX targets require 'rdrand' target feature. Enable by using -C target-feature=+rdrnd."
+    "SGX targets require 'rdrand' target feature. Enable by using -C target-feature=+rdrand."
 );
 
 #[cfg(target_feature = "rdrand")]
--- a/src/rdrand.rs
+++ b/src/rdrand.rs
@@ -50,8 +50,8 @@ compile_error!(
 // Fails with probability < 2^(-90) on 32-bit systems
 #[target_feature(enable = "rdrand")]
 unsafe fn self_test() -> bool {
-    // On AMD, RDRAND returns usize::MAX on failure, count it as a collision.
-    let mut prev = usize::MAX;
+    // On AMD, RDRAND returns 0xFF...FF on failure, count it as a collision.
+    let mut prev = !0; // TODO(MSRV 1.43): Move to usize::MAX
     let mut fails = 0;
     for _ in 0..8 {
         match rdrand() {
--- a/src/rdrand.rs
+++ b/src/rdrand.rs
@@ -5,8 +5,6 @@
 // <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
-
-//! Implementation for SGX using RDRAND instruction
 use crate::{
     util::{slice_as_uninit, LazyBool},
     Error,
--- a/src/rdrand.rs
+++ b/src/rdrand.rs
@@ -7,8 +7,11 @@
 // except according to those terms.
 
 //! Implementation for SGX using RDRAND instruction
-use crate::{util::slice_as_uninit, Error};
-use core::mem::{self, MaybeUninit};
+use crate::{
+    util::{slice_as_uninit, LazyBool},
+    Error,
+};
+use core::mem::{size_of, MaybeUninit};
 
 cfg_if! {
     if #[cfg(target_arch = "x86_64")] {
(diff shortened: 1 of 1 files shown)


newpavlov (approved):

src/rdrand.rs
```diff
@@ -24,74 +27,105 @@ cfg_if! {
...
-#[cfg(target_feature = "rdrand")]
-fn is_rdrand_supported() -> bool {
-    true
+// Run a small self-test to make sure we aren't repeating values
+// Adapted from Linux's test in arch/x86/kernel/cpu/rdrand.c
+// Fails with probability < 2^(-90) on 32-bit systems
+#[target_feature(enable = "rdrand")]
+unsafe fn self_test() -> bool {
+    // On AMD, RDRAND returns usize::MAX on failure, count it as a collision.
+    let mut prev = usize::MAX;
```
newpavlov, line 54: As CI shows, this does not work on 1.36, you have to use `core::usize::MAX`.

src/rdrand.rs
```diff
@@ -24,74 +27,105 @@ cfg_if! {
...
+        // rdrand is not undefined behavior.
+        let src = unsafe { rdrand() }?.to_ne_bytes();
+        chunk.copy_from_slice(slice_as_uninit(&src));
     }
 
     let tail = chunks.into_remainder();
     let n = tail.len();
     if n > 0 {
-        tail.copy_from_slice(slice_as_uninit(&rdrand()?[..n]));
+        let src = unsafe { rdrand() }?.to_ne_bytes();
```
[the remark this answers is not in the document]
josephlr: Good point on the safety stuff, I moved `#[target_feature(enable = "rdrand")]` to also be on `rdrand_exact`. The code is also now slightly faster:

Before:
```
test aes128::bench_rdrand           ... bench:          36 ns/iter (+/- 0) = 444 MB/s
test p256::bench_rdrand             ... bench:          68 ns/iter (+/- 0) = 470 MB/s
test p384::bench_rdrand             ... bench:         101 ns/iter (+/- 1) = 475 MB/s
test page::bench_rdrand             ... bench:       8,334 ns/iter (+/- 62) = 491 MB/s
```
After:
```
test aes128::bench_rdrand           ... bench:          34 ns/iter (+/- 0) = 470 MB/s
test p256::bench_rdrand             ... bench:          66 ns/iter (+/- 0) = 484 MB/s
test p384::bench_rdrand             ... bench:          98 ns/iter (+/- 0) = 489 MB/s
test page::bench_rdrand             ... bench:       8,178 ns/iter (+/- 49) = 500 MB/s
```

src/rdrand.rs
josephlr: As `core::usize::MAX` is deprecated and we can't use `usize:MAX` yet, I just switched to `!0`, which also better expresses the "all Fs" nature of the return value.

PR #337, opened 2023-02-16
Add in a RtlGenRandom fallback for non-UWP Windows

### LINKED ISSUE #314: Windows: should we fall back to RtlGenRandom if BCryptGenRandom fails

chutten:
In some instances BCryptRandom will fail when RtlGenRandom will work. On UWP, we might be unable to actually use RtlGenRandom.

Thread the needle and use RtlGenRandom when we have to, when we're able.

See also rust-lang/rust#108060

Fixes #314

### DIFF
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -21,6 +21,14 @@ extern "system" {
     ) -> u32;
 }
 
+// Forbidden when targetting UWP
+#[cfg(not(target_vendor = "uwp"))]
+#[link(name = "advapi32")]
+extern "system" {
+    #[link_name = "SystemFunction036"]
+    fn RtlGenRandom(RandomBuffer: *mut c_void, RandomBufferLength: u32) -> u8;
+}
+
 pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
     // Prevent overflow of u32
     for chunk in dest.chunks_mut(u32::max_value() as usize) {
@@ -35,6 +43,15 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
         };
         // NTSTATUS codes use the two highest bits for severity status.
         if ret >> 30 == 0b11 {
+            // Failed. Try RtlGenRandom as a fallback.
+            #[cfg(not(target_vendor = "uwp"))]
+            {
+                if unsafe { RtlGenRandom(chunk.as_mut_ptr() as *mut c_void, chunk.len() as u32) }
+                    != 0
+                {
+                    continue;
+                }
+            }
             // We zeroize the highest bit, so the error code will reside
             // inside the range designated for OS codes.
             let code = ret ^ (1 << 31);
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -46,9 +46,8 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
             // Failed. Try RtlGenRandom as a fallback.
             #[cfg(not(target_vendor = "uwp"))]
             {
-                if unsafe { RtlGenRandom(chunk.as_mut_ptr() as *mut c_void, chunk.len() as u32) }
-                    != 0
-                {
+                let ret = unsafe { RtlGenRandom(chunk.as_mut_ptr() as *mut c_void, chunk.len() as u32) };
+                if ret != 0 {
                     continue;
                 }
             }
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -46,7 +46,8 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
             // Failed. Try RtlGenRandom as a fallback.
             #[cfg(not(target_vendor = "uwp"))]
             {
-                let ret = unsafe { RtlGenRandom(chunk.as_mut_ptr() as *mut c_void, chunk.len() as u32) };
+                let ret =
+                    unsafe { RtlGenRandom(chunk.as_mut_ptr() as *mut c_void, chunk.len() as u32) };
                 if ret != 0 {
                     continue;
                 }


chutten:
(Whoops. Built and tested it on my Windows machine _within WSL which isn't at all Windows_. Lemme just tweak and rebuild that real quick.)

src/windows.rs
```diff
@@ -33,17 +39,33 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
...
+
+#[cfg(not(target_vendor = "uwp"))]
+#[inline(never)]
+fn fallback_rng(chunk: &mut [MaybeUninit<u8>]) -> u8 {
+    unsafe { RtlGenRandom(chunk.as_mut_ptr() as *mut u8, chunk.len() as u32) }
+}
+
+#[cfg(target_vendor = "uwp")]
+#[inline(never)]
+fn fallback_rng(_chunk: &mut [MaybeUninit<u8>]) -> u8 {
```
yjugl, line 69: The return type does not seem to match the return value here? `RtlGenRandom` return no more information than TRUE or FALSE (respectively a non-zero or a zero `u8` value), so I think `fallback_rng` could return `bool` or `Maybe<bool>` depending on whether or not we want to be able to distinguish "`RtlGenRandom` is unavailable" from "`RtlGenRandom` failed". Given that at the moment we do not need to be able to distinguish since the error we propagate will be the one from `BCryptGenRandom`, maybe `bool` is enough here.

src/windows.rs
```diff
@@ -33,17 +39,33 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
...
-            return Err(Error::from(code));
+        if ret > 0 {
+            // Failed. Try RtlGenRandom as a fallback.
+            let fallback_ret = fallback_rng(chunk);
+            // NTSTATUS codes use the two highest bits for severity status.
+            if fallback_ret > 0 {
+                // We zeroize the highest bit, so the error code will reside
+                // inside the range designated for OS codes.
+                let code = ret ^ (1 << 31);
+                // SAFETY: the second highest bit is always equal to one,
```
yjugl, line 50: The code before this patch considered that warning values for NTSTATUS are not a failure (that's why it was checking for the two most significant bits of the return value from `BCryptGenRandom`). This is what this remark explains (although I think it's a mistake to consider that warning values are not a failure).

If in this patch we consider that warning values for NTSTATUS are a failure (and I think we should), then this remark does not hold anymore and should be updated. But I think we can still use `NonZeroU32::new_unchecked`, because `NTSTATUS` warning values start with 0x80000001 STATUS_GUARD_PAGE_VIOLATION, so there is no 0x80000000.

src/windows.rs
```diff
@@ -33,17 +39,33 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
...
-            let code = ret ^ (1 << 31);
-            // SAFETY: the second highest bit is always equal to one,
-            // so it's impossible to get zero. Unfortunately the type
-            // system does not have a way to express this yet.
-            let code = unsafe { NonZeroU32::new_unchecked(code) };
-            return Err(Error::from(code));
+        if ret > 0 {
+            // Failed. Try RtlGenRandom as a fallback.
+            let fallback_ret = fallback_rng(chunk);
+            // NTSTATUS codes use the two highest bits for severity status.
```
yjugl, line 45: This comment is outdated / does not match with what the code does.

yjugl (changes requested):
Thank you @chutten for working on this! Here are my remarks on the Windows aspects of this change.

src/windows.rs
```diff
@@ -33,17 +39,33 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
...
-        if ret >> 30 == 0b11 {
-            // We zeroize the highest bit, so the error code will reside
-            // inside the range designated for OS codes.
-            let code = ret ^ (1 << 31);
-            // SAFETY: the second highest bit is always equal to one,
-            // so it's impossible to get zero. Unfortunately the type
-            // system does not have a way to express this yet.
-            let code = unsafe { NonZeroU32::new_unchecked(code) };
-            return Err(Error::from(code));
+        if ret > 0 {
```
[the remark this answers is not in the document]
chutten: Ah, and here I thought I was clever by adapting rust-lang's nt_success... but without thinking about it first. Lemme give this another go, paying more close attention to the types and values (that's what I get for rushing it at the end of my day yesterday)

src/windows.rs
chutten: I'm a fan of patches doing one thing at a time, so I'd prefer to restrain ourselves to not having opinions about the return values (like I accidentally did). If we wish to propose a change so warnings are failures as well (now that we have a fallback to attempt), I feel like that ought to be its own Issue and PR.

chutten:
And to complete the see also links, see also: https://bugzilla.mozilla.org/show_bug.cgi?id=1816953

src/windows.rs
yjugl: OK, good for me then.

src/windows.rs
```diff
@@ -21,6 +21,12 @@ extern "system" {
     ) -> u32;
 }
 
+extern "system" {
+    // Forbidden when targetting UWP
```
josephlr, line 25: Instead of a comment, we should have a `#[cfg(not(target_vendor = "uwp"))]` attribute. Something like:
```rust
#[link(name = "bcrypt")]
extern "system" {
    fn BCryptGenRandom(
        hAlgorithm: *mut c_void,
        pBuffer: *mut u8,
        cbBuffer: u32,
        dwFlags: u32,
    ) -> i32;
}

#[cfg(not(target_vendor = "uwp"))]
#[link(name = "advapi32")]
extern "system" {
  #[link_name = "SystemFunction036"]
  pub fn RtlGenRandom(RandomBuffer: *mut c_void, RandomBufferLength: u32) -> u8;
}
```

Note that the buffer is of type `PVOID` and we need to make sure we link to `Advapi32`

src/windows.rs
```diff
@@ -35,15 +41,33 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
...
-            // We zeroize the highest bit, so the error code will reside
-            // inside the range designated for OS codes.
-            let code = ret ^ (1 << 31);
-            // SAFETY: the second highest bit is always equal to one,
-            // so it's impossible to get zero. Unfortunately the type
-            // system does not have a way to express this yet.
-            let code = unsafe { NonZeroU32::new_unchecked(code) };
-            return Err(Error::from(code));
+            // Failed. Try RtlGenRandom as a fallback.
+            let success = fallback_rng(chunk);
```
josephlr, line 45: This introduces a larger diff and more indentation than needed. Could we just do:

```rust
// NTSTATUS codes use the two highest bits for severity status.
if ret >> 30 == 0b11 {
    // Failed. Try RtlGenRandom as a fallback.
    #[cfg(not(target_vendor = "uwp"))]
    if unsafe { RtlGenRandom(chunk.as_mut_ptr() as *mut c_void, chunk.len() as u32) } != 0 {
        continue;
    }

    // We zeroize the highest bit, so the error code will reside
    // inside the range designated for OS codes.
    let code = ret ^ (1 << 31);
    ...
```

src/windows.rs
josephlr: I agree that our handling of error values here is not ideal. We should mimic what `NT_SUCCESS(Status)` does and only treat values `0 − 0x7FFFFFFF` as "succeeding". However, this is better addressed in a followup PR.

chutten:
Pushed to run a few tests on moz's tryservers here: https://treeherder.mozilla.org/jobs?repo=try&revision=81ff7509fa334c3bf3d2fe0fcce8ca9f2376a6a7

(I expect some oranges here and there from flaky tests, but for it to overall work out).

Attn @yjugl for if you want to doublecheck that this more-final version fixes your particular testcase.

josephlr (approved):

newpavlov:
@chutten 
You need to wrap the `if` in a block to pass our MSRV tests.

chutten:
> @chutten You need to wrap the `if` in a block to pass our MSRV tests.

Oh, because of the `cfg`? Will do.

chutten:
There, that ought to do it. I pulled down 1.36 and it compiles with the extra block. (For anyone reading, I did it via `rustup install 1.36.0` then `rm Cargo.lock` (because 1.36 can't parse the modern lockfile) then `cargo +1.36.0 build`)

src/windows.rs
```diff
@@ -35,6 +43,15 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
...
         // NTSTATUS codes use the two highest bits for severity status.
         if ret >> 30 == 0b11 {
+            // Failed. Try RtlGenRandom as a fallback.
+            #[cfg(not(target_vendor = "uwp"))]
+            {
+                if unsafe { RtlGenRandom(chunk.as_mut_ptr() as *mut c_void, chunk.len() as u32) }
+                    != 0
+                {
+                    continue;
+                }
```
newpavlov, line 53:
```suggestion
                let ret = unsafe { RtlGenRandom(chunk.as_mut_ptr() as *mut c_void, chunk.len() as u32) };
                if ret != 0 {
                    continue;
                }
```

src/windows.rs
```diff
@@ -35,6 +43,14 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
         };
         // NTSTATUS codes use the two highest bits for severity status.
         if ret >> 30 == 0b11 {
+            // Failed. Try RtlGenRandom as a fallback.
+            #[cfg(not(target_vendor = "uwp"))]
+            {
+                let ret = unsafe { RtlGenRandom(chunk.as_mut_ptr() as *mut c_void, chunk.len() as u32) };
```
chutten, line 49:
```suggestion
                let ret =
                    unsafe { RtlGenRandom(chunk.as_mut_ptr() as *mut c_void, chunk.len() as u32) };
```

yjugl:
> Attn @yjugl for if you want to doublecheck that this more-final version fixes your particular testcase.

I was able to confirm that on a Windows 7 machine where `bcryptprimitives.dll` is unavailable, `Uuid::new_v4()` no longer panics and successfully generates a random uuid. Thank you!

chutten:
@newpavlov @josephlr sooooo... when would it be polite to ask about your plans _vis a vis_ cutting a v0.2.9...? : D

notgull:
Bit late to the party, but you probably shouldn't be matching on `target_vendor`. See rust-lang/rust#100343

chutten:
> Bit late to the party, but you probably shouldn't be matching on `target_vendor`. See [rust-lang/rust#100343](https://github.com/rust-lang/rust/issues/100343)

TIL. The specific proposed guidance for UWP will be to replace `target_vendor = "uwp"` with `target_abi = "uwp"`. Plenty of time for that yet, by the looks of things. e.g. [the code we're adapting](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/windows/rand.rs#L24) still uses `target_vendor` for now, at least.

PR #338, opened 2023-02-22
ci: Don't use rust-cache for Cross Build/Link

josephlr:
I think the caching isn't interacting well with the libc build script.

Lets see if this fixes things.


josephlr:
This seems to have fixed the Cross Build and Cross Test CI jobs. It also doesn't seem like it make those jobs take significantly longer.

newpavlov (approved):

PR #357, opened 2023-04-10
Fix v0.2.9 release date

newpavlov:


josephlr (approved):
Thank you for fixing this, sorry that I missed it!

PR #362, opened 2023-05-20
emscripten: use getentropy from libc

newpavlov:
As noted by @kleisauke [here](https://github.com/rust-random/getrandom/pull/307#issuecomment-1555914030) `libc` had added `getenropy` for Emscripten, so we can use it instead of our own declaration. v0.2.143 already [contains](https://github.com/rust-lang/libc/blob/0.2.143/src/unix/linux_like/emscripten/mod.rs) the declaration, so version bump is not needed.


newpavlov:
@josephlr ping

PR #367, opened 2023-06-21
Hash pin github actions and configure dependabot

### LINKED ISSUE #366: Hash pin github workflow dependencies
## Description

Hi again (related issue #360), I'd like to suggest another security practice recommended by the [OpenSSF Scorecard][scorecard-repo] and the [GitHub itself][github-hashpin] which is to hash pin the CI dependencies to prevent tag renaming attacks. This means hash pinning GitHub Workflow actions.

Along with hash-pinning dependencies, I also recommend adopting dependabot or renovatebot to help keep the dependencies up to date. Both tools can update  hashes and associated semantic version comments. 

Together with this issue I'll also suggest a PR with the changes since they are quite simple. 

Any questions or concerns just let me know.
Thanks!

## Additional Context

A [tag renaming attack][tag-renaming] is a type of attack whereby an attacker:
- Hijack an action.
- Upload a malicious version.
- Replace existing tags with malicious versions.

For more informations about the dependency-update tools:
- [Dependabot][dependabot]
- [Renovatebot][renovatebot]

[scorecard-repo]: https://github.com/ossf/scorecard
[tag-renaming]: https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash
[deps-confusion]: https://www.websecuritylens.org/how-dependency-confusion-attack-works-and-how-to-prevent-it/ 
[typosquatting]: https://snyk.io/blog/typosquatting-attacks/ 
[dependabot]: https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/
[renovatebot]:https://www.mend.io/renovate/
[github-hashpin]: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

joycebrum:
Closes #366 
- Hash pin github actions
- Configure dependabot

It is possible to configure renovatebot instead. Let me know if you are interested.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions" # Necessary to update action hashes	
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    # Allow up to 3 opened pull requests for github-actions versions
+    open-pull-requests-limit: 3
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,8 +20,9 @@ jobs:
     name: Docs, deadlinks, minimal dependencies
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v3
-      - uses: dtolnay/rust-toolchain@nightly # Needed for -Z minimal-versions and doc_cfg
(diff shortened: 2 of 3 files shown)


newpavlov:
As commented in the issue, the vulnerability does not affect this repository because we use read-only workflows, so we can leave the CI config as-is.

PR #370, opened 2023-09-21
Add GNU/Hurd support

sthibaul:


josephlr:
Could we have a link to GNU/Hurd's documentation (for both urandom and  the errno location). We should also add it to the list of targets in the CI

sthibaul:
Well, the Hurd doesn't really document it since it's just following GNU's interface. At best we can like to the implementation:

- https://git.savannah.gnu.org/cgit/hurd/hurd.git/tree/sutils/MAKEDEV.sh#n122
- https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/mach/hurd/errno-loc.c;h=823ce696b03bbc58100b5df34f7437c301717e20;hb=HEAD

sthibaul:
Concerning the CI, the patches needed to build rustc and std are not finished committing in the various repositories. Actually getrandom is one of them :) so we have a chicken-and-egg issue if CI needs to be included before support is included

src/lib.rs
```diff
@@ -216,7 +217,7 @@ pub use crate::error::Error;
 // The function MUST NOT ever write uninitialized bytes into `dest`,
 // regardless of what value it returns.
 cfg_if! {
-    if #[cfg(any(target_os = "haiku", target_os = "redox", target_os = "nto", target_os = "aix"))] {
+    if #[cfg(any(target_os = "haiku", target_os = "hurd", target_os = "redox", target_os = "nto", target_os = "aix"))] {
```
newpavlov, line 220: We either should use `all(target_os = "hurd", target_env = "gnu")` or use `*-hurd-*` in the docs table.

newpavlov (approved):

src/lib.rs
sthibaul: ok, used `*-hurd-*`

josephlr:
> Concerning the CI, the patches needed to build rustc and std are not finished committing in the various repositories. Actually getrandom is one of them :) so we have a chicken-and-egg issue if CI needs to be included before support is included

For new targets, we don't require that Rust ships a `libstd` or `libcore` for that target (much less require that Rust have Host Tools for the target).  This is due both to the chicken-and-egg problem you describe and because some targets which we support (like `x86_64-wrs-vxworks` or `aarch64-apple-tvos`) do not have `libstd` support.

However, provided that Rust _supports your compiler target_, we can still test them. [See here](https://github.com/rust-random/getrandom/blob/1308930985a43bf26799699474045f89755c59db/.github/workflows/tests.yml#L289-L326).

However, it looks like https://github.com/rust-lang/rust/pull/115230 is merged, so we should be able to at least have build tests for this once that change hits nightly. @sthibaul do you know if it's possible to link rust binaries for GNU/Hurd on GNU/Linux?

sthibaul:
> do you know if it's possible to link rust binaries for GNU/Hurd on GNU/Linux?

Yes it is possible, but as usual one needs to have the GNU/Hurd libc.so etc. at hand to be able to link. One can relatively easily build them from glibc's `scripts/build-many-glibcs.py`.

josephlr:
> > do you know if it's possible to link rust binaries for GNU/Hurd on GNU/Linux?
> 
> Yes it is possible, but as usual one needs to have the GNU/Hurd libc.so etc. at hand to be able to link. One can relatively easily build them from glibc's `scripts/build-many-glibcs.py`.

I was able to compile a C program for HURD (with Clang) on my Debian Box by just installing `gcc-multilib`, so I hopeful we can also get it to work w/ Rust.

josephlr:
> Well, the Hurd doesn't really document it since it's just following GNU's interface. At best we can link to the implementation:
> 
> * https://git.savannah.gnu.org/cgit/hurd/hurd.git/tree/sutils/MAKEDEV.sh#n122
> * https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/mach/hurd/errno-loc.c;h=823ce696b03bbc58100b5df34f7437c301717e20;hb=HEAD

If Hurd is just implementing glibc interfaces, does it have support for `getrandom` or `getentropy` (which are part of glibc)? If those APIs exist, we should use them instead of the file-based APIs (as the file APIs generally have a bunch of problems).

josephlr:
> > Well, the Hurd doesn't really document it since it's just following GNU's interface. At best we can link to the implementation:
> > 
> > * https://git.savannah.gnu.org/cgit/hurd/hurd.git/tree/sutils/MAKEDEV.sh#n122
> > * https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/mach/hurd/errno-loc.c;h=823ce696b03bbc58100b5df34f7437c301717e20;hb=HEAD
> 
> If Hurd is just implementing glibc interfaces, does it have support for `getrandom` or `getentropy` (which are part of glibc)? If those APIs exist, we should use them instead of the file-based APIs (as the file APIs can sometimes cause problems/toil).

It looks like [`getrandom` is implemented for Hurd](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/mach/hurd/getrandom.c;hb=HEAD) it was added [here](https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=fe75ee0ca65536352fcdaebfbf39aac9bee271f0) in 2019, so it's probably fine to unconditionally depend on it. 

We should have the docs link to the glibc man page: https://www.gnu.org/software/libc/manual/2.38/html_mono/libc.html#index-getrandom

josephlr:
I think this PR is also blocked on libc support, see https://github.com/rust-lang/libc/pull/3325#issuecomment-1730760085

sthibaul:
> > > do you know if it's possible to link rust binaries for GNU/Hurd on GNU/Linux?
> > 
> > 
> > Yes it is possible, but as usual one needs to have the GNU/Hurd libc.so etc. at hand to be able to link. One can relatively easily build them from glibc's `scripts/build-many-glibcs.py`.
> 
> I was able to compile a C program for HURD (with Clang) on my Debian Box by just installing `gcc-multilib`, so I'm hopeful we can also get it to work w/ Rust.

I guess your toolchain was  then using linux's 32bit libc, so while this indeed exercises a hurd build, the actual linked result is a Linux binary, not a Hurd binary :)

sthibaul:
> > Well, the Hurd doesn't really document it since it's just following GNU's interface. At best we can link to the implementation:
> > 
> > * https://git.savannah.gnu.org/cgit/hurd/hurd.git/tree/sutils/MAKEDEV.sh#n122
> > * https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/mach/hurd/errno-loc.c;h=823ce696b03bbc58100b5df34f7437c301717e20;hb=HEAD
> 
> If Hurd is just implementing glibc interfaces, does it have support for `getrandom` or `getentropy` (which are part of glibc)? If those APIs exist, we should use them instead of the file-based APIs (as the file APIs can sometimes cause problems/toil).

Ok, done so!

josephlr:
> (the mips failure is unrelated)

Should be fixed by #372

sthibaul:
I rebased on top of #372, and added std=core for now (building std depends on libc dep bump in rustc)

sthibaul:
The netbsd build failure is unrelated

.github/workflows/tests.yml
```diff
@@ -303,6 +303,8 @@ jobs:
           # `std` support still in progress. Can be moved up with the other
           # apple targets after https://github.com/rust-lang/rust/pull/103503
           aarch64-apple-tvos,
+          # `std` support being commited, see https://github.com/rust-lang/rust/pull/116527
+          i686-unknown-hurd-gnu,
```
newpavlov, line 307: The PR was merged, so we can move it to the `std` targets section (though we may need wait for a new Nightly release).

josephlr:
> > > Well, the Hurd doesn't really document it since it's just following GNU's interface. At best we can link to the implementation:
> > > 
> > > * https://git.savannah.gnu.org/cgit/hurd/hurd.git/tree/sutils/MAKEDEV.sh#n122
> > > * https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/mach/hurd/errno-loc.c;h=823ce696b03bbc58100b5df34f7437c301717e20;hb=HEAD
> > 
> > 
> > If Hurd is just implementing glibc interfaces, does it have support for `getrandom` or `getentropy` (which are part of glibc)? If those APIs exist, we should use them instead of the file-based APIs (as the file APIs can sometimes cause problems/toil).
> 
> Ok, done so!

It seems like this current PR doesn't use the `getrandom` API, we should use that API if possible.

josephlr:
> The netbsd build failure is unrelated

Yep, it's caused by https://github.com/cross-rs/cross/issues/1345

.github/workflows/tests.yml
sthibaul: Indeed. The nightly release seems alright, tests now pass.

sthibaul:
> > > > Well, the Hurd doesn't really document it since it's just following GNU's interface. At best we can link to the implementation:
> > > > 
> > > > * https://git.savannah.gnu.org/cgit/hurd/hurd.git/tree/sutils/MAKEDEV.sh#n122
> > > > * https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/mach/hurd/errno-loc.c;h=823ce696b03bbc58100b5df34f7437c301717e20;hb=HEAD
> > > 
> > > 
> > > If Hurd is just implementing glibc interfaces, does it have support for `getrandom` or `getentropy` (which are part of glibc)? If those APIs exist, we should use them instead of the file-based APIs (as the file APIs can sometimes cause problems/toil).
> > 
> > 
> > Ok, done so!
> 
> It seems like this current PR doesn't use the `getrandom` API, we should use that API if possible.

Indeed, fixed so.

sthibaul:
Any news on this, is there still something missing?

PR #372, opened 2023-10-02
CI: replace MIPS cross tests with PPC32

newpavlov:
`mips-unknown-linux-gnu` is now a Tier 3 target: rust-lang/rust#115218.

This means we can't use it for cross tests anymore since `std` is no longer built for it.

This commit replaces it with `powerpc-unknown-linux-gnu`, a big endian Tier 2 target.


josephlr (approved):

RalfJung:
FWIW Miri now uses s390x-unknown-linux-gnu as big-endian target, it seems to be the most actively maintained.

PR #380, opened 2023-11-07
Add workaround for cross-rs/cross#1345

newpavlov:
While the issue is fixed in the cross' master branch, according to https://github.com/cross-rs/cross/issues/1356 the next release will be a breaking one (v0.3) and it will take some time for all planned changes to be finished.


josephlr (approved):
Thanks for fixing this

Issue #383, opened 2023-11-07
Add support for targets without atomic operations


zeegomo:
This crate cannot currently be built for targets that do not support atomic operations (e.g. riscv32im) due to `LazyBool/Usize` in https://github.com/rust-random/getrandom/blob/baf2198d2a52e16e3c8a820ce9c47f2fe39cbe2d/src/util.rs#L33, even though those are not used when providing a custom implementation.

Example error when compiling for riscv32im
```
error[E0432]: unresolved import `core::sync::atomic::AtomicUsize`
  --> src/util.rs:12:20
   |
12 |     sync::atomic::{AtomicUsize, Ordering::Relaxed},
   |                    ^^^^^^^^^^^ no `AtomicUsize` in `sync::atomic`
```

It would be nice to have this working there as well, a quick workaround could be to conditionally compile those additional structs only when atomics are available.

PR #384, opened 2023-11-07
Add support for targets without atomics

zeegomo:
Conditionally compile LazyUsize/Bool based on whether atomics are available on the target.

Workaround for https://github.com/rust-random/getrandom/issues/383

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -6,66 +6,10 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 #![allow(dead_code)]
-use core::{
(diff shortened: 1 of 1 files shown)


newpavlov:
Right now `LazyUsize` is used only in the `use_file` module, so I think we can simply move it there. `LazyBool` is used in `linux_android` and `rdrand` modules, so I think we can move it into a new `lazy_bool` module and "mount" it inside `cfg_if` similarly to how we handle the `use_file` module.

newpavlov:
Closing in favor of #385

Issue #371, opened 2023-10-01
What versions do platform support track?


BlackHoleFox:
Hiyo. I was poking around random generation for Apple platforms and while tweaking Miri realized `getrandom` attempts to `dlsym` the `getentropy` function. The reasoning has been pretty simple for a while: Old macOS didn't have it. But once Rust 1.74 comes out that won't follow since upstream has now [dropped pre 10.12](https://github.com/rust-lang/rust/pull/104385). I'm working on [cleaning up](https://github.com/rust-lang/rust/pull/116319) things in std as a result.

Similarly, the [Android](https://github.com/rust-random/getrandom/blob/master/src/linux_android.rs#L18) and Linux implementations are also following this as even with their semi-recent bumps, the `getrandom` syscall isn't yet guaranteed to be present.

So that nicely swings around to the question: What is `getrandom`'s policy for platform version assumptions? Does it just follow what the latest `std` does (as [this comment](https://github.com/rust-random/getrandom/blob/master/src/macos.rs#L20) sort of implies)? Or does the 1.36 MSRV imply `getrandom` can't make newer platform assumptions and is stuck supporting versions of OSes that are long deprecated? `libc` seems to have taken the path of not supporting the removed versions via changes such as https://github.com/rust-lang/libc/pull/3362.

josephlr:
I think it would be ok to drop now-unsupoorted OS versions in a non-major release (i.e. this would not be a breaking change).

@newpavlov any thoughts here? It might be nice to have an official policy which we mention in the README.

newpavlov:
We have the precedent of removing Windows XP support in v0.2.1, which has followed the removal in `std`. So I think it's a reasonable policy to follow `std` and allow removal of target support in non-major releases. Post-1.0 we probably should do it in minor releases, but before that we have no choice but to do it in patch releases.

BlackHoleFox:
Great, that makes things pretty easy then. Agree that the minor version increases in the future would be a polite allowance to downstream users.

Since both of you agree on the policy, does this plan seem fine?
- Wait for Rust 1.74 to come out and abandon older Apple OSes.
- I make two PRs:
    - Mirror the in-progress stdlib changes (which will probably be merged by then) for Apple's RNG logic here to match
    - Formally document the policy in the README

josephlr:
@BlackHoleFox that sounds perfect! Thanks for taking care of this.

[...]

PR #387, opened 2023-12-19
Document platform support policy

BlackHoleFox:
This writes down `getrandom`'s unofficial platform support policy to make it official, like I discussed a few months ago. Per that, the policy will be to track the latest `std`'s platform support and allow removing support for the OS versions that upstream Rust has decided aren't worth keeping around.

cc @newpavlov @josephlr 

Closes https://github.com/rust-random/getrandom/issues/371

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ This crate requires Rust 1.36.0 or later.
 ## Platform Support
 
 This crate generally supports the same operating system and platform versions that the Rust standard library does. 
-Additional targets may be supported using plugable custom implementations.
+Additional targets may be supported using pluggable custom implementations.
 
 This means that as Rust drops support for old versions of operating systems (such as old Linux kernel versions, Android API levels, etc)
 in stable releases, `getrandom` may create new patch releases (`0.N.x`) that remove support for outdated platform versions.
--- a/README.md
+++ b/README.md
@@ -54,6 +54,14 @@ crate features, WASM support and Custom RNGs see the
 
 This crate requires Rust 1.36.0 or later.
 
+## Platform Support
+
+This crate generally supports the same operating system and platform versions that the Rust standard library does. 
+Additional targets may be supported using plugable custom implementations.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -54,6 +54,14 @@ crate features, WASM support and Custom RNGs see the
 
 This crate requires Rust 1.36.0 or later.
 
+## Platform Support
+
+This crate supports the same operating system or platform versions that the Rust standard library does, unless a
+[custom implementation](https://docs.rs/getrandom/latest/getrandom/#custom-implementations) is used instead.
```
newpavlov, line 60: The "unless" part looks a bit unclear to me. Currently we do not allow custom backends to overwrite support for built-in targets. I think it could be better to write it like this:

>This crate generally supports the same operating system and platform versions that the Rust standard library does. Additional targets may be supported using plugable custom implementations.

newpavlov (commented):
I think it's worth to add a similar section to the crate docs as well.

README.md
BlackHoleFox: That sounds better to be as well, thanks. Applied it and copy/pasted into the crate documentation.

README.md
```diff
@@ -54,6 +54,14 @@ crate features, WASM support and Custom RNGs see the
 
 This crate requires Rust 1.36.0 or later.
 
+## Platform Support
+
+This crate generally supports the same operating system and platform versions that the Rust standard library does. 
+Additional targets may be supported using plugable custom implementations.
```
newpavlov, line 60:
```suggestion
Additional targets may be supported using pluggable custom implementations.
```

src/lib.rs
```diff
@@ -106,6 +106,13 @@
 //! ```
 //! This crate will then use the provided `webcrypto` implementation.
 //!
+//! ### Platform Support
+//! This crate generally supports the same operating system and platform versions that the Rust standard library does.
+//! Additional targets may be supported using plugable custom implementations.
```
newpavlov, line 111:
```suggestion
//! Additional targets may be supported using pluggable custom implementations.
```

PR #388, opened 2023-12-19
Raise minimum supported Apple OS versions

BlackHoleFox:
This is the second part of #371, and follows the new policy to significantly improve the two Apple OS backends of `getrandom` now that Rust has officially removed support for versions of macOS before 10.12, among others. 

The actual code changes are really a mirror of my work in `std` [here](https://github.com/rust-lang/rust/pull/116319) from a few months back. The two main things are:
- `getentropy` is now unconditionally linked on macOS since its assumed 10.12 is always the minimum deployment target. So macOS no longer needs the `urandom` file reading logic.
- iOS and derivatives changed to use `CCRandomGenerateBytes`, which is part of CommonCrypto inside `libSystem.framework`. This is the same function the previous [SecRandomCopyBytes](https://github.com/apple-oss-distributions/Security/blob/ef677c3d667a44e1737c1b0245e9ed04d11c51c1/OSX/sec/Security/SecBase.c#L107-L109) call wrapped around anyway, so other then dropping the `Security.framework` dependency its identical. Its also what [BoringSSL uses](https://github.com/google/boringssl/blob/bd20800c22fc8402611b537287bd6948c3f2a5a8/crypto/rand_extra/ios.c#L25) if its need better selling.

### DIFF
--- a/src/apple-other.rs
+++ b/src/apple-other.rs
@@ -1,4 +1,4 @@
-// Copyright 2018 Developers of the Rand project.
+// Copyright 2023 Developers of the Rand project.
 //
 // Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
 // https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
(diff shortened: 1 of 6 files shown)


src/apple-other.rs
```diff
@@ -1,24 +1,29 @@
...
 }
 
 pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
-    // Apple's documentation guarantees kSecRandomDefault is a synonym for NULL.
-    let ret = unsafe { SecRandomCopyBytes(null(), dest.len(), dest.as_mut_ptr() as *mut u8) };
-    // errSecSuccess (from SecBase.h) is always zero.
+    let ret = unsafe { CCRandomGenerateBytes(dest.as_mut_ptr() as *mut c_void, dest.len()) };
+    // kCCSuccess (from CommonCryptoError.h) is always zero.
     if ret != 0 {
         Err(Error::IOS_SEC_RANDOM)
```
BlackHoleFox, line 28: This is incorrect now I noticed, but I'm pretty certain this can't be corrected without a breaking API change. I've left it alone for now but let me know if the maintainers think otherwise.

newpavlov: What do you mean by "incorrect" and what API changes may be needed? Is it about the `IOS_SEC_RANDOM` error code?

BlackHoleFox: By "incorrect" I mean this is no longer a `SecRandom` error, and also this isn't just iOS anymore (although it wasn't be that did that). It'd be better to call it `APPLE_RAND` error or something.

newpavlov: I think we can simply update `IOS_SEC_RANDOM` docs and add a todo note to rename it in a future breaking release.

BlackHoleFox: Done.

newpavlov:
Looks good! Please also add an appropriate CHANGELOG entry.

BlackHoleFox:
Added this entry to the CHANGELOG:
- Raise minimum supported Apple OS versions to macOS 10.12 and iOS 10.

newpavlov (approved):

PR #389, opened 2024-01-07
Release v0.2.12

newpavlov:
### Fixed
- Custom backend for targets without atomics [#385]

### Changed
- Improve robustness of the Hermit backend and `sys_fill_exact` [#386]
- Raise minimum supported Apple OS versions to macOS 10.12 and iOS 10 [#388]

### Added
- Document platform support policy [#387]

[#385]: https://github.com/rust-random/getrandom/pull/385
[#386]: https://github.com/rust-random/getrandom/pull/386
[#387]: https://github.com/rust-random/getrandom/pull/387
[#388]: https://github.com/rust-random/getrandom/pull/388

### DIFF
--- a/LICENSE-MIT
+++ b/LICENSE-MIT
@@ -1,4 +1,4 @@
-Copyright 2018 Developers of the Rand project
+Copyright 2018-2024 Developers of the Rand project
 Copyright (c) 2014 The Rust Project Developers
 
 Permission is hereby granted, free of charge, to any
--- a/LICENSE-MIT
+++ b/LICENSE-MIT
@@ -1,4 +1,4 @@
-Copyright 2018-2024 Developers of the Rand project
+Copyright (c) 2018-2024 The rust-random Project Developers
 Copyright (c) 2014 The Rust Project Developers
 
 Permission is hereby granted, free of charge, to any
--- a/src/apple-other.rs
+++ b/src/apple-other.rs
@@ -1,4 +1,4 @@
-// Copyright 2023 Developers of the Rand project.
+// Copyright 2023-2024 Developers of the Rand project.
 //
 // Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
 // https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
--- a/src/apple-other.rs
+++ b/src/apple-other.rs
@@ -1,11 +1,3 @@
-// Copyright 2023-2024 Developers of the Rand project.
-//
-// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
-// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
-// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
-// option. This file may not be copied, modified, or distributed
-// except according to those terms.
-
 //! Implementation for iOS, tvOS, and watchOS where `getentropy` is unavailable.
 use crate::Error;
 use core::{ffi::c_void, mem::MaybeUninit};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,4 @@
-// Copyright 2019 Developers of the Rand project.
+// Copyright 2019-2024 Developers of the Rand project.
 //
 // Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
 // https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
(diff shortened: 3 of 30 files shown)


src/apple-other.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2023 Developers of the Rand project.
+// Copyright 2023-2024 Developers of the Rand project.
```
newpavlov, line 1: Maybe we should remove copyright lines from source code files? Personally, I think they are somewhat pointless, especially considering that we do not include full copyright notices as stated by MIT and Apache licenses

dhardy: I can't offer legal advice @newpavlov.

That said, I guess the purpose may be to have a documented source/copyright owner in case a file is copied in absense of the rest of the code. That said, searching "the Rand project" on Google doesn't appear to help (we should have chosen a better name?).

newpavlov: I think that including license information in each file is important for project which mix different licenses (like the horrible mess in the pre-3.0 OpenSSL). In our case we require that all contributions are licensed under MIT OR Apache 2.0 terms (could be worth to add an explicit [contribution section](https://github.com/RustCrypto/traits#contribution) to README) and we use the umbrella copyright "owner" instead of recording personal contributions in licensing notices.

We do not distribute separate source files, in our case unit of distribution is the whole crate and we already include the appropriate license files in published crates and keep them as part of the repository.

In the Apache license appendix it's implied that the notice should be included in each source file to apply the license. But in my understanding it's nothing more than a recommendation and, either way, we do not follow the notice format in the first place.

dhardy: Do we lose anything by removing the notice? I don't think so since copyright is supposed to be automatic.

Maybe the best option is to mention the licence with a link to the COPYRIGHT file (if we even have that here) in the README and in the top-level library docs (so that it shows up in generated docs; strange that nobody seems to do that), but not bother in other source files (aside from the Cargo.toml line).

newpavlov: I think the current [notice](https://github.com/rust-random/getrandom#license) in README should be sufficient and adding a similar notice to the library docs sounds reasonable. In future, after an appropriate MSRV bump we should use `#![doc = include_str!("../README.md")]` to reduce duplication.

>strange that nobody seems to do that

I think most people simply consider this information redundant in docs. Documentation is for people who want to learn how to use the crate, while licensing information is available in Cargo.toml, on crates.io, and even on docs.rs. Plus, it can be retrieved automatically using tools like `cargo license` and overwhelming amount of crates on crates.io is published under liberal licenses, so having similar MIT/Apache terms in most of crates' docs would feel like a waste of space.

LICENSE-MIT
```diff
@@ -1,4 +1,4 @@
-Copyright 2018 Developers of the Rand project
+Copyright 2018-2024 Developers of the Rand project
 Copyright (c) 2014 The Rust Project Developers
```
dhardy, line 2: Can we change this from "Rand project" to "the rust-random project developers"? That at least is searchable.

src/lib.rs
```diff
@@ -158,6 +150,21 @@
...
+//! * [Apache License, Version 2.0][LICENSE-APACHE]
+//! * [MIT license][LICENSE-MIT]
+//!
+//! at your option.
+//!
+//! ## Contribution
+//!
+//! Unless you explicitly state otherwise, any contribution intentionally submitted
+//! for inclusion in the work by you, as defined in the Apache-2.0 license, shall be
+//! dual licensed as above, without any additional terms or conditions.
```
dhardy, line 166: Good. (Though probably the contribution part only needs to appear in the README.)

LICENSE-MIT
newpavlov: Done.

dhardy (approved):

dhardy:
(Ignore that; I shouldn't be reviewing this repo.)

josephlr (approved):
LGTM. Thanks for removing all the per-file copyright headers. We were never consistent about this stuff regardless. Having the LICENSE files, README, and `Cargo.toml` metadata be the authoritative place for this works for me.

josephlr:
@newpavlov I'll let you merge and publish this (as I've published on the wrong date before). Also, at your option, we could remove the Licensing documentation from the RustDocs (lib.rs), as it doesn't really cover how to _use_ the crate, but rather how to _contribute_ to the crate.

I like the documentation updates though, much clearer now.

PR #390, opened 2024-01-07
Remove NetBSD workaround

newpavlov:
The issue was fixed in October, so we probably no longer need the workaround.


newpavlov:
Unfortunately, maintainers of cross haven't made a release with the fix, so we have to wait for v0.3.0 release.

PR #391, opened 2024-01-16
Correct comments regarding LazyUsize

divergentdave:
This corrects a couple comments in the LazyUsize struct. This code was only moved between files recently. I removed a reference to the now-nonexistent sync_init, and fixed an "at least"/"at most" typo.


newpavlov (approved):
Thank you!

 Can you also fix docs on line 20 from "synchronization methods with be needed" to "synchronization methods may be needed"?

PR #392, opened 2024-01-19
Use `doc_auto_cfg` instead of `doc_cfg`

newpavlov:


josephlr (approved):
Looks good to me, I ran
```
RUSTDOCFLAGS="--cfg docsrs" cargo doc --features=custom,std
```
and the results look good.

Pic3
Pic2
Pic1

PR #396, opened 2024-02-19
Conditionally disable file fallback for Android and Linux

### LINKED ISSUE #229: Should `getrandom()` really fall back to File I/O when the getrandom syscall returns EPERM?
If the `getrandom` syscall returns `EPERM` then it's probably because (a) the OS is buggy, e.g. QNAP apparently, or (b) the syscall is blocked by seccomp filtering or similar.

In the case where the syscall is blocked by a sandbox, I think it doesn't make sense to fall back to File I/O; why would a sandbox intentionally block `getrandom` but also intentionally allow file I/O that/s equivalent or worse (w.r.t security)? This indicates to me that the sandbox is buggy and needs to be fixed.

In the case of the QNAP situation, it is purely a bug on their part and it seems like it should be fixed on their side, instead of making security worse on our side. Note that we don't fall back on EFAULT which is another buggy result that sometimes occurs for this syscall.

At the very least, I'd love to have a way to opt out of falling back to File I/O on EPERM, but ideally we'd only fall back on ENOSYS.

### LINKED ISSUE #376: Don't fall back to file I/O on Linux when Rust only supports kernels guaranteed to have it
According to https://doc.rust-lang.org/beta/rustc/platform-support.html, there are only a handful of target architectures where Rust supports kernels that are so old that they would not have the `getrandom` syscall. Notably `riscv64gc` and `aarch64` are guaranteed to have `getrandom` according to the Rust kernel support policy. For those architectures we should not have enable the fallback logic at all. I suggest in particular that we have an allowlist of target architectures where we allow the fallback, which would be `i686`, `x86_64`, `arm`, `armv7`, `powerpc`, `powerpc64`, `powerpc64le`. `s390x`, `i586`, according to the Rust documents.

newpavlov:
The file fallback is left enabled for Android targets and for Linux targets which support pre-3.17 kernels.

This PR also adds `linux_disable_fallback` crate feature and removes catching of `EPERM` during `getrandom` availability check on Android targets.

Closes #376
Closes #229

cc @briansmith

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ wasm-bindgen-test = "0.3.18"
 std = []
 # Disable `/dev/urandom` fallback for Linux and Android targets.
 # Bumps minimum supported Linux kernel version to 3.17 and Android API level to 23 (Marshmallow).
-disable_urandom_fallback = []
+linux_disable_fallback = []
 # Feature to enable fallback RDRAND-based implementation on x86/x86_64
 rdrand = []
 # Feature to enable JavaScript bindings on wasm*-unknown-unknown
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,9 @@ wasm-bindgen-test = "0.3.18"
 [features]
 # Implement std-only traits for getrandom::Error
 std = []
+# Disable `/dev/urandom` fallback for Linux and Android targets.
+# Bumps minimum supported Linux kernel version to 3.17 and Android API level to 23 (Marshmallow).
+disable_urandom_fallback = []
 # Feature to enable fallback RDRAND-based implementation on x86/x86_64
 rdrand = []
 # Feature to enable JavaScript bindings on wasm*-unknown-unknown
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -224,7 +224,7 @@ cfg_if! {
         mod util_libc;
         #[path = "use_file.rs"] mod imp;
     } else if #[cfg(all(
-        not(feature = "disable_urandom_fallback"),
+        not(feature = "linux_disable_fallback"),
         any(
             // Rust supports Android API level 19 (KitKat) [0] and the next upgrade targets
             // level 21 (Lollipop) [1], while `getrandom(2)` was added only in
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -99,11 +99,14 @@
 //! This crate will then use the provided `webcrypto` implementation.
 //!
 //! ### Platform Support
-//! This crate generally supports the same operating system and platform versions that the Rust standard library does.
-//! Additional targets may be supported using pluggable custom implementations.
+//! This crate generally supports the same operating system and platform versions
+//! that the Rust standard library does. Additional targets may be supported using
+//! pluggable custom implementations.
 //!
-//! This means that as Rust drops support for old versions of operating systems (such as old Linux kernel versions, Android API levels, etc)
-//! in stable releases, `getrandom` may create new patch releases (`0.N.x`) that remove support for outdated platform versions.
+//! This means that as Rust drops support for old versions of operating systems
+//! (such as old Linux kernel versions, Android API levels, etc) in stable releases,
+//! `getrandom` may create new patch releases (`0.N.x`) that remove support for
+//! outdated platform versions.
 //!
 //! ### Custom implementations
 //!
@@ -220,10 +223,48 @@ cfg_if! {
     if #[cfg(any(target_os = "haiku", target_os = "redox", target_os = "nto", target_os = "aix"))] {
         mod util_libc;
         #[path = "use_file.rs"] mod imp;
-    } else if #[cfg(any(target_os = "android", target_os = "linux"))] {
+    } else if #[cfg(all(
+        not(feature = "disable_urandom_fallback"),
+        any(
+            // Rust supports Android API level 19 (KitKat) [0] and the next upgrade targets
+            // level 21 (Lollipop) [1], while `getrandom(2)` was added only in
+            // level 23 (Marshmallow). Note that it applies only to the "old" `target_arch`es,
+            // RISC-V Android targets sufficiently new API level, same will apply for potential
+            // new Android `target_arch`es.
+            // [0]: https://blog.rust-lang.org/2023/01/09/android-ndk-update-r25.html
+            // [1]: https://github.com/rust-lang/rust/pull/120593
+            all(
+                target_os = "android",
+                any(
+                    target_arch = "aarch64",
+                    target_arch = "arm",
+                    target_arch = "x86",
+                    target_arch = "x86_64",
+                ),
+            ),
+            // Only on these `target_arch`es Rust supports Linux kernel versions (3.2+)
+            // that precede the version (3.17) in which `getrandom(2)` was added:
+            // https://doc.rust-lang.org/stable/rustc/platform-support.html
+            all(
+                target_os = "linux",
+                any(
+                    target_arch = "aarch64",
+                    target_arch = "arm",
+                    target_arch = "powerpc",
+                    target_arch = "powerpc64",
+                    target_arch = "s390x",
+                    target_arch = "x86",
+                    target_arch = "x86_64",
+                ),
+            )
+        ),
+    ))] {
         mod util_libc;
         mod use_file;
         mod lazy;
+        #[path = "linux_android_with_fallback.rs"] mod imp;
+    } else if #[cfg(any(target_os = "android", target_os = "linux"))] {
+        mod util_libc;
         #[path = "linux_android.rs"] mod imp;
     } else if #[cfg(any(target_os = "illumos", target_os = "solaris"))] {
         mod util_libc;
--- a/src/linux_android.rs
+++ b/src/linux_android.rs
@@ -1,40 +1,7 @@
-//! Implementation for Linux / Android
-use crate::{
-    lazy::LazyBool,
-    util_libc::{last_os_error, sys_fill_exact},
-    {use_file, Error},
-};
+//! Implementation for Linux / Android without `/dev/urandom` fallback
+use crate::{util_libc, Error};
 use core::mem::MaybeUninit;
 
 pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
-    // getrandom(2) was introduced in Linux 3.17
-    static HAS_GETRANDOM: LazyBool = LazyBool::new();
-    if HAS_GETRANDOM.unsync_init(is_getrandom_available) {
-        sys_fill_exact(dest, |buf| unsafe {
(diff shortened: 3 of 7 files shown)


src/lib.rs
```diff
@@ -220,10 +223,27 @@ cfg_if! {
...
+        target_os = "linux",
+        any(
+            target_arch = "i586",
+            target_arch = "i686",
+            target_arch = "x86",
+            target_arch = "x86_64",
+            target_arch = "arm",
+            target_arch = "armv7",
+            target_arch = "powerpc",
+            target_arch = "powerpc64le",
```
briansmith, line 237:
```
$ rustc --print=cfg --target powerpc64le-unknown-linux-gnu | grep target_arch
target_arch="powerpc64"
```
This should be `powerpc64`.

src/lib.rs
```diff
@@ -220,10 +223,27 @@ cfg_if! {
...
+    } else if #[cfg(all(
+        not(feature = "disable_urandom_fallback"),
+        target_os = "linux",
+        any(
+            target_arch = "i586",
+            target_arch = "i686",
+            target_arch = "x86",
+            target_arch = "x86_64",
+            target_arch = "arm",
+            target_arch = "armv7",
```
briansmith, line 235: "armv7" should be removed:
```
$ rustc --print=cfg --target armv7-unknown-linux-gnueabi | grep target_arch
target_arch="arm"
```

src/lib.rs
```diff
@@ -220,10 +223,27 @@ cfg_if! {
     if #[cfg(any(target_os = "haiku", target_os = "redox", target_os = "nto", target_os = "aix"))] {
         mod util_libc;
         #[path = "use_file.rs"] mod imp;
-    } else if #[cfg(any(target_os = "android", target_os = "linux"))] {
+    } else if #[cfg(all(
+        not(feature = "disable_urandom_fallback"),
+        target_os = "linux",
+        any(
+            target_arch = "i586",
```
briansmith, line 230: Please sort these alphabetically.

src/linux_with_fallback.rs
```diff
@@ -0,0 +1,40 @@
+//! Implementation for Linux / Android
```
briansmith, line 1: This comment is now inaccurate since this doesn't apply to Android.

src/linux_android.rs
```diff
@@ -1,40 +1,14 @@
 //! Implementation for Linux / Android
```
briansmith, line 1: It's better to qualify this " when fallback to /dev/urandom is disabled."

src/linux_android.rs
```diff
@@ -1,40 +1,14 @@
...
-            Some(libc::ENOSYS) => false, // No kernel support
-            Some(libc::EPERM) => false,  // Blocked by seccomp
-            _ => true,
-        }
-    } else {
-        true
-    }
-}
-
-unsafe fn getrandom(
```
briansmith, line 34: I would keep the getrandom function (renamed `getrandom_inner_inner`?) and put it in a place where both implementations can use it. I would also change the signature to use `buf: dest: &mut [MaybeUninit<u8>]` instead of `buf` and `buflen` so that the unsafe `buf.as_mut_ptr() as *mut libc::c_void` bit is in one place.

src/lib.rs
```diff
@@ -220,10 +223,27 @@ cfg_if! {
...
+            target_arch = "i686",
+            target_arch = "x86",
+            target_arch = "x86_64",
+            target_arch = "arm",
+            target_arch = "armv7",
+            target_arch = "powerpc",
+            target_arch = "powerpc64le",
+            target_arch = "s390x",
+        ),
+    ))] {
```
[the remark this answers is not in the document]
briansmith: Note that the link I suggest above is for the Rust 1.47.0 platform support documentation. The MSRV of getrandom is older than 1.47.0, but 1.47.0 is the oldest version that has documentation online that I could find. Perhaps it would be better to link to the source of the file in GitHub as of whatever revision was in effect for our MSRV.

src/lib.rs
```diff
@@ -220,10 +223,27 @@ cfg_if! {
     if #[cfg(any(target_os = "haiku", target_os = "redox", target_os = "nto", target_os = "aix"))] {
         mod util_libc;
         #[path = "use_file.rs"] mod imp;
-    } else if #[cfg(any(target_os = "android", target_os = "linux"))] {
+    } else if #[cfg(all(
+        not(feature = "disable_urandom_fallback"),
+        target_os = "linux",
+        any(
+            target_arch = "i586",
+            target_arch = "i686",
```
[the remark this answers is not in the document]
newpavlov: Oh, you are right. I've copied list from your comment without properly checking it.

src/lib.rs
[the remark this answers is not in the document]
newpavlov: I've looked into it a bit more and it looks like Rust supports Android API level 19 (KitKat), while `getrandom(2)` was added only in level 23 (Marshmallow). So it's too soon to disable the file fallback for Android.

src/linux_android.rs
newpavlov: Added `getrandom_syscall` function in the `util_libc` module.

src/lib.rs
```diff
@@ -220,10 +223,35 @@ cfg_if! {
     if #[cfg(any(target_os = "haiku", target_os = "redox", target_os = "nto", target_os = "aix"))] {
         mod util_libc;
         #[path = "use_file.rs"] mod imp;
-    } else if #[cfg(any(target_os = "android", target_os = "linux"))] {
+    } else if #[cfg(all(
+        not(feature = "disable_urandom_fallback"),
+        any(
+            // Rust supports Android API level 19 (KitKat) [0], while `getrandom(2)`
+            // was added only in level 23 (Marshmallow).
+            // [0]: https://blog.rust-lang.org/2023/01/09/android-ndk-update-r25.html
```
briansmith, line 231: I also see that rust-lang/rust is about to bump up to API level 21 in https://github.com/rust-lang/rust/pull/120593, which is still not enough.

src/linux_android_with_fallback.rs
```diff
@@ -0,0 +1,30 @@
...
+        use_file::getrandom_inner(dest)
+    }
+}
+
+fn is_getrandom_available() -> bool {
+    if getrandom_syscall(&mut []) < 0 {
+        match last_os_error().raw_os_error() {
+            Some(libc::ENOSYS) => false, // No kernel support
+            #[cfg(target_os = "linux")]
+            Some(libc::EPERM) => false,  // Blocked by seccomp
```
briansmith, line 24: Maybe it's better to make this change to the Android behavior in a separate PR that can be managed separately. I will add comments to the issue about this.

src/lib.rs
```diff
@@ -220,10 +223,35 @@ cfg_if! {
...
+            target_os = "android",
```
briansmith, line 232: I think this should also be conditional on target_arch in the same way that it is for "linux", except it should also include "aarch64" for Android only.

For example, RISC-V Android would never (I hope) ship with a kernel so old that it would need this fallback.

src/linux_android_with_fallback.rs
briansmith: If we do leave this change in, then we should add a comment like this:
```
// The fallback on EPERM is intentionally not done on Android since this workaround seems to be needed only for specific Linux-based products that aren't based on Android. See https://github.com/rust-random/getrandom/issues/229.
```

src/linux_android_with_fallback.rs
```diff
@@ -0,0 +1,30 @@
...
+    util_libc::{getrandom_syscall, last_os_error, sys_fill_exact},
+    {use_file, Error},
+};
+use core::mem::MaybeUninit;
+
+pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
+    // getrandom(2) was introduced in Linux 3.17
+    static HAS_GETRANDOM: LazyBool = LazyBool::new();
+    if HAS_GETRANDOM.unsync_init(is_getrandom_available) {
+        sys_fill_exact(dest, getrandom_syscall)
```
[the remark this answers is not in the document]
newpavlov: We need `getrandom_syscall` for checking `getrandom` availability, so I think the current solution is fine.

src/lib.rs
newpavlov: To be safe I've also added `x86` and `x86_64` to this list.

src/linux_android_with_fallback.rs
newpavlov: Added comment.

src/lib.rs
newpavlov: Added link to the Rust PR.

newpavlov:
The CI failure is caused by https://github.com/rust-lang/rust/issues/121460.

newpavlov:
@josephlr WDYT?

Cargo.toml
```diff
@@ -32,6 +32,9 @@ wasm-bindgen-test = "0.3.18"
 [features]
 # Implement std-only traits for getrandom::Error
 std = []
+# Disable `/dev/urandom` fallback for Linux and Android targets.
+# Bumps minimum supported Linux kernel version to 3.17 and Android API level to 23 (Marshmallow).
+disable_urandom_fallback = []
```
josephlr, line 37: This feature name is a little confusing, because it is Linux-specific. Enabling this feature won't disable reading from `/dev/urandom` on other non-Linux targets.

Ideally, we would just have a `disable_fallback` feature which could be enabled on whatever targets/arches people wanted. However, doing this is complicated (https://github.com/rust-lang/cargo/issues/1197), so I'm not sure how well this approach would work.

Given this the two alternatives would be:
  - Having a `disable_legacy_fallback` which disables all sufficiently old fallback methods.   For example, this would also disable the fallback code on Solaris / Illumos
  - Rename the feature to `linux_disable_file_fallback` to make it clear that this is a feature-specific target.

We could also just remove this feature from this PR and have a followup PR that just adds the fallback-disabling

josephlr (approved):
This looks good! Thanks for working on this.

My only concerns are with how the new Cargo feature should work. If you want, you can merge the rest of this and we can discuss in a followup PR (or we can discuss it here).

Cargo.toml
newpavlov: `linux_disable_file_fallback` sounds good. It's a bit long, but should be tolerable enough.

josephlr: `linux_disable_fallback` would also be fine. On linux, there's only one relevant fallback (reading from a file).

newpavlov:
>My only concerns are with how the new Cargo feature should work.

I think we can discuss it here. It looks like the Nightly is broken on some targets which causes breakage of some CI jobs, so I do not plan to merge it right away.

josephlr (approved):

newpavlov:
The CI failures are caused by Nightly issues and not relevant to this PR, so we can merge it.

PR #404, opened 2024-03-25
Replace man7.org links with manned.org

### LINKED ISSUE #403: man7 link should be replaced
man7 hosts a very outdated version of manual pages from the linux man-pages project, among other sources. https://github.com/mkerrisk/man-pages/issues/33

There are sites that provide up-to-date manpages, as suggested in https://github.com/mkerrisk/man-pages/issues/33#issuecomment-1501450876. Compare:
* https://man7.org/linux/man-pages/man4/urandom.4.html
* https://manned.org/urandom.4
* https://www.mankier.com/4/urandom

(At the time of writing the pages are the same. Which is kind of my fault, because I was supposed to follow through on my patch to update them after the big Linux RNG changes.)

newpavlov:
It looks like man7.org website is no longer maintained (for at least 3 years), so as suggested by @Artoria2e5 manned.org links are used instead.

Closes #403

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -150,8 +150,8 @@
 //! on every call to `getrandom`, hence after the first successful call one
 //! can be reasonably confident that no errors will occur.
 //!
-//! [1]: http://man7.org/linux/man-pages/man2/getrandom.2.html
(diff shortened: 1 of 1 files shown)


josephlr (approved):
Nice! Thanks

Issue #400, opened 2024-02-21
Spurious failures of Web tests on Windows


newpavlov:
Web tests spuriously [fail](https://github.com/rust-random/getrandom/actions/runs/7989698114/job/21816805556?pr=396) on Windows on `Run wasm-pack test --headless --chrome --features=js,test-in-browser` with the following message:

```
 driver status: exit code: 1
driver stdout:
    Starting ChromeDriver 121.0.6167.184 (057a8ae7deb3374d0f1b04b36304d236f0136188-refs/branch-heads/6167@{#1818}) on port 50916
    Only local connections are allowed.
    Please see https://chromedriver.chromium.org/security-considerations for suggestions on keeping ChromeDriver safe.

Error: driver failed to bind port during startup
[2024-02-21T13:23:08Z ERROR tiny_http] Error accepting new client: A blocking operation was interrupted by a call to WSACancelBlockingCall. (os error 10004)
```

Similar errors also happen with Firefox and Edge.

newpavlov:
As a workaround we have temporarily marked Web Windows tests as `continue-on-error: true` in #405.

PR #405, opened 2024-04-05
Release v0.2.13

newpavlov:
### Added
- `linux_disable_fallback` crate feature to disable `/dev/urandom`-based fallback on Linux and Android targets. Enabling this feature bumps minimum supported Linux kernel version to 3.17 and Android API level to 23 (Marshmallow). [#396]

### Changed
- Disable `/dev/urandom` fallback for Linux targets outside of the following `target_arch`es: `aarch64`, `arm`, `powerpc`, `powerpc64`, `s390x`, `x86`, `x86_64` [#396]
- Do not catch `EPERM` error code on Android while checking availability of the `getrandom` syscall [#396]

[#396]: https://github.com/rust-random/getrandom/pull/396

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,9 +1,1 @@
-# Allow normal use of "cargo run" and "cargo test" on these wasm32 platforms.
-[target.wasm32-unknown-unknown]
-runner = 'wasm-bindgen-test-runner'
-[target.wasm32-wasi]
-runner = 'wasmtime'
-
-# Just run on node by default (that's where emscripten is tested)
-[target.'cfg(target_os = "emscripten")']
-runner = 'node'
+.cargo/config.toml
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,1 +1,1 @@
-.cargo/config.toml
+config.toml
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -229,6 +229,9 @@ jobs:
             host: x86_64-unknown-linux-musl
           - os: windows-2022
             host: x86_64-pc-windows-msvc
+            # We get spurious failures on Windows, see:
+            # https://github.com/rust-random/getrandom/issues/400
+            continue-on-error: true
           - os: macos-12
             host: x86_64-apple-darwin
     runs-on: ${{ matrix.os }}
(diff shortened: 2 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -250,6 +250,9 @@ jobs:
       - name: Test (Firefox)
         run: wasm-pack test --headless --firefox --features=js,test-in-browser
       - name: Test (Chrome)
+        # We get spurious failures, see:
+        # https://github.com/rust-random/getrandom/issues/400
```
josephlr, line 254: Ideally, we would still detect Chrome failures on working OSes (Linux/macOS). We could either:
  - Not test Chrome on  Windows. We are already running Edge there which is just reskinned Chrome.
  - Update the matrix to just mark Windows as flaky in general (we have Edge and Chrome problems on Windows)

```
matrix:
  include:
    - os: ubuntu-22.04
      host: x86_64-unknown-linux-musl
    - os: windows-2022
      host: x86_64-pc-windows-msvc
      continue-on-error: true
    - os: macos-12
      host: x86_64-apple-darwin
```

josephlr (approved):
Looks nice! Thanks for fixing up the Cargo and CI files!

.github/workflows/tests.yml
newpavlov: Yeah, I also [observed](https://github.com/rust-random/getrandom/actions/runs/8569924021/job/23487836088) spurious failures with Firefox on Windows as well, though they are less frequent than with Chrome, so temporarily applying `continue-on-error: true` for the whole Web Windows job sounds like a good idea.

.cargo/config
josephlr, line 1: This symlink is also invalid, it needs to be `config.toml` not `.cargo/config.toml` (relative to current directory, not root directory).

josephlr (changes requested):

.cargo/config
newpavlov: Oh, true.

.github/workflows/tests.yml
josephlr: Sounds good, we can use #400 to track how to actually fix these tests.

.github/workflows/tests.yml
josephlr, line 1: While we're changing the web tests, we should probably also update wasm-pack's `VERSION` from `0.11.0` to `0.12.1`.

newpavlov: Done.

newpavlov:
@josephlr 
Could you re-approve?

josephlr (approved):

newpavlov:
@josephlr 
BTW I usually erase commit history from merge commit description, since, unless intentionally kept clean, it does not contain useful information for main commit history.

josephlr:
> @josephlr BTW I usually erase commit history from merge commit description, since, unless intentionally kept clean, it does not contain useful information for main commit history.

That's a good point. I'' try to remeber to do that in the future.

PR #406, opened 2024-04-05
Fix web tests on windows

josephlr:


newpavlov:
This PR is out of date, so I think we can close it. It will be better to start with a fresh PR.

PR #410, opened 2024-04-16
Add Apple visionOS support

QuentinPerez:
## Motivation

Add the support of Apple VisionOS  [121419](https://github.com/rust-lang/rust/pull/121419)

## Solution
Add target_os = "visionos"


madsmtm:
CC @agg23, does this match what you've done locally?

agg23:
I don't think I've used any RNG, so I don't have any such change locally (I was temporarily confused though, because I _did_ make unrelated changes for _riscv_ in `random`).

newpavlov (approved):
Thank you!

Issue #414, opened 2024-04-30
Windows: Switch to using ProcessPRNG by default.


josephlr:
[`ProcessPRNG`](https://learn.microsoft.com/en-us/windows/win32/seccng/processprng) was introduced in Windows 10, which is the minimum supported Windows for most Rust targets. Upstream change: https://github.com/rust-lang/rust/pull/121337

BoringSSL now uses `ProcessPRNG`: https://github.com/google/boringssl/blob/2db0eb3f96a5756298dcd7f9319e56a98585bd10/crypto/rand_extra/windows.c#L64
Golang also switched to `ProcessPRNG`: https://go-review.googlesource.com/c/go/+/536235

Should also help with the Chromium Sandbox: https://issues.chromium.org/issues/40277768

ChrisDenton:
The fourth option would be to include your own pre-built `.lib` file (possibly generated by using `raw-dylib` and storing the output or else copying the windows-rs lib generator code).

> Adds call to dlltool during build time

Note that msvc targets don't require dlltool.

josephlr:
> The fourth option would be to include your own pre-built `.lib` file (possibly generated by using `raw-dylib` and storing the output or else copying the windows-rs lib generator code).
> 
> > Adds call to dlltool during build time
> 
> Note that msvc targets don't require dlltool.

Thanks! Updated the list. I do think that maintaining a collection of `.lib` files for each windows target is a lot more effort than just writing the ~20 lines of DLL loading code.

@ChrisDenton I have a question though, if `getrandom` lazily calls `LoadLibraryA`/`GetProcAddress` on first use, while `libstd` uses `raw-dylib`, will this cause any issues? When experimenting, it seems like `LoadLibraryA`/`GetProcAddress` always return the same handle/address when run in the same process, so I think there's some sort of caching?

josephlr:
@newpavlov @ChrisDenton  I thinking of the following approach which would combine approaches (1) and (3).

- Add a `build.rs` script to check the compiler version:
  - Output `cargo::rustc-cfg=raw_dylib` if version >= 1.65 (See https://github.com/rust-lang/rust/pull/99916)
  - Output `cargo::rustc-cfg=raw_dylib_x86` if version >= 1.71 (See https://github.com/rust-lang/rust/pull/109677)
- In Windows.rs we have a `cfg_if` block like:
```rust
cfg_if! {
    if #[cfg(all(raw_dylib, not(target_arch = "x86")))] {
[...]

Issue #285, opened 2022-10-04
check libc before using `SYS_getrandom` on Linux 


ahgamut:
Hi,

I was porting [HVM](https://github.com/Kindelia/HVM) to run on [Cosmopolitan Libc](https://github.com/jart/cosmopolitan), and I noticed a small difference between how `getrandom` is handled by this crate when compared to the Rust standard library.

The Rust standard library has a [`syscall` macro which creates a weak symbol](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/weak.rs#L167) for `getrandom`, so if the underlying libc provides `getrandom`, that function is called instead of making a syscall. The standard library uses `getrandom` for HashMap keys [here](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/rand.rs#L40).

The `getrandom` crate makes the Linux syscall [directly](https://github.com/rust-random/getrandom/blob/master/src/linux_android.rs#L46) without first checking for `getrandom` in the libc. 

The pattern used in the Rust standard library is flexible to which libc is used, be it glibc, musl, or Cosmopolitan. 

Could you change https://github.com/rust-random/getrandom/blob/master/src/linux_android.rs to check if `getrandom` is provided by the libc first, before using the syscall?

josephlr:
Good catch, this library aims to have an identical implementation to the standard library (see https://github.com/rust-lang/rust/issues/62079), so we shouldn't deviate unless it's necessary. I think adding this sort of detection would be fine. Our logic would then become:
 - Use `libc::getrandom` if it exists (requires glibc 2.25 or musl 1.1.20)
 - Use raw syscall if it exists (requires Linux 3.17 or a backported kernel like RHEL 7)
 - Use the `/dev/urandom` file as the final fallback.

newpavlov:
Personally, on Linux I prefer using syscalls directly, to the point of thinking that ideally we should not rely on `libc::syscall` and instead use something like the `sycall` crate (it's a real shame in my opinion that we do not have a `libc`-free target).

Regardless of the approach used by `std`, what are the practical motivations for prioritizing `libc::getrandom` over `libc::syscall` + `libc::SYS_getrandom`? On top of my head I can think of the following:
[...]

PR #415, opened 2024-04-30
Use ProcessPrng on Windows

josephlr:
Use [`ProcessPrng`](https://learn.microsoft.com/en-us/windows/win32/seccng/processprng) on Windows 10 and up, and use [`RtlGenRandom`](https://learn.microsoft.com/en-us/windows/win32/api/ntsecapi/nf-ntsecapi-rtlgenrandom) on older legacy Windows versions. Don't use [`BCryptGenRandom`](https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom) due to stability issues.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -159,6 +159,18 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo test --features=std
 
+  windows7:
+    name: Test Windows 7 impl on Windows 10
+    runs-on: windows-2022
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@nightly
+        with:
+          components: rust-src
+      - uses: Swatinem/rust-cache@v2
+      - run: cargo test --target=x86_64-win7-windows-msvc -Z build-std --features=std
+      - run: cargo test --target=i686-win7-windows-msvc -Z build-std --features=std
+
   cross-tests:
     name: Cross Test
     runs-on: ubuntu-22.04
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,7 +44,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-22.04, windows-2022]
-        toolchain: [nightly, beta, stable, 1.38]
+        toolchain: [nightly, beta, stable, 1.56]
         # Only Test macOS on stable to reduce macOS CI jobs
         include:
           # x86_64-apple-darwin.
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -164,8 +164,10 @@ jobs:
     runs-on: windows-2022
     steps:
       - uses: actions/checkout@v3
-      - uses: dtolnay/rust-toolchain@nightly
+      # Win7 targets are Tier3, so pin a nightly where libstd builds.
+      - uses: dtolnay/rust-toolchain@master
         with:
+          toolchain: nightly-2024-05-20
           components: rust-src
       - uses: Swatinem/rust-cache@v2
       - run: cargo test --target=x86_64-win7-windows-msvc -Z build-std --features=std
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -1,60 +1,40 @@
-//! Implementation for Windows
+//! Implementation for Windows 10 and later
+//!
+//! On Windows 10 and later, ProcessPrng "is the primary interface to the
+//! user-mode per-processer PRNGs" and only requires BCryptPrimitives.dll,
+//! making it a better option than the other Windows RNG APIs:
+//!   - BCryptGenRandom: https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom
+//!     - Requires Bcrypt.dll (which loads BCryptPrimitives.dll anyway)
+//!     - Can cause crashes/hangs as BCrypt accesses the Windows Registry:
+//!       https://github.com/rust-lang/rust/issues/99341
+//!     - Causes issues inside sandboxed code:
+//!       https://issues.chromium.org/issues/40277768
+//!   - CryptGenRandom: https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptgenrandom
+//!     - Deprecated and not available on UWP targets
+//!     - Requires Advapi32.lib/Advapi32.dll
+//!     - Wrapper around ProcessPrng
+//!   - RtlGenRandom: https://learn.microsoft.com/en-us/windows/win32/api/ntsecapi/nf-ntsecapi-rtlgenrandom
+//!     - Deprecated and not available on UWP targets
+//!     - Requires Advapi32.dll (and using name "SystemFunction036")
+//!     - Wrapper around ProcessPrng
+//! For more information see the Windows RNG Whitepaper: https://aka.ms/win10rng
 use crate::Error;
-use core::{ffi::c_void, mem::MaybeUninit, num::NonZeroU32, ptr};
+use core::mem::MaybeUninit;
 
-const BCRYPT_USE_SYSTEM_PREFERRED_RNG: u32 = 0x00000002;
-
-#[link(name = "bcrypt")]
-extern "system" {
-    fn BCryptGenRandom(
-        hAlgorithm: *mut c_void,
-        pBuffer: *mut u8,
-        cbBuffer: u32,
-        dwFlags: u32,
-    ) -> u32;
-}
-
-// Forbidden when targetting UWP
-#[cfg(not(target_vendor = "uwp"))]
-#[link(name = "advapi32")]
-extern "system" {
-    #[link_name = "SystemFunction036"]
-    fn RtlGenRandom(RandomBuffer: *mut c_void, RandomBufferLength: u32) -> u8;
-}
+// ProcessPrng lacks an import library, so we use the windows-targets crate to
+// link to it. The following code was generated via windows-bindgen with APIs:
+//   Windows.Win32.Foundation.TRUE
+//   Windows.Win32.Security.Cryptography.ProcessPrng
+windows_targets::link!("bcryptprimitives.dll" "system" fn ProcessPrng(pbdata: *mut u8, cbdata: usize) -> BOOL);
+#[repr(transparent)]
+#[derive(PartialEq, Eq)]
+pub struct BOOL(pub i32);
+pub const TRUE: BOOL = BOOL(1i32);
 
 pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
-    // Prevent overflow of u32
-    for chunk in dest.chunks_mut(u32::max_value() as usize) {
-        // BCryptGenRandom was introduced in Windows Vista
-        let ret = unsafe {
-            BCryptGenRandom(
-                ptr::null_mut(),
-                chunk.as_mut_ptr().cast::<u8>(),
-                chunk.len() as u32,
-                BCRYPT_USE_SYSTEM_PREFERRED_RNG,
-            )
-        };
-        // NTSTATUS codes use the two highest bits for severity status.
-        if ret >> 30 == 0b11 {
-            // Failed. Try RtlGenRandom as a fallback.
-            #[cfg(not(target_vendor = "uwp"))]
-            {
-                let ret = unsafe {
-                    RtlGenRandom(chunk.as_mut_ptr().cast::<c_void>(), chunk.len() as u32)
-                };
-                if ret != 0 {
-                    continue;
-                }
-            }
-            // We zeroize the highest bit, so the error code will reside
-            // inside the range designated for OS codes.
-            let code = ret ^ (1 << 31);
-            // SAFETY: the second highest bit is always equal to one,
-            // so it's impossible to get zero. Unfortunately the type
-            // system does not have a way to express this yet.
-            let code = unsafe { NonZeroU32::new_unchecked(code) };
-            return Err(Error::from(code));
-        }
+    // ProcessPrng should always return TRUE, but we check just in case.
+    match unsafe { ProcessPrng(dest.as_mut_ptr().cast::<u8>(), dest.len()) } {
+        TRUE => Ok(()),
+        _ => Err(Error::WINDOWS_PROCESS_PRNG),
     }
-    Ok(())
 }
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -1,35 +1,33 @@
 //! Implementation for Windows 10 and later
 //!
 //! On Windows 10 and later, ProcessPrng "is the primary interface to the
-//! user-mode per-processer PRNGs" and only requires BCryptPrimitives.dll,
+//! user-mode per-processer PRNGs" and only requires bcryptprimitives.dll,
 //! making it a better option than the other Windows RNG APIs:
 //!   - BCryptGenRandom: https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom
-//!     - Requires Bcrypt.dll (which loads BCryptPrimitives.dll anyway)
+//!     - Requires bcrypt.dll (which loads bcryptprimitives.dll anyway)
 //!     - Can cause crashes/hangs as BCrypt accesses the Windows Registry:
 //!       https://github.com/rust-lang/rust/issues/99341
 //!     - Causes issues inside sandboxed code:
 //!       https://issues.chromium.org/issues/40277768
 //!   - CryptGenRandom: https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptgenrandom
 //!     - Deprecated and not available on UWP targets
-//!     - Requires Advapi32.lib/Advapi32.dll
-//!     - Wrapper around ProcessPrng
+//!     - Requires advapi32.lib/advapi32.dll (in addition to bcryptprimitives.dll)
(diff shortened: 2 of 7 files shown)


src/windows.rs
```diff
@@ -1,59 +1,30 @@
...
+type ProcessPrng = unsafe extern "system" fn(*mut u8, usize) -> BOOL;
 
-// Forbidden when targetting UWP
-#[cfg(not(target_vendor = "uwp"))]
-#[link(name = "advapi32")]
+#[link(name = "kernel32")]
 extern "system" {
-    #[link_name = "SystemFunction036"]
-    fn RtlGenRandom(RandomBuffer: *mut c_void, RandomBufferLength: u32) -> u8;
+    fn LoadLibraryA(libfilename: *const u8) -> HMODULE;
```
briansmith, line 13: Chromium and some other projects specifically uses `LoadLibraryW`, so it would be good to use it instead of `LoadLibraryA`.

More generally, this won't work inside many sandboxes, including Chromium's.

See https://source.chromium.org/chromium/chromium/src/+/main:sandbox/policy/win/sandbox_warmup.cc;bpv=1;bpt=1. In order to support sandboxes, `getrandom` should have a similar "call this before enabling the sandbox" API that does the bare minimum (LoadLibrary, and then store the function pointer, without actualy calling `ProcessPrng`).

IIRC, on Windows there is a way to tell the linker to generate metadata that says "we will delay-load bcryptprimitives.dll from system32". using `/delayload:bcryptprimitives.dll` or similar. I can't remember why, but in other projects I've worked on, we specifically had to ensure that we did this for every delay-loaded library. I think we had to do this because there is something (internal to those projects? anti-virus in general?) that freaks out when LoadLibrary is used on something that isn't mentioned in that metadata.

src/windows.rs
```diff
@@ -1,59 +1,30 @@
...
-            // We zeroize the highest bit, so the error code will reside
-            // inside the range designated for OS codes.
-            let code = ret ^ (1 << 31);
-            // SAFETY: the second highest bit is always equal to one,
-            // so it's impossible to get zero. Unfortunately the type
-            // system does not have a way to express this yet.
-            let code = unsafe { NonZeroU32::new_unchecked(code) };
-            return Err(Error::from(code));
-        }
+    let dll = unsafe { LoadLibraryA(b"bcryptprimitives.dll\0".as_ptr()) };
```
briansmith, line 18: We should use `LoadLibraryeX` with `LOAD_LIBRARY_SEARCH_SYSTEM32` instead (but see the documentation about that flag and compatibility) or `LOAD_WITH_ALTERED_SEARCH_PATH` with an absolute path to the System32 folder.

Google Chrome doesn't security bug reports about this issue to be valid, but not everybody agrees with Google Chrome. Regardless, I can tell you from past experience that this will trigger security vulnerability bug reports to every application that uses getrandom, if we don't restrict the search path to `System32`. These bug reports are difficult for many projects to deal with. It's worth avoiding.

src/windows.rs
briansmith: > IIRC, on Windows there is a way to tell the linker to generate metadata that says "we will delay-load bcryptprimitives.dll from system32". using /delayload:bcryptprimitives.dll or similar. I can't remember why, but in other projects I've worked on, we specifically had to ensure that we did this for every delay-loaded library. I think we had to do this because there is something (internal to those projects? anti-virus in general?) that freaks out when LoadLibrary is used on something that isn't mentioned in that metadata.

Also, these were Rust-based applications that used (indirectly) `getrandom`, though it has been a while since I've worked on them.

src/windows.rs
ChrisDenton: If it's in `\KnownDlls` then it doesn't particularly matter because it will indeed prioritize the system directory for those dlls. However `bcryptprimitives` is a bit more awkward because it's a late addition, iirc the earliest versions of Windows 10 did not have it as a known dll.

briansmith, line 18: Another thing I noticed: This is basically implementing `Weak` for Windows. Issue #285 suggests doing similar for Linux. I don't know about NetBSD (the current sole user of `Weak`), but in general concerns about `dlsym`/LoadLibrary/etc., especially regarding sandboxes, are similar across platforms. Therefore, I think having an explicit `Weak` abstraction that explicitly provides an interface for handling these concerns (e.g. minimal pre-sandbox-enabling initialization) may be helpful.

josephlr:
@briansmith the above are really good point w.r.t. sandboxing. I think that it would be good to have general documentation along the lines of "before starting a sandbox code should first successfully call `getrandom()` on a non-empty buffer". That should be good platform-agnostic advice, and will handle things like `LoadLibrary` and `libc::dlsym`.

> More generally, this won't work inside many sandboxes, including Chromium's.

I think that this won't cause issues in some sandboxes provided that `ProcessPRNG` is already loaded. IIRC the sandbox only complains on loading a new dll, not upon looking up a symbol from an already loaded DLL. Regardless, having specific documentation will be good here.

.github/workflows/tests.yml
```diff
@@ -159,6 +159,18 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo test --features=std
 
+  windows7:
+    name: Windows 7 Test
+    runs-on: windows-2022
```
briansmith, line 164: We should add a note that this tests the windows 7 target build, but the actual runtime environment is still like Windows 10+.

src/windows.rs
```diff
@@ -1,59 +1,15 @@
...
-}
-
-// Forbidden when targetting UWP
-#[cfg(not(target_vendor = "uwp"))]
-#[link(name = "advapi32")]
-extern "system" {
-    #[link_name = "SystemFunction036"]
-    fn RtlGenRandom(RandomBuffer: *mut c_void, RandomBufferLength: u32) -> u8;
-}
+windows_targets::link!("bcryptprimitives.dll" "system" fn ProcessPrng(pbdata : *mut u8, cbdata : usize) -> BOOL);
```
briansmith, line 8: I don't think there's any real advantage to using windows-targets instead of windows-sys. The issue I mentioned in #414 applies equally to windows-targets because they seem to do the major version bumps of these crates at the same time. We should just use windows-sys. And we should then avoid defining `BOOL` and `TRUE`.

src/windows.rs
```diff
@@ -1,59 +1,15 @@
...
-            // We zeroize the highest bit, so the error code will reside
-            // inside the range designated for OS codes.
-            let code = ret ^ (1 << 31);
-            // SAFETY: the second highest bit is always equal to one,
-            // so it's impossible to get zero. Unfortunately the type
-            // system does not have a way to express this yet.
-            let code = unsafe { NonZeroU32::new_unchecked(code) };
-            return Err(Error::from(code));
-        }
+    match unsafe { ProcessPrng(dest.as_mut_ptr() as *mut u8, dest.len()) } {
```
briansmith, line 11: NIT: `.cast::<u8>()`

src/windows.rs
```diff
@@ -1,59 +1,15 @@
...
+        TRUE => Ok(()),
+        _ => Err(Error::WINDOWS_PROCESS_PRNG),
```
briansmith, line 13: According to the documentation, this is a permanently unreachable branch. At a minimum we should document that we expect this to be unreachable.

src/windows7.rs
```diff
@@ -0,0 +1,23 @@
...
+const TRUE: BOOLEAN = 1;
+
+#[link(name = "advapi32")]
+extern "system" {
+    #[link_name = "SystemFunction036"]
+    fn RtlGenRandom(RandomBuffer: *mut c_void, RandomBufferLength: ULONG) -> BOOLEAN;
+}
+
+pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
+    // Prevent overflow of ULONG
```
briansmith, line 15: Even when building for a win7 target, this is actually probably going to be running on Windows 10.

Especially with that in mind, but regardless, we should document here why we don't try `BCryptGenRandom` before `RtlGenRandom` like we did before. It seems to be like in the abstract we should keep the same logic we had before. One concrete reason to avoid trying `BCryptGenRandom` before `RtlGenRandom` is that we would end up never testing the `RtlGenRandom` code path because we have no runtime environment that would trigger the fallback logic.

Regardless of what we do, we need to update the documentation about what we use on each of win7 and regular *-windows-* targets as part of this change.

src/windows.rs
josephlr: We no longer dynamically choose the implementation on Windows, so resolving this.

src/windows.rs
josephlr: We no longer dynamically choose the implementation on Windows, so resolving this.

src/windows.rs
josephlr: Per discussion in #414, this now uses `windows-targets`, but with code based on that output by `windows-bindgen`.

src/windows.rs
josephlr: Added a note that this is not expected to happen.

src/windows7.rs
josephlr: Added a bunch of comments explaining the different options and why we do what we do.

josephlr:
@newpavlov and @briansmith this is now ready for review!

.github/workflows/tests.yml
josephlr: Renamed the test to indicate this.

src/windows7.rs
```diff
@@ -0,0 +1,37 @@
...
+// but we avoid using windows-targets as it doesn't support older Windows.
+#[link(name = "advapi32")]
+extern "system" {
+    #[link_name = "SystemFunction036"]
+    fn RtlGenRandom(randombuffer: *mut c_void, randombufferlength: u32) -> BOOLEAN;
+}
+#[repr(transparent)]
+#[derive(PartialEq, Eq)]
+pub struct BOOLEAN(pub u8);
+pub const TRUE: BOOLEAN = BOOLEAN(1u8);
```
newpavlov, line 26: I think this looks... redundant? We can just use `u8` for `RtlGenRandom` like we did before. Same for the `ProcessPrng` signature.

src/windows.rs
```diff
@@ -1,60 +1,40 @@
...
-            // SAFETY: the second highest bit is always equal to one,
-            // so it's impossible to get zero. Unfortunately the type
-            // system does not have a way to express this yet.
-            let code = unsafe { NonZeroU32::new_unchecked(code) };
-            return Err(Error::from(code));
-        }
+    // ProcessPrng should always return TRUE, but we check just in case.
+    match unsafe { ProcessPrng(dest.as_mut_ptr().cast::<u8>(), dest.len()) } {
+        TRUE => Ok(()),
+        _ => Err(Error::WINDOWS_PROCESS_PRNG),
```
newpavlov, line 38: Since the return value is guaranteed by the function docs, we can use `Error::UNEXPECTED`.

newpavlov (approved):
Other than the two minor nits, it looks good to me. Bu as I noted in the issue, I would prefer if we released this change as v0.3.

src/windows.rs
```diff
@@ -1,60 +1,40 @@
...
+//!       https://github.com/rust-lang/rust/issues/99341
+//!     - Causes issues inside sandboxed code:
+//!       https://issues.chromium.org/issues/40277768
+//!   - CryptGenRandom: https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptgenrandom
+//!     - Deprecated and not available on UWP targets
+//!     - Requires Advapi32.lib/Advapi32.dll
+//!     - Wrapper around ProcessPrng
+//!   - RtlGenRandom: https://learn.microsoft.com/en-us/windows/win32/api/ntsecapi/nf-ntsecapi-rtlgenrandom
+//!     - Deprecated and not available on UWP targets
+//!     - Requires Advapi32.dll (and using name "SystemFunction036")
```
briansmith, line 18: I suggest adding "in addition to BCryptPrimitives.dll` to clarify this.

src/windows.rs
```diff
@@ -1,60 +1,40 @@
...
-}
+// ProcessPrng lacks an import library, so we use the windows-targets crate to
+// link to it. The following code was generated via windows-bindgen with APIs:
+//   Windows.Win32.Foundation.TRUE
+//   Windows.Win32.Security.Cryptography.ProcessPrng
+windows_targets::link!("bcryptprimitives.dll" "system" fn ProcessPrng(pbdata: *mut u8, cbdata: usize) -> BOOL);
+#[repr(transparent)]
+#[derive(PartialEq, Eq)]
+pub struct BOOL(pub i32);
+pub const TRUE: BOOL = BOOL(1i32);
```
briansmith, line 32: It's unfortunate to have this ugliness without the program that generated it to confirm that it actually created this ugliness. However, I also think it's fine because of how `repr(transparent)` works so it isn't the end of the world.

src/windows7.rs
```diff
@@ -0,0 +1,37 @@
+//! Implementation for Windows 7 and 8
```
briansmith, line 1: This is misleading, because it for Windows 7 and later, and in some ways even Windows XP and later, not just 7 and 8.

src/windows7.rs
```diff
@@ -0,0 +1,37 @@
...
+//!
+//! For targets where we cannot use ProcessPrng (added in Windows 10), we use
+//! RtlGenRandom. See windows.rs for a more detailed discussion of the Windows
+//! RNG APIs (and why we don't use BCryptGenRandom). On versions prior to
+//! Windows 10, this implementation works, while on Windows 10 and later, this
+//! is a thin wrapper around ProcessPrng.
```
briansmith, line 7: So basically when the program is running on Windows 10 it is the same as what's in windows.rs for non-UWP targets, except it forces the loading of an additional DLL, and is less aesthetically pleasing? I think we should write down that expectation here so we remember it.

src/windows7.rs
briansmith: Unfortunate but fine, IMO.

briansmith (changes requested):

src/windows.rs
josephlr: It doesn't hurt to have an additional error code, and this is (essentially) the same pattern we already do for `Error::WINDOWS_RTL_GEN_RANDOM`, so the consistency here seems fine.

src/windows.rs
josephlr: Done

src/windows7.rs
josephlr: Fixed to note that it works for XP and later

src/windows7.rs
josephlr: Fixed to note that the windows7 implementation is identical modulo the loading of DLLs.

src/windows7.rs
josephlr: Moved to the easier to read:
```rust
type BOOLEAN = u8;
const TRUE: BOOLEAN = 1u8;
```
as this code wasn't directly generated by `windows-bindgen` anyway.

src/windows.rs
josephlr: I think this is a good point. For both APIs I didn't _directly_ use the output of `windows-bindgen`, but instead took that output as a starting point, then deleted lines until I got code that looked decent. I think that referencing `windows-bindgen` in the comments at all was misleading, so I changed the comments to just reference the API metadata names directly.

josephlr: For reference, here's the full output for running `windows-bindgen` with the following arguments:
```rust
// Bindings generated by `windows-bindgen` 0.56.0

#![allow(
    non_snake_case,
    non_upper_case_globals,
    non_camel_case_types,
    dead_code,
    clippy::all
)]
#[inline]
pub unsafe fn RtlGenRandom(
    randombuffer: *mut core::ffi::c_void,
    randombufferlength: u32,
) -> BOOLEAN {
    windows_targets::link!("advapi32.dll" "system" "SystemFunction036" fn RtlGenRandom(randombuffer : *mut core::ffi::c_void, randombufferlength : u32) -> BOOLEAN);
    RtlGenRandom(randombuffer, randombufferlength)
}
#[inline]
pub unsafe fn ProcessPrng(pbdata: &mut [u8]) -> BOOL {
    windows_targets::link!("bcryptprimitives.dll" "system" fn ProcessPrng(pbdata : *mut u8, cbdata : usize) -> BOOL);
    ProcessPrng(
        core::mem::transmute(pbdata.as_ptr()),
        pbdata.len().try_into().unwrap(),
    )
}
#[repr(transparent)]
#[derive(PartialEq, Eq)]
pub struct BOOL(pub i32);
impl Default for BOOL {
    fn default() -> Self {
        unsafe { core::mem::zeroed() }
    }
}
impl Clone for BOOL {
    fn clone(&self) -> Self {
        *self
    }
}
impl Copy for BOOL {}
impl core::fmt::Debug for BOOL {
    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
        f.debug_tuple("BOOL").field(&self.0).finish()
    }
}
impl windows_core::TypeKind for BOOL {
    type TypeKind = windows_core::CopyType;
}
#[repr(transparent)]
#[derive(PartialEq, Eq)]
pub struct BOOLEAN(pub u8);
impl Default for BOOLEAN {
    fn default() -> Self {
        unsafe { core::mem::zeroed() }
    }
}
impl Clone for BOOLEAN {
    fn clone(&self) -> Self {
        *self
    }
}
impl Copy for BOOLEAN {}
impl core::fmt::Debug for BOOLEAN {
    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
        f.debug_tuple("BOOLEAN").field(&self.0).finish()
    }
}
impl windows_core::TypeKind for BOOLEAN {
    type TypeKind = windows_core::CopyType;
}
pub const TRUE: BOOL = BOOL(1i32);

```

PR #416, opened 2024-05-02
Unconditionally use libc::getrandom on FreeBSD

### LINKED ISSUE #412: Remove fallback logic for FreeBSD
Rust's support for FreeBSD was bumped to 12.0 in https://github.com/rust-lang/rust/issues/89058#issuecomment-1441529961

The standard library changed to unconditionally using `getrandom(2)` on freebsd: https://github.com/rust-lang/rust/pull/107159

josephlr:
Fixes #412 

Rust's minimum version is now FreeBSD 12, so we can drop the fallback code. We have to keep the NetBSD fallback code as NetBSD 10 is still quite new.


newpavlov (approved):
Maybe rename `bsd_arandom.rs` to `netbsd.rs`?

briansmith:
> Rust's minimum version is now FreeBSD 12, so we can drop the fallback code.

The current MSRV does support FreeBSD 11, so this seems premature if the reason is what Rust supports. A better reason is that FreeBSD 11 is EOL.

PR #417, opened 2024-05-02
Unconditionally use `libc::getrandom` on Illumos and `libc::geentropy` on Solaris

### LINKED ISSUE #413: Investigate removing fallback logic for Solaris/Illumos
I filed https://github.com/rust-lang/rust/issues/124549 to see what the minimum version of Illumos/Solaris is for Rust. If we get clarification on this, we can remove the file-based fallback code for these targets.

josephlr:
Fixes #413 

Also removes the use of `GRND_RANDOM`, which appears to be based on outdated staements about the RNGs. For Solaris, see [this blog post](https://blogs.oracle.com/solaris/post/solaris-new-system-calls-getentropy2-and-getrandom2). For Illumos, the algorithms are less clear, but I don't see a clear reason to continue using `GRND_RANDOM`.

I updated the documentation in `getrandom.rs` to full document this decision and to have a common place listing when `getrandom(2)` became avalible on each platform. I also updated the main lib.rs docs to point to the correct man pages.

Note that Solaris 11.3 has a maximum buffer length of 1024 bytes, while Illumos doesn't have this sort of issue.


src/getrandom.rs
```diff
@@ -1,9 +1,42 @@
...
+//!   - Solaris 11.3
+//!   - Illumos since Dec 2018
+//!   - DragonFly 5.7
+//!   - Hurd Glibc 2.31
+//!   - shim-3ds since Feb 2022
+//!
+//! For all platforms, we use the default randomness source (the one used
+//! by /dev/urandom) rather than the /dev/random (GRND_RANDOM) source.
+//!   - On Linux, "/dev/urandom is preferred and sufficient in all use cases".
+//!   - On NetBSD, "there is no reason to ever use [GRND_RANDOM]".
```
newpavlov, line 17: `[GRND_RANDOM]` is a dangling link. Also it's worth to add source links for the quotes.

src/getrandom.rs
```diff
@@ -1,9 +1,42 @@
...
+#[cfg(not(target_os = "solaris"))]
+const MAX_BYTES: usize = usize::MAX;
+
 pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
-    sys_fill_exact(dest, |buf| unsafe {
-        libc::getrandom(buf.as_mut_ptr() as *mut libc::c_void, buf.len(), 0)
-    })
+    for chunk in dest.chunks_mut(MAX_BYTES) {
+        sys_fill_exact(chunk, |buf| unsafe {
+            libc::getrandom(buf.as_mut_ptr() as *mut libc::c_void, buf.len(), FLAGS)
```
newpavlov, line 38: `FLAGS` is not defined anywhere.

src/getrandom.rs
```diff
@@ -1,9 +1,42 @@
...
+//!     https://github.com/illumos/illumos-gate/blob/89cf0c2ce8a47dcf555bb1596f9034f07b9467fa/usr/src/lib/libc/port/gen/getentropy.c#L33
+//!   - On Solaris, both sources use FIPS 140-2 / NIST SP-900-90A DRBGs, see:
+//!     https://blogs.oracle.com/solaris/post/solaris-new-system-calls-getentropy2-and-getrandom2
+//!   - On Redox, only /dev/urandom is provided.
+//!   - On AIX, /dev/urandom will "provide cryptographically secure output".
+//!   - On Haiku, QNX Neutrino, DragonFly, and FreeBSD, they are identical.
 use crate::{util_libc::sys_fill_exact, Error};
 use core::mem::MaybeUninit;
 
+// On some Solaris versions, getrandom() will fail if bufsz > 1024. This issue
```
newpavlov, line 28: Mention the Solaris version with this restriction? IIUC it no longer applies to latest releases?

src/getrandom.rs
josephlr: Fixed the link issue. The additional quotes are from the man pages linked in `lib.rs`, so I just updated the comment to "see the linked man pages in lib.rs".

src/getrandom.rs
josephlr: Fixed

src/getrandom.rs
josephlr: Fixed, this issue also applies in the latest version, Solaris 11.4 (but with a larger threshold)

src/getrandom.rs
```diff
@@ -1,9 +1,43 @@
...
+#[cfg(not(target_os = "solaris"))]
+const MAX_BYTES: usize = usize::MAX;
+
 pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
-    sys_fill_exact(dest, |buf| unsafe {
-        libc::getrandom(buf.as_mut_ptr() as *mut libc::c_void, buf.len(), 0)
-    })
+    for chunk in dest.chunks_mut(MAX_BYTES) {
+        sys_fill_exact(chunk, |buf| unsafe {
+            libc::getrandom(buf.as_mut_ptr() as *mut libc::c_void, buf.len(), 0)
```
newpavlov, line 39: Per Solaris [docs](https://docs.oracle.com/cd/E88353_01/html/E37841/getrandom-2.html):
>This means this is not an acceptable calling sequence: `(void) getrandom(&buf, sizeof (buf), 0);`

Maybe keep Solaris code in a separate module for now, since it's the only target with `MAX_BYTES` and zero flag idiosyncrasies.

josephlr: Could we just move solaris to use `getentropy`? Seems like it might be a better fit with the buffer limitations and everything.

newpavlov: IIUC `getentropy` has even stricter buffer limitation equal to 256 bytes (the `EIO` error). The docs also not clear on what exactly `getentropy` does. It may be equivalent to calling `getrandom` with `GRND_RANDOM`.

josephlr: After #418 goes in, I can switch the Solaris implementation in this PR to `getentropy`.

newpavlov: I am fine with moving Solaris to `getentropy`. I may be somewhat suboptimal as noted above, but it should be good enough in most cases.

jclulow:
Seems reasonable from an illumos perspective, with a small nit: we don't capitalise the "i" in illumos when it appears in print.  Thanks!

newpavlov:
@josephlr 
I've switched Soalris to the `getentropy` backend and tweaked some docs. If you do not have any other changes in mind, I think we can merge this PR.

josephlr:
Thanks for merging this (and sorry for not being responsive).

> Per Solaris [docs](https://docs.oracle.com/cd/E88353_01/html/E37841/getrandom-2.html):
> 
> > This means this is not an acceptable calling sequence: `(void) getrandom(&buf, sizeof (buf), 0);`
> 
> Maybe keep Solaris code in a separate module for now, since it's the only target with `MAX_BYTES` and zero flag idiosyncrasies? IIUC we will need to use `GRND_NONBLOCK` and if it returns `EAGAIN` wait on `GRND_RANDOM`.

@newpavlov after rereading the solaris documentation, I think we misread things. That comment isn't saying "passing `0` for `flags` can be insecure during early boot". It's saying "you need to check the return value to make sure the buffer actually got filled with random bytes" (which we obviously do). Furthermore, the following line from the man page seems to suggest that this function behaves correctly during early boot:

> If no entropy is available in the pool, the getrandom() function will block

There's also the following comment from [the blog post introducing `getrandom()` and `getentropy()`](https://blogs.oracle.com/solaris/post/solaris-new-system-calls-getentropy2-and-getrandom2)

> On Solaris the output of getentropy(2) is entropy and should not be used where randomness is needed, in particular it must not be used where an IV or nonce is needed when calling a cryptographic operation.  It is intended only for seeding a user space RBG (Random Bit Generator) system. More specifically the data returned by getentropy(2) has not had the required FIPS 140-2 processing for the DRBG applied to it.

Given this, should we use `getrandom(2)` on Solaris (with the additional max buffer logic)?

newpavlov:
@josephlr 
It looks you are right, though it's a bit concerning that all "correct" examples use `GRND_RANDOM`.

>Given this, should we use getrandom(2) on Solaris (with the additional max buffer logic)?

Probably, yes. But we may want to keep it in a separate module. In addition to the max buffer logic, Solaris also [guarantees](https://blogs.oracle.com/solaris/post/solaris-new-system-calls-getentropy2-and-getrandom2) that input buffer will be fully filled:
>The other difference is that on Solaris `getrandom(2)` will either fail completely or will return a buffer filled with the requested size, where as the Linux implementation can return partial buffers.

So we do not need to call `sys_fill_exact`. IIUC number of bytes can be smaller only if the `GRND_NONBLOCK` flag is used.

PR #418, opened 2024-05-02
Unify getentropy-based implementations

josephlr:
This is similar to what we did for `getrandom.rs`, and makes switching just Solaris to `getentropy` easier.


newpavlov (approved):
This adds 256-byte chunking for Emscripten, which, strictly speaking, is not needed, but I guess it's fine.

josephlr:
> This adds 256-byte chunking for Emscripten, which, strictly speaking, is not needed, but I guess it's fine.

Ya I thought about special-casing this, but I think the fact that emscripten's `getentropy` works with buffers longer than 256 bytes is undocumented, so maybe it's not the best that we were depending on it.

PR #427, opened 2024-05-20
Rename `Weak` to `LazyPtr` and move it to `lazy.rs`

newpavlov:
Currently the `Weak` type is used only by the NetBSD backend, but in future it may be used by other backends. To simplify code a bit and prepare for potential use on Windows, `Weak` now accepts a "pointer initialization function" instead of a function name, i.e. it now works similarly to `LazyUsize` and `LazyBool`.


briansmith:
I think we should have a plan for #285 first. It may be that we start using `Weak` on Linux too?

Also, I think we should sort out what we're doing for Windows, as we're basically implementing `Weak` using the Windows APIs in #415. Although PR #415 isn't written as an implementation of `Weak` for Windows, I think we should do that, if we're going with the LoadLibrary(Ex)-based approach, because the sandboxing concerns (at least) are analogous across platforms.

josephlr:
Agreed with @briansmith here, lets figure out if we need Weak on other targets first.

newpavlov:
I have changed this PR and now `Weak` resides in a separate module.

src/weak.rs
```diff
@@ -0,0 +1,59 @@
...
+impl Weak {
+    // A non-null pointer value which indicates we are uninitialized. This
+    // constant should ideally not be a valid address of a function pointer.
+    // However, if by chance libc::dlsym does return UNINIT, there will not
+    // be undefined behavior. libc::dlsym will just be called each time ptr()
+    // is called. This would be inefficient, but correct.
+    // TODO: Replace with core::ptr::invalid_mut(1) when that is stable.
+    const UNINIT: *mut c_void = 1 as *mut c_void;
+
+    // Construct a weak binding a C function.
```
josephlr, line 26: Nit: "to a C function."

src/weak.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub const fn new(link_fn: fn() -> *mut c_void) -> Self {
+        Self {
+            addr: AtomicPtr::new(Self::UNINIT),
+            link_fn,
+        }
+    }
+
+    // Return the address of a function if present at runtime. Otherwise,
+    // return None. Multiple callers can call ptr() concurrently. It will
+    // always return _some_ value returned by libc::dlsym. However, the
```
josephlr, line 36: "_some_ value returned by link_fn. However, link_fn may be called multiple times."

src/weak.rs
```diff
@@ -0,0 +1,59 @@
...
+    // dlsym function may be called multiple times.
+    pub fn ptr(&self) -> Option<NonNull<c_void>> {
+        // Despite having only a single atomic variable (self.addr), we still
+        // cannot always use Ordering::Relaxed, as we need to make sure a
+        // successful call to dlsym() is "ordered before" any data read through
```
josephlr, line 41: "call to link_fn()"

src/weak.rs
```diff
@@ -0,0 +1,59 @@
...
+// except that the caller must manually cast self.ptr() to a function pointer.
+pub struct Weak {
+    addr: AtomicPtr<c_void>,
+    link_fn: fn() -> *mut c_void,
+}
+
+impl Weak {
+    // A non-null pointer value which indicates we are uninitialized. This
+    // constant should ideally not be a valid address of a function pointer.
+    // However, if by chance libc::dlsym does return UNINIT, there will not
```
josephlr, line 20: // However, if by chance link_fn does return UNINIT, there will not
    // be undefined behavior. link_fn will just be called each time ptr()

src/weak.rs
josephlr: This should also explain what link_fn should do.

josephlr (approved):
Looks good, just a few nits.

src/netbsd.rs
```diff
@@ -28,8 +25,14 @@ fn kern_arnd(buf: &mut [MaybeUninit<u8>]) -> libc::ssize_t {
 type GetRandomFn = unsafe extern "C" fn(*mut u8, libc::size_t, libc::c_uint) -> libc::ssize_t;
 
 pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
+    fn link_getrandom() -> *mut c_void {
+        static NAME: &[u8] = b"getrandom\0";
+        unsafe { libc::dlsym(libc::RTLD_DEFAULT, NAME.as_ptr() as *const _) }
```
josephlr, line 30: We can have this cast be explict `NAME.as_ptr() as *const libc::c_char`

src/netbsd.rs
```diff
@@ -28,8 +25,14 @@ fn kern_arnd(buf: &mut [MaybeUninit<u8>]) -> libc::ssize_t {
...
+    }
+
     // getrandom(2) was introduced in NetBSD 10.0
-    static GETRANDOM: Weak = unsafe { Weak::new("getrandom\0") };
+    static GETRANDOM: Weak = Weak::new(link_getrandom);
```
josephlr, line 34: It might be a little more readable to have these `static`s outside of `getrandom_inner`, like:

```rust
type GetRandomFn = unsafe extern "C" fn(*mut u8, libc::size_t, libc::c_uint) -> libc::ssize_t;
static NAME: &[u8] = b"getrandom\0";
static GETRANDOM: Weak = Weak::new(|| unsafe { libc::dlsym(libc::RTLD_DEFAULT, NAME.as_ptr() as *const _) });
```

having them all together makes the code a little easier to check.

newpavlov:
@josephlr 
I've tweaked the code a bit and moved linking function to the `get` method (previously `ptr`). I think it's a slightly better approach since we do not need redundant store of the function pointer. Theoretically this API could be misused by calling `get` with different functions, but since it's a private API, it should not cause any issues.

briansmith:
It would be good for somebody to actually test this and any future refactorings of `Weak` on a NetBSD machine.

josephlr (changes requested):
I like this implementation. One final question, should this now live in lazy.rs as its now quite similar to `LazyUsize`. If we do decide to move it, we should probably rename to `LazyPtr`.

Regarding @briansmith's comment:

> It would be good for somebody to actually test this and any future refactorings of `Weak` on a NetBSD machine.

Could we add a test that invokes this on Linux? Might be better than nothing. Could look like a `tests/linux_dlsym.rs`

```rust
// Only test on Linux/Android
#![cfg(any(target_os = "linux", target_os = "android"))]

#[path = "../src/lazy.rs"]
mod lazy;
use lazy::LazyPtr;

static GETRANDOM: LazyPtr = LazyPtr::new();

fn dlsym_getrandom() -> *mut c_void {
    static NAME: &[u8] = b"getrandom\0";
    let name_ptr = NAME.as_ptr() as *const libc::c_char;
    unsafe { libc::dlsym(libc::RTLD_DEFAULT, name_ptr) }
}

fn getrandom_impl(dest: &mut [u8]) -> Result<(), Error> {
    let fptr = GETRANDOM.get(dlsym_getrandom).unwrap();
    ...
}
mod common;
```

No need to add it if you don't think the test is maintainable or its too much work.

newpavlov:
>One final question, should this now live in lazy.rs as its now quite similar to LazyUsize. If we do decide to move it, we should probably rename to LazyPtr.

Good observation! Although there is a certain difference between `LazyUsize` and `LazyPtr` in the orderings used  under the hood.

src/lazy.rs
```diff
@@ -54,3 +58,57 @@ impl LazyBool {
...
+    // init(). Multiple callers can run their init() functions in parallel.
+    // init() should always return the same value, if it succeeds.
+    pub fn unsync_init(&self, f: impl Fn() -> *mut c_void) -> *mut c_void {
+        // Despite having only a single atomic variable (self.addr), we still
+        // cannot always use Ordering::Relaxed, as we need to make sure a
+        // successful call to `f` is "ordered before" any data read through
+        // the returned pointer (which occurs when the function is called).
+        // Our implementation mirrors that of the one in libstd, meaning that
+        // the use of non-Relaxed operations is probably unnecessary.
+        match self.addr.load(Ordering::Relaxed) {
```
newpavlov, line 101: Maybe we should use this code instead?
```rust
match self.addr.load(Ordering::Acquire) {
    Self::UNINIT => {
        let addr = f();
        self.addr.store(addr, Ordering::Release);
        addr
    }
    addr => addr,
}
```
I don't think "relaxed load + acquire fence" has advantages of simply doing "acquire load" in this case.  If anything, it may be even worse.

josephlr: Yeah I think doing the simpler acquire/release, is more obviously correct. I don't think we really gain anything in practice from using a relaxed load.

briansmith:
I do think having the test run on *-*-linux-gnu (where we know dlsym makes sense) is a good idea.

> ```
> let fptr = GETRANDOM.get(dlsym_getrandom).unwrap();
    ...
> ```

Note that we should actually call the function through `fptr`. We should have `LazyPtr::get()` do the pointer cast `transmute` at the end so that the test doesn't need to repeat it.

briansmith:
> I think we should have a plan for https://github.com/rust-random/getrandom/issues/285 first. It may be that we start using Weak on Linux too?

> Also, I think we should sort out what we're doing for Windows, 

It seems like for both of those, we're close to deciding now to use `LazyPtr`. If so, maybe we don't need to make this change, unless we want to make a change to improve the testing.

src/lazy.rs
```diff
@@ -54,3 +58,53 @@ impl LazyBool {
...
+
+impl LazyPtr {
+    /// A non-null pointer value which indicates we are uninitialized.
+    ///
+    /// This constant should ideally not be a valid pointer.
+    /// However, if by chance initialization function passed to the `get`
+    /// method does return UNINIT, there will not be undefined behavior.
+    /// The initialization function will just be called each time `get()`
+    /// is called. This would be inefficient, but correct.
+    const UNINIT: *mut c_void = !0usize as *mut c_void;
```
briansmith, line 82: Isn't `!0usize` more likely to be a valid pointer than 1 since at least on some targets the lowest word/byte/page is designed to trap when executed or dereferenced to catch bugs? I think if we want to make a change like this, we should do it in a separate PR.

newpavlov: It can not be a valid function pointer since such function will not be able contain even one instruction.

newpavlov:
>maybe we don't need to make this change, unless we want to make a change to improve the testing.

The result is cleaner and this approach does not need the `CStr` ersatz implemented in your other PR.

>I do think having the test run on --linux-gnu (where we know dlsym makes sense) is a good idea.

I would prefer to have a proper VM-based testing implemented in a separate PR.

src/lazy.rs
josephlr: SGTM, `!0` makes more sense for "Impossible function pointer value" than `1` (although either ever being used is vanishingly unlikely).

josephlr (approved):
Thanks!!

src/lazy.rs
```diff
@@ -54,3 +58,53 @@ impl LazyBool {
...
+/// except that the caller must manually cast self.ptr() to a function pointer.
+pub struct LazyPtr {
+    addr: AtomicPtr<c_void>,
+}
+
+impl LazyPtr {
+    /// A non-null pointer value which indicates we are uninitialized.
+    ///
+    /// This constant should ideally not be a valid pointer.
+    /// However, if by chance initialization function passed to the `get`
```
josephlr, line 78: Nit: "passed to the `unsync_init` function"

src/lazy.rs
```diff
@@ -54,3 +58,53 @@ impl LazyBool {
...
+    pub const fn new() -> Self {
+        Self {
+            addr: AtomicPtr::new(Self::UNINIT),
+        }
+    }
+
+    // Runs the init() function at most once, returning the value of some run of
+    // init(). Multiple callers can run their init() functions in parallel.
+    // init() should always return the same value, if it succeeds.
+    pub fn unsync_init(&self, f: impl Fn() -> *mut c_void) -> *mut c_void {
```
josephlr, line 94: `f` here should be `init`

josephlr (approved):
> I would prefer to have a proper VM-based testing implemented in a separate PR.

Sounds good to me.

PR #434, opened 2024-05-27
Tweak `util_libc::open_readonly`

newpavlov:
Based on discussion in #432, `open_readonly` now accepts `&[u8]` instead of `&str` for path and asserts that it contains at least one zero byte. This allows to make the function safe and compiler should be able to remove the assert since we use static paths.

### DIFF
--- a/src/util_libc.rs
+++ b/src/util_libc.rs
@@ -70,14 +70,20 @@ pub fn sys_fill_exact(
     Ok(())
 }
 
-// SAFETY: path must be null terminated, FD must be manually closed.
-pub unsafe fn open_readonly(path: &str) -> Result<libc::c_int, Error> {
-    debug_assert_eq!(path.as_bytes().last(), Some(&0));
+/// Open a file in read-only mode.
+///
+/// # Panics
+/// If `path` does not contain any zeros.
+#[inline(always)]
+pub fn open_readonly(path: &[u8]) -> Result<libc::c_int, Error> {
+    assert!(path.iter().any(|&b| b == 0));
     loop {
-        let fd = libc::open(
-            path.as_ptr().cast::<libc::c_char>(),
-            libc::O_RDONLY | libc::O_CLOEXEC,
-        );
+        let fd = unsafe {
+            libc::open(
+                path.as_ptr().cast::<libc::c_char>(),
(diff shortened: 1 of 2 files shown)


src/util_libc.rs
```diff
@@ -70,14 +70,20 @@ pub fn sys_fill_exact(
...
-// SAFETY: path must be null terminated, FD must be manually closed.
-pub unsafe fn open_readonly(path: &str) -> Result<libc::c_int, Error> {
-    debug_assert_eq!(path.as_bytes().last(), Some(&0));
+/// Open a file in read-only mode.
+///
+/// # Panics
+/// If `path` does not contain any zeros.
+#[inline(always)]
+pub fn open_readonly(path: &[u8]) -> Result<libc::c_int, Error> {
+    assert!(path.iter().any(|&b| b == 0));
```
briansmith, line 79: OK, but since this already returns a `Result`, we should avoid `assert!`. Either the compiler is going to optimize away the check completely anyway, or we'd be left with dead code that constructs an `Err`, or we'd be left with dead code that panics. The possibility of dead code that panics is problematic for some static analysis mechanisms that try to ensure there are no panics in a program.

josephlr: I think having an `assert!` here is fine. If this ever triggers, it's an error in our code, not an error in the user's calling environment or input. 

For `opt-level=1` (https://rust.godbolt.org/z/K3cjrsxqa) or `opt-level=s` (https://rust.godbolt.org/z/MT9jozE1e) , the assert is optimized away, and on `opt-level=0` there's a bunch of dead code that panics, even if the `assert` is replaced with a branch(https://rust.godbolt.org/z/939exhnKn). Leaving it as-is also makes things more readable/debuggable.

A better approach for this is documenting and testing that our implementation does not panic, I opened #435.

josephlr, line 79: Add a comment for us to move to [`CStr::from_bytes_until_nul`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_until_nul) when the MSRV is high enough:
```rust 
// TODO(MSRV 1.69): Move to CStr::from_bytes_until_nul
```

josephlr (approved):
Looks good!

PR #466, opened 2024-06-07
Cleanup Auxiliary Linux tests

josephlr:
This PR merges the "extra" Linux tests into a single job. This reduces the number of jobs we have to run as part of the CI. We also explictly install libc and libgcc (to make it clear these are the only extra dev dependancies).

I also added building/linking of the x32 target. As noted in #464, this will not run in the current github actions runner due to an incompatible kernel config. CC @briansmith

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -71,20 +71,26 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        target: [
-          x86_64-unknown-linux-musl,
-          i686-unknown-linux-gnu,
-          i686-unknown-linux-musl,
-        ]
+        target: [x86_64-unknown-linux-musl, i686-unknown-linux-musl]
+        include:
+          - target: i686-unknown-linux-gnu
+            packages: libc6-dev-i386 lib32gcc-11-dev
+          - target: x86_64-unknown-linux-gnux32
+            packages: libc6-dev-x32 libx32gcc-11-dev
+            # TODO: Find a Linux image/runner with CONFIG_X86_X32_ABI set
+            cargo_test_opts: --no-run
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: ${{ matrix.target }}
-      - name: Install multilib
-        run: sudo apt-get update && sudo apt-get install gcc-multilib
+      - name: Install libc and libgcc
+        if: matrix.packages
+        run: |
+          sudo apt-get update
+          sudo apt-get install --no-install-recommends ${{ matrix.packages }}
       - uses: Swatinem/rust-cache@v2
-      - run: cargo test --target=${{ matrix.target }} --features=std
+      - run: cargo test ${{ matrix.cargo_test_opts }} --target=${{ matrix.target }} --features=std
 
   ios-tests:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -69,22 +69,21 @@ jobs:
...
-          i686-unknown-linux-musl,
-        ]
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
         with:
-          targets: ${{ matrix.target }}
-      - name: Install multilib
-        run: sudo apt-get update && sudo apt-get install gcc-multilib
+          targets: x86_64-unknown-linux-musl, i686-unknown-linux-gnu, i686-unknown-linux-musl, x86_64-unknown-linux-gnux32
```
briansmith, line 76: I do not like replacing multiple separate jobs with a single job that mixes targets. I think we should keep the matrix like before, and add a variable `packages` to each specific target:
```
        include:
          - target: x86_64-unknown-linux-gnux32
            cargo_test_options: --no-run --features=rdrand
            packages: "libc6-dev-x32 libx32gcc-11-dev"
...

- if ${{ matrix.packages != "" }}
  run sudo apt-get install ${{ matrix.packages }}
...
- cargo test ${{ matrix.cargo_test_options }} --features=std
```

.github/workflows/tests.yml
```diff
@@ -69,22 +69,21 @@ jobs:
...
+        run: |
+          sudo apt-get update
+          sudo apt-get install --no-install-recommends libc6-dev-i386 libc6-dev-x32 lib32gcc-11-dev libx32gcc-11-dev
       - uses: Swatinem/rust-cache@v2
-      - run: cargo test --target=${{ matrix.target }} --features=std
+      - run: cargo test --target=x86_64-unknown-linux-musl --features=std
+      - run: cargo test --target=i686-unknown-linux-gnu --features=std
+      - run: cargo test --target=i686-unknown-linux-musl --features=std
+      # TODO: Find a Linux image/runner with CONFIG_X86_X32_ABI set
+      - run: cargo test --no-run --target=x86_64-unknown-linux-gnux32 --features=std
```
briansmith, line 86: We should also include the rdrand feature for the x86_64 targets, especially x32.

briansmith (commented):
>  This reduces the number of jobs we have to run as part of the CI. 

Why is that a goal?

josephlr:
> Why is that a goal?

Initially I thought the time was dominated by installing the deps, but after rerunning some tests, this isn't true. Spllitting back into multiple jobs.

.github/workflows/tests.yml
josephlr: Nice idea, this setup also makes it clear that only the `gnu` targets need these libs installed (which makes perfect sense).

Should be fixed now, let me know what you think.

.github/workflows/tests.yml
josephlr: Note that the `rdrand` feature doesn't do anything on supported targets. However, we do test the rdrand implementation on all x86/x86_64 targets (see `tests/rdrand.rs`). See how [this run](https://github.com/rust-random/getrandom/actions/runs/9423752238/job/25962770313) runs `tests/rdrand.rs`.

briansmith (approved):

Issue #401, opened 2024-02-26
Port from libc to rustix


notgull:
rustix is a crate which provides I/O-safe wrappers around raw system calls on Linux and libc on other Unixes. By using it we can use raw syscalls instead of libc calls, which has a number of performance, safety and code size advantages.

I am wiling to write a PR for this; however there are a couple of items of discussions we have to go over first.

- This will bump this crate's MSRV to v1.63. Is this acceptable?
- It will also add a couple of (mostly inconsequential) dependencies. Is this acceptable?

newpavlov:
I considered it (personally, I like `rustix` and use it instead of `libc` in my projects), but there are several issues:
- MSRV 1.63 is a bit too high for us.
- Some users are quite sensitive to additional dependencies, regardless if they are "mostly inconsequential" or not.
- `rustix::rand::getrandom` accepts `&mut [u8]`, while we provide API which works over `&mut [MaybeUninit<u8>]`.

So we probably will not be considering migrating to `rustix` until hypothetical `getrandom v0.3`.

notgull:
For your last point, [`getrandom_uninit`](https://docs.rs/rustix/latest/rustix/rand/fn.getrandom_uninit.html) is exposed in `rustix` now.

For the rest, I agree. We’ll have to wait.

newpavlov:
>`getrandom_uninit` is exposed in `rustix` now.

Ah, good to know!

josephlr:
@notgull is it the intent of `rustix` to support all `target_family = unix` Rust targets? We try to support every rust target, and I would be a pain to deal with _both_ `rustix` and `libc` crates.

We could maybe switch just the `use_file.rs` code to use `rustix`, which would require Linux, Android, Haiku, Redox, NTO and AIX support.

newpavlov:
I think it makes sense to use `rustix` only on Linux for the raw syscall feature. We probably can limit it's use only to the `getrandom` path and make it a (disabled by default) crate feature.

notgull:
> @notgull is it the intent of `rustix` to support all `target_family = unix` Rust targets? We try to support every rust target, and I would be a pain to deal with _both_ `rustix` and `libc` crates.

`rustix` supports all of the Unix targets, as well as some of the other targets, like Windows (through winsock).

> We could maybe switch just the `use_file.rs` code to use `rustix`, which would require Linux, Android, Haiku, Redox, NTO and AIX support.

Yes, I think this would work.

[...]

PR #470, opened 2024-06-08
Add rustix backend

notgull:
This commit adds a backend based on the rustix crate. The main advantage
of rustix, in addition to greater safety, is that it uses raw syscalls
instead of going through libc.

I've tried to modify the existing libc code as little as possible, in
order to make this change as auditable as possible. I haven't touched
the pthreads code yet, as that's a little tricky.

This exists for discussion purposes.

cc #401

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,14 +18,14 @@ cfg-if = "1"
 compiler_builtins = { version = "0.1", optional = true }
 core = { version = "1.0", optional = true, package = "rustc-std-workspace-core" }
 
-[target.'cfg(unix)'.dependencies]
+[target.'cfg(all(unix, not(target_os = "linux")))'.dependencies]
 libc = { version = "0.2.154", default-features = false }
 
 [target.'cfg(target_os = "wasi")'.dependencies]
 wasi = { version = "0.11", default-features = false }
 
-[target.'cfg(any(target_os = "linux", target_os = "android"))'.dependencies]
-rustix = { version = "0.38.0", features = ["event", "fs", "rand", "thread"], default-features = false, optional = true }
+[target.'cfg(target_os = "linux")'.dependencies]
+rustix = { version = "0.38.0", features = ["event", "fs", "rand", "thread"], default-features = false }
 
 [target.'cfg(all(any(target_arch = "wasm32", target_arch = "wasm64"), target_os = "unknown"))'.dependencies]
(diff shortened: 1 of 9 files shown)


src/use_file.rs
```diff
@@ -93,6 +115,25 @@ fn wait_until_rng_ready() -> Result<(), Error> {
...
+    loop {
+        match event::poll(&mut pfd, -1) {
+            Ok(_) => return Ok(()),
+            Err(rustix::io::Errno::INTR) => continue,
+            Err(err) => return Err(crate::util_rustix::cvt(err)),
+        }
+    }
+}
+
 struct Mutex(UnsafeCell<libc::pthread_mutex_t>);
```
briansmith, line 137: The most important improvement rustix could make to this code would be replacing the pthreads_mutex_t here with whatever is used by libstd on the target (e.g. futex on Linux/Android).

notgull: Good point. I've replaced `Mutex` with one based on futexes, only on Rustix.

Cargo.toml
```diff
@@ -18,14 +18,14 @@ cfg-if = "1"
...
+[target.'cfg(all(unix, not(target_os = "linux")))'.dependencies]
 libc = { version = "0.2.154", default-features = false }
 
 [target.'cfg(target_os = "wasi")'.dependencies]
 wasi = { version = "0.11", default-features = false }
 
-[target.'cfg(any(target_os = "linux", target_os = "android"))'.dependencies]
-rustix = { version = "0.38.0", features = ["event", "fs", "rand", "thread"], default-features = false, optional = true }
+[target.'cfg(target_os = "linux")'.dependencies]
+rustix = { version = "0.38.0", features = ["event", "fs", "rand", "thread"], default-features = false }
```
briansmith, line 28: We should only need `event`, `fs`, and `thread` if we're going to use the file-based fallback, but unfortunately that's very hard to specify in Cargo.toml, as you can see by the actual cfg logic in lib.rs.

I guess there must be a reason why Rustix makes those things optional features. (Why?) Would people who don't want those features activated care that `getrandom` would be activating them?

notgull: I don't think so, the features are mostly to reduce code size.

newpavlov:
This PR is outdated, so I will close it. We probably will revisit the `rustix` question while adding support for `none` Linux targets.

PR #471, opened 2024-06-10
Cleanup Testing code

josephlr:
Depends on #473

This stops using weird include hacks to reuse testing code. Instead, we move the code to a normal `tests.rs` file and uses helper functions to avoid code duplication. Now when running `cargo test` the tests appear as normal unit tests:
```
   Compiling getrandom v0.2.15 (/home/joe/dev/rust/getrandom)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.23s
     Running unittests src/lib.rs (target/debug/deps/getrandom-1d050696e8dd96df)

running 6 tests
test tests::fill_large ... ok
test tests::fill_zero ... ok
test tests::fill_small ... ok
test tests::fill_huge ... ok
```

Follow up PRs (which explain some of the design logic in this PR):
  - Adds back RDRAND testing: #476 
  - Tests `getrandom_uninit`: #474 
  - Tests file-based fallback: #475


josephlr:
CC @briansmith who is doing other testing improvements.

josephlr:
Haiku failure is unrelated, fixed by from https://github.com/rust-lang/rust/pull/126212

src/lib.rs
```diff
@@ -216,6 +216,15 @@ use crate::util::{slice_as_uninit_mut, slice_assume_init_mut};
...
 mod error;
+// Conditionally import lazy, as not all targets support atomics.
+#[cfg(any(
+    target_os = "android",
+    target_os = "linux",
+    target_os = "netbsd",
+    target_arch = "x86_64",
+    target_arch = "x86"
+))]
+mod lazy;
```
newpavlov, line 227: I don't like this change. Maybe it's better to use the following code in `rdrand`?
```rust
#[path = "lazy.rs"]
mod lazy;
```

newpavlov (approved):

src/rdrand.rs
```diff
@@ -121,3 +121,23 @@ unsafe fn rdrand_exact(dest: &mut [MaybeUninit<u8>]) -> Option<()> {
...
+#[cfg(test)]
+mod tests {
+    use crate::{tests::*, *};
+
+    fn rdrand_fill_uninit(dest: &mut [MaybeUninit<u8>]) -> Result<&mut [u8], Error> {
+        super::getrandom_inner(dest)?;
+        Ok(unsafe { slice_assume_init_mut(dest) })
+    }
+
+    #[test]
```
briansmith, line 134: I support the general approach, I don't think it makes sense to manually repeat these tests in each module. We should have something that ensures the full suite of tests (more-or-less) is run for each implementation.

I have done this in other projects:
```
macro_rules! define_tests {
    ( $fill_init:path, ... ) => {
        #[test]
        fn test_small() { 
            tests::test_small($fill_uninit);
        }
       #[test]
      fn test_empty() { tests::test_empty($fill_uniit) }
   }
}
```
Then in each implementation (rdrand.rs, use_file.rs, etc.):
```
mod tests;
tests::define_tests!(rdrand_fill_uninit)

fn fill_uninit(dest: &mut [MaybeUninit<u8>]) -> Result<&mut [u8], Error> { ... }
```

Otherwise, we'd have to decide for each implementation whether we should copy/paste code into it, and naturally we'd gravitate towards less comprehensive testing of each implementation, which is the opposite of what we want--we want more comprehensive testing to be easier, not harder.

src/tests.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    // start all the tasks
+    for tx in txs.iter() {
+        tx.send(()).unwrap();
+    }
+}
+
+// A simple Linear Congruential Generator (LCG) that passes our RNG tests.
+const A: u32 = 134775813;
```
briansmith, line 121: I don't think we should spend effort on a custom implementation that passes the tests. The fact that this implementation passes the current test suite is a defect in the test suite that can and should be fixed.

Also, we're not testing a custom implementation; we're testing the mechanics of the custom implementation registration and how we deal with the unique situations that a custom implementation gives us.

What we'll likely need is a mock implementation that we can programatically control the behavior of.

src/tests.rs
```diff
@@ -0,0 +1,176 @@
...
+#[test]
+fn custom_not_used() {
+    getrandom(&mut [0; 1337]).unwrap();
+}
+
+// On an unsupported platform, make sure the custom implementation is used.
+#[cfg(all(feature = "custom", target_arch = "wasm32", target_os = "unknown"))]
+#[test]
+fn custom_used() {
+    let err = getrandom(&mut [0; 1337]).unwrap_err();
```
briansmith, line 167: In this kind of test, we should just use a mock custom implementation that has state that we can directly query (e.g. using Mutex-protected global state), instead of indirectly interpreting its output.

briansmith (changes requested):
This PR tries to do three things, AFAICT:
* Refactoring how common logic is shared between tests.
* Add support for testing the `_uninit()` API directly.
* Change how we test custom implementations and what is being tested in those cases.

Each of these types of changes should be its own PR with its own discussion. I think if I had submitted this PR, nobody would want to review it because it mixes all of these things.

src/lib.rs
briansmith: I think in general we should probably try to reduce the `#[path]` stuff that's being done. I think we could lust leave `lazy` the way it currently it, or move these changes to a separate PR, since I don't see how they are related to the testing.

newpavlov: It's related to the testing changes because of the `rdrand` module. It requires `lazy` and this PR mounts `rdrand` for x86(-64) targets during testing. Without doing this and leaving `rdrand` unchanges we would get `lazy` mounted twice during testing.

josephlr:
> This PR tries to do three things, AFAICT:
> 
>     * Refactoring how common logic is shared between tests.
> 
>     * Add support for testing the `_uninit()` API directly.
> 
>     * Change how we test custom implementations and what is being tested in those cases.
> 
> 
> Each of these types of changes should be its own PR with its own discussion. I think if I had submitted this PR, nobody would want to review it because it mixes all of these things.

This is a very good point. I've reduced the scope of this PR (see the updated description), and I've split out some of the prerequisites into their own PRs (#473, #472). I've also moved testing of RDRAND / `getrandom_uninit` / file-fallback to followup PRs (#476 , #475 , #474).

@newpavlov @briansmith let me know if you thing the `Byte`-based generic code makes sense. Having a single method which is "give me a random `Vec<u8>`" seemed like the easiest way to express the differences between the `u8` and `MaybeUninit<u8>` code.

src/tests.rs
josephlr: I moved the custom RNG changes to #473 and simplified the implementation there. It doesn't use a mock (I couldn't figure out a clean way to do that in a test).

src/tests.rs
josephlr: This is a good point, I agree. The RNG quality tests now don't run if the `"custom"` feature is enabled. Other changes to the testing of the custom RNG are moved to #473.

src/rdrand.rs
josephlr: Nice, I like this. Added a `define_tests` macro which makes reusing the tests much easier (see #476 #475 #474).

src/lib.rs
josephlr: Lazy changes moved to #472 which then avoids us needing to change any imports for the testing PRs.

newpavlov:
I would prefer to keep bulk of our tests in the `tests/` folder. I will try to experiment with the testing code a bit later.

src/lib.rs
```diff
@@ -404,3 +404,10 @@ pub fn getrandom_uninit(dest: &mut [MaybeUninit<u8>]) -> Result<&mut [u8], Error
     // since it returned `Ok`.
     Ok(unsafe { slice_assume_init_mut(dest) })
 }
+
+// Don't run normal unit tests when testing custom getrandom
+#[cfg(all(
+    test,
+    not(all(target_family = "wasm", target_os = "unknown", feature = "custom"))
+))]
+pub(crate) mod tests;
```
briansmith, line 413: I think the "normal" tests should still be defined in tests/normal.rs (See comment below).

I think we're all keen to avoid duplicating complex `cfg` all over. I suggest we add a new api:
```
/// true if the implementation defined by `register_custom_getrandom!` would be used; false if it will be ignored.
pub const CUSTOM_USED: bool;
```
Then we can condition all the entropy checks in the common tests on that new API.

I think this API would also be useful to external users.

I also expected to see here something like:
```
cfg_if::cfg_if! {
    if #[cfg(any(target_arch = "x86", target_arch = "x86_64")] {
        mod rdrand;
        #[cfg(test)]
        mod rdrand_tests {
            tests::define_tests!(rdrand::getrandom_uninit);
        }
    }
}
```
This way, tests/rdrand.rs is properly replaced instead of just being lost.

Then presumably in future PRs we would do the same for use_file, etc.

src/tests.rs
```diff
@@ -0,0 +1,117 @@
...
+fn num_diff_bits(s1: &[u8], s2: &[u8]) -> usize {
+    assert_eq!(s1.len(), s2.len());
+    s1.iter()
+        .zip(s2.iter())
+        .map(|(a, b)| (a ^ b).count_ones() as usize)
+        .sum()
+}
+
+// A function which fills a buffer with random bytes.
+type FillFn<B> = fn(&mut [B]) -> Result<(), Error>;
```
briansmith, line 19: I think this should be type `FillFn<B, R> = fn(&mut [B]) -> Result<R, Error>` so that all of `getrandom::getrandom_uninit`, `getrandom::getrandom`, and the internal `getrandom_impl` functions can all be used with it.

src/tests.rs
```diff
@@ -0,0 +1,117 @@
...
+        #[test]
+        #[cfg_attr(target_family = "wasm", ignore)]
+        fn multithreading() {
+            crate::tests::check_multithreading($fill)
+        }
+    };
+}
+pub(crate) use define_tests;
+
+define_tests!(crate::getrandom);
```
briansmith, line 117: I would expect that instead of having this line here, we could have it in tests/normal.rs:
```
#[path = "../src/tests.rs"]
mod tests;

mod init {
    define_tests!(getrandom::getrandom);
}
mod uninit {
    define_tests!(getrandom::getrandom_uninit);
}
```
This way we're defining integration tests in the normal way.

briansmith (changes requested):
This PR is hard to reivew in the GitHub UI because the move of tests/common/mod.rs to src/tests.rs is combined with extensive edits of the file. If you could, please split moving the unmodified file into a separate commit and then rebase this commit over it, so that the PR consists of two commits. Then each commit in the PR can be reviewed in the GitHub UI individually.

src/lib.rs
josephlr: > I think we're all keen to avoid duplicating complex `cfg` all over. I suggest we add a new api:

After rearranging some stuff, this `cfg` is no longer needed. We can just do `#[cfg(test)]` and everything works as expected. Note that the CI tests are updated to run `wasm-pack test --node --features=custom --test=custom` when testing on an unsupported platform. This makes more sense to me, as the in-library unit tests will not build if you just enable the `custom` feature on an unsupported platform (you have to register a custom handler, which is only done in our integration test).



> I also expected to see here something like:
> 
> ```
> cfg_if::cfg_if! {
>     if #[cfg(any(target_arch = "x86", target_arch = "x86_64")] {
>         mod rdrand;
>         #[cfg(test)]
>         mod rdrand_tests {
>             tests::define_tests!(rdrand::getrandom_uninit);
>         }
>     }
> }
> ```
>
> This way, tests/rdrand.rs is properly replaced instead of just being lost.

I changed this PR to just unconditionally have `mod rdrand` present if the `"rdrand"` feature is enabled. This is similar to what we do for `"custom"`, and allows us to just have normal unit tests inside `rdrand.rs`. We can't put such things inside the `cfg_if`, but having the `mod rdrand` outside of that macro ends up being more readable anyway.

josephlr:
> This PR is hard to reivew in the GitHub UI because the move of tests/common/mod.rs to src/tests.rs is combined with extensive edits of the file. If you could, please split moving the unmodified file into a separate commit and then rebase this commit over it, so that the PR consists of two commits. Then each commit in the PR can be reviewed in the GitHub UI individually.

Split into 3 commits:
  - Rename/Delete files
  - Common tests for `getrandom` and `getrandom_uninit`
  - Restoring RDRAND testing

I also closed #474 and #476 as those changes seem to work better as part of this PR (#475 is still its own PR).

src/tests.rs
josephlr: Conceptually, I consider the tests created by `define_tests!` to be unit tests rather than integration tests (they just happen to test the public API), so it makes more sense to me for them to use the unit testing framework. Having them as unit tests also avoids `#[path = "..."]` shenanigans (which can be brittle and annoying at times).

src/tests.rs
josephlr: Unfortunately, this doesn't work. Due to lifetime-related problems, there's no way to have a type alias which refers to the type of all 3 of those functions. Specifically, there's no `R` for which:
```rust
getrandom_uninit: FillFn<B,R>
```
as the full type is:
```rust
getrandom_uninit: for<'a> fn(&'a mut [MaybeUninit<u8>]) -> Result<&'a mut [u8], Error>
```

Instead, I changed things to have `FillFn` be a trait (implemented by the relevant `Fn` objects) which defines a `make_vec` method. The only downside is that it requires a dummy generic parameter to avoid conflicting implementations. Let me know what you think.

src/tests.rs
```diff
@@ -0,0 +1,139 @@
...
+
+macro_rules! define_tests {
+    ($fill:path) => {
+        use crate::tests::FillFn;
+        #[cfg(all(target_family = "wasm", target_os = "unknown"))]
+        use wasm_bindgen_test::wasm_bindgen_test as test;
+
+        #[test]
+        fn zero() {
+            $fill.make_vec(0);
```
briansmith, line 110: `getrandom::getrandom_uninit()` does the `!dest.is_empty()` check before to avoid calling `getrandom_inner` with an empty `dest`, `custom::getrandom_inner()` doesn't have a similar check, but also we guarantee that the custom implementation will never be called with a zero-length `dest`. But, having this test here implies to me that we do expect every `getrandom_inner()` to correctly handle an empty `dest`. It seems contradictory. Perhaps instead `zero()` should be an integration test that tests specifically/only `getrandom::getrandom()` and `getrandom::getrandom_uninit()`? Or, perhaps we should move that check out of `getrandom_uninit` and into the `getrandom_inner` implementations since that would be more robust.

Regardless of how we address this, we should have a comment here explaining what we're expecting this test to be testing.

src/tests.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+        #[test]
```
briansmith, line 112: Let's separate the functions inside the macro body with blank lines.

src/tests.rs
```diff
@@ -0,0 +1,139 @@
...
+        fn diff_large() {
+            crate::tests::check_diff_large(|len| $fill.make_vec(len));
+        }
+        #[test]
+        fn small() {
+            crate::tests::check_small(|len| $fill.make_vec(len));
+        }
+        #[test]
+        fn huge() {
+            $fill.make_vec(100_000);
```
briansmith, line 122: We need a comment here about why were not using check_diff_large

briansmith (changes requested):
The main objection I still have is moving all these tests from integration tests to unit tests. The `#[path]` stuff is suboptimal and we should reduce our usage of it. But, I think we should also still stick to preferring having integration tests for `getrandom::getrandom` and `getrandom::getrandom_uninit` and more generally the whole public API.

newpavlov:
I think #504 has resolved most of the issues targeted by this PR.

PR #472, opened 2024-06-12
MSRV 1.60: Conditionally import lasy.rs module

josephlr:
This simplifies lib.rs and allows for our testing code to be less convoluted. Split out from #471. CC @briansmith 

The MSRV for rand 0.9 is 1.61 and Debian stable is 1.63, so this is a fine MSRV bump.


briansmith (approved):

briansmith:
Note typo in commit message: lasy -> lazy.

josephlr:
> Note typo in commit message: lasy -> lazy.

Fixed in the merged commit message.

PR #473, opened 2024-06-12
Cleanup Custom Tests

josephlr:
This moves the tests for the custom RNG registration into custom.rs. It also:
  - Simplifies the registered custom RNG
  - Makes sure the custom RNG _is not_ used on supported platforms
  - Makes sure the custom RNG _is_ used on unsupported platforms

Split out from #471 CC @briansmith

### DIFF
--- a/tests/custom.rs
+++ b/tests/custom.rs
@@ -1,54 +1,52 @@
-// Test that a custom handler works on wasm32-unknown-unknown
-#![cfg(all(
-    target_arch = "wasm32",
-    target_os = "unknown",
-    feature = "custom",
-    not(feature = "js")
-))]
-
-use wasm_bindgen_test::wasm_bindgen_test as test;
-
 use core::num::NonZeroU32;
 use getrandom::{getrandom, register_custom_getrandom, Error};
+#[cfg(all(target_family = "wasm", target_os = "unknown"))]
+use wasm_bindgen_test::wasm_bindgen_test as test;
 
-fn len7_err() -> Error {
-    NonZeroU32::new(Error::INTERNAL_START + 7).unwrap().into()
-}
-
-fn super_insecure_rng(buf: &mut [u8]) -> Result<(), Error> {
+const LEN7_CODE: u32 = Error::CUSTOM_START + 7;
+// Returns a custom error if input is length 7, otherwise fills with 0x55.
+fn mock_rng(buf: &mut [u8]) -> Result<(), Error> {
     // `getrandom` guarantees it will not call any implementation if the output
     // buffer is empty.
     assert!(!buf.is_empty());
-    // Length 7 buffers return a custom error
     if buf.len() == 7 {
-        return Err(len7_err());
+        return Err(NonZeroU32::new(LEN7_CODE).unwrap().into());
     }
-    // Otherwise, fill bytes based on input length
-    let mut start = buf.len() as u8;
-    for b in buf {
-        *b = start;
-        start = start.wrapping_mul(3);
-    }
+    buf.fill(0x55);
     Ok(())
 }
 
-register_custom_getrandom!(super_insecure_rng);
+// Test registering a custom implementation, even on supported platforms.
+register_custom_getrandom!(mock_rng);
 
-use getrandom::getrandom as getrandom_impl;
-mod common;
-
+// Invoking with an empty buffer should never call the custom implementation.
 #[test]
-fn custom_rng_output() {
-    let mut buf = [0u8; 4];
-    assert_eq!(getrandom(&mut buf), Ok(()));
-    assert_eq!(buf, [4, 12, 36, 108]);
-
-    let mut buf = [0u8; 3];
-    assert_eq!(getrandom(&mut buf), Ok(()));
-    assert_eq!(buf, [3, 9, 27]);
+fn custom_empty() {
+    getrandom(&mut []).unwrap();
 }
 
+// On a supported platform, make sure the custom implementation isn't used. We
+// test on a few common platfroms, rather than duplicating the lib.rs logic.
+#[cfg(any(
+    target_os = "linux",
+    target_os = "windows",
+    target_os = "macos",
+    target_os = "espidf",
+    target_os = "wasi",
+    all(target_family = "wasm", target_os = "unknown", feature = "js"),
+))]
 #[test]
-fn rng_err_output() {
-    assert_eq!(getrandom(&mut [0; 7]), Err(len7_err()));
+fn custom_not_used() {
+    getrandom(&mut [0; 7]).unwrap();
+}
+// On an unsupported platform, make sure the custom implementation is used.
+#[cfg(all(target_family = "wasm", target_os = "unknown", not(feature = "js")))]
+#[test]
+fn custom_used() {
+    let err = getrandom(&mut [0; 7]).unwrap_err();
+    assert_eq!(err.code().get(), LEN7_CODE);
+
+    let mut buf = [0; 12];
+    getrandom(&mut buf).unwrap();
+    assert_eq!(buf, [0x55; 12]);
 }
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,9 +38,9 @@ jobs:
           toolchain: ${{ matrix.toolchain }}
       - uses: Swatinem/rust-cache@v2
       - run: cargo test
-      - run: cargo test --features=std
+      # Make sure enabling the std and custom features don't break anything
+      - run: cargo test --features=std,custom
       - run: cargo test --features=linux_disable_fallback
-      - run: cargo test --features=custom # custom should do nothing here
       - if: ${{ matrix.toolchain == 'nightly' }}
         run: cargo test --benches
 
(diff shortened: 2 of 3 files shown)


src/custom.rs
```diff
@@ -103,3 +103,39 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
...
+        }
+        dest.fill(0x55);
+        Ok(())
+    }
+
+    // Test registering a custom implementation, even on supported platforms.
+    register_custom_getrandom!(insecure_rng);
+
+    // On a supported platform, make sure the custom implementation isn't used.
+    #[cfg(target_os = "linux")]
```
briansmith, line 125: I suggest use use:
```
cfg_if! {
    if #[cfg(we should be using custom implementation)] {
         #[test]
         fn custom_used() { ... }
    } else { 
        #[test]
         fn custom_not_used() { .... }
    }
}
```
In particular, we should run custom_not_used on all targets, not just Linux, right?

src/custom.rs
```diff
@@ -103,3 +103,39 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const LEN7_CODE: u32 = Error::CUSTOM_START + 7;
+    // Returns a custom error if input is length 7, otherwise fills with 0x55.
+    fn insecure_rng(dest: &mut [u8]) -> Result<(), Error> {
```
briansmith, line 113: I suggest `mock_` instead of `insecure_`.

src/custom.rs
```diff
@@ -103,3 +103,39 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
         Some(code) => Err(Error::from(code)),
     }
 }
+
+#[cfg(test)]
+mod tests {
```
briansmith, line 108: Shouldn't these tests be defined in `tests/custom.rs` like before, since it is exercising the semantics of the public API? This way it would serve as a test that, for example, the register...! macro is defined on all targets and not just on ones for which we actually use custom implementations.

tests/custom.rs
```diff
@@ -1,54 +0,0 @@
...
-use getrandom::{getrandom, register_custom_getrandom, Error};
-
-fn len7_err() -> Error {
-    NonZeroU32::new(Error::INTERNAL_START + 7).unwrap().into()
-}
-
-fn super_insecure_rng(buf: &mut [u8]) -> Result<(), Error> {
-    // `getrandom` guarantees it will not call any implementation if the output
-    // buffer is empty.
-    assert!(!buf.is_empty());
```
briansmith, line 21: I think it is useful to have a test that calls `getrandom::getrandom[_uninit()]` with an empty `dest`, and then also add this assertion to our insecure/mock custom implementation.

briansmith (changes requested):

src/custom.rs
briansmith: Not really a big deal since it was already named `insecure_` before.

newpavlov (commented):
This looks fine to me, but I am not sure we need to specifically test that custom implementations do not overwrite the default backends.

Also, I have an alternative proposal. I think we can do the following:
```rust
// in lib.rs
cfg_if! {
    if #[cfg(all(test, test_rdrand))] {
        mod lazy;
    } else if #[cfg(all(test, test_custom, feature = "custom"))] {
        use custom as imp;
    } else if #[cfg( ... ))] {
        // ...
    }
}

// in tests module:
register_custom_getrandom!(mock_rng);

#[cfg(not(all(feature = "custom", test_custom)))]
#[test]
fn mock_rng_is_not_used() {
    // ..
}
```
This will eliminate the weird juggling around RDRAND and will allow us to test a custom implementation on standard Linux targets using `cfg(test)` will ensure that the testing branches will not escape outside of this crate.

src/custom.rs
briansmith: > Shouldn't these tests be defined in tests/custom.rs.

I see now that your intent is to be able to run the tests on Linux (at least), where we don't enable the API to work, normally.

I think the best thing to do is to keep the tests/custom.rs integration test, but also have a unit test that can run on Linux. Perhaps this can be the one place we use "#[path]` tricks in the testing? Basically we'd have a "we expect custom to be ignored" test file `custom_ignored.rs` and a custom-used test file `custom_used.rs`, and in tests/* we'd select custom-ignored or custom-used based on the target OS, and in src/ we'd use `custom_used.rs` so that we can run the tests on targets for which we normally don't enable the custom feature?

Though I don't even know that that complexity is necessary because we already do have a good way to run the custom tests, via the websasembly CI runners. So I also think we could just have the custom tests only as integration tests (in `tests/` only) and test the "custom used" configuration only in webassembly for now.

josephlr:
@newpavlov @briansmith  PTAL. I moved the custom.rs tests back to being an integration test, so both supported and unsupported platforms are now tested without changing the `cfg_if!` in `lib.rs`.

> This looks fine to me, but I am not sure we need to specifically test that custom implementations do not overwrite the default backends.
> 
> Also, I have an alternative proposal. I think we can do the following:

I think its important to test that:
  - The custom backend works
  - The custom backend won't override existing implementations

 My concern with changing the `lib.rs` logic is that by adding too many `test` cfgs, we might end up not actually testing what we want.

> This will eliminate the weird juggling around RDRAND and will allow us to test a custom implementation on standard Linux targets. Using `cfg(test)` will ensure that the testing branches will not escape outside of this crate.

I agree that RDRAND has some weird juggling, but that's probably best left for

tests/custom.rs
josephlr: Added, good point.

src/custom.rs
josephlr: Good point. The main purposes of these tests are that a custom implementation can be registered _outside_ of this crate. So for this reason, I moved the `custom.rs` tests back to being integration tests. Given that we can run tests on unsupported platforms (via wasm CI runners), I agree that any additional complexity w.r.t. testing is unnecessary.

src/custom.rs
josephlr: Renamed to `mock_rng` to better express our intent.

src/custom.rs
josephlr: I added a few additional OS targets (Windows and macOS) I don't think it's necessary to run them on _every_ supported target, checking a few common platforms seems fine.

tests/custom.rs
```diff
@@ -1,54 +1,43 @@
...
-    let mut buf = [0u8; 3];
-    assert_eq!(getrandom(&mut buf), Ok(()));
-    assert_eq!(buf, [3, 9, 27]);
+fn custom_empty() {
+    getrandom(&mut []).unwrap();
 }
 
+// On a supported platform, make sure the custom implementation isn't used. We
+// test on a few common platfroms, rather than duplicating the lib.rs logic.
+#[cfg(any(target_os = "linux", target_os = "windows", target_os = "macos"))]
```
briansmith, line 28: I would add `all(target_arch = "wasm32", target_os = "unknown", feature = "js")` to this so we know what happens when both "js" and "custom" are enabled.

tests/custom.rs
```diff
@@ -1,54 +1,43 @@
...
 #[test]
-fn rng_err_output() {
-    assert_eq!(getrandom(&mut [0; 7]), Err(len7_err()));
+fn custom_not_used() {
+    getrandom(&mut [0; 7]).unwrap();
+}
+// On an unsupported platform, make sure the custom implementation is used.
+#[cfg(all(target_arch = "wasm32", target_os = "unknown", not(feature = "js")))]
+#[wasm_bindgen_test::wasm_bindgen_test]
```
briansmith, line 35: I suggest we use this common idiom, at the top of the file:
```
#[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
use wasm_bindgen_test::wasm_bindgen_test as test;
```
This way, we can easily support tests that run in wasm-bindgen and libtests.

tests/custom.rs
```diff
@@ -1,54 +1,43 @@
-// Test that a custom handler works on wasm32-unknown-unknown
-#![cfg(all(
-    target_arch = "wasm32",
-    target_os = "unknown",
-    feature = "custom",
-    not(feature = "js")
-))]
-
-use wasm_bindgen_test::wasm_bindgen_test as test;
```
briansmith, line 9: I think we should keep this, but do it this way, which is commonly done in other crates:
```
#[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
use wasm_bindgen_test::wasm_bindgen_test as test;
```
This way, all tests marked `#[test]` will be run using either libtest or wasm_bindgen_test.

briansmith (changes requested):
This makes a lot more sense to me. Just a couple comments.

tests/custom.rs
briansmith: Also, we should add `target_os = "espdif"` since that one has been debated recently, and also `all(target_arch = "wasm32", target_os = "wasi")`.

josephlr: Good idea. I also added a test in the CI for enabling both "js" and "custom" on a wasm target.

tests/custom.rs
josephlr: Good idea, added (with `target_family = "wasm"`)

tests/custom.rs
josephlr: Done.

newpavlov (approved):

PR #474, opened 2024-06-12
Test `getrandom_uninit`

josephlr:
Depends on #471, Alternative to #462 

I verified that running the tests with MSAN:
```
RUSTFLAGS="-Zsanitizer=memory" cargo +nightly test -Zbuild-std --target=x86_64-unknown-linux-gnu
```
succeeds before this PR and fails with this PR, so we are doing the right thing here.


josephlr:

PR #475, opened 2024-06-12
Test fallback file implementation on Linux/Android

### LINKED ISSUE #289: For platforms that implement runtime fallback logic, all paths should be tested
Consider the case of Linux. We hope to use the `getrandom` syscall and then if that fails, we fall back to doing file I/O. We should test at least the following scenerios:

* Linux returns `ENOSYS`, which should trigger the fallback to file I/O.
* Linux returns `EPERM`, which is presently expected to do the same fallback, though there's an issue on file to change this.
* The syscall succeeds.

This type of exhaustive testing might not be practical on all targets. However, BoringSSL has shown how to do it on Linux; see https://boringssl.googlesource.com/boringssl/+/refs/heads/master/crypto/fipsmodule/rand/urandom_test.cc.

Backends with runtime fallback logic:
- [x] `linux_android_with_fallback` (fallack to `/dev/urandom`, [tested](https://github.com/rust-random/getrandom/blob/master/.github/workflows/tests.yml#L62-L64) using the `getrandom_test_linux_fallback` configuration flag)
- [ ] `netbsd` (fallback to `KERN_ARND`)
- [x] `wasm_js` (runtime selection between Web and Node.js APIs, [tested](https://github.com/rust-random/getrandom/blob/master/.github/workflows/tests.yml#L229-L272) in CI)

josephlr:
Fixes #289 for Linux. Depends on #471

Now when running `cargo test` on Linux, we see additional tests for the file-based implementation:
```
running 11 tests
test imp::fallback::fill_large ... ok
test imp::fallback::fill_zero ... ok
test tests::fill_large ... ok
test tests::fill_zero ... ok
test imp::fallback::multithreading ... ok
test imp::fallback::fill_small ... ok
test tests::fill_small ... ok
test imp::fallback::fill_huge ... ok
test tests::fill_huge ... ok
```


josephlr:

newpavlov:
Closing in favor of #506.

PR #476, opened 2024-06-12
Test RDRAND on x86 / x86_64 targets

josephlr:
Depends on #471 

Now when running `cargo test` we see the additional `rdrand` tests:
```
running 11 tests
test rdrand::tests::fill_large ... ok
test rdrand::tests::fill_zero ... ok
test tests::fill_large ... ok
test tests::fill_zero ... ok
test rdrand::tests::multithreading ... ok
test rdrand::tests::fill_small ... ok
test tests::fill_small ... ok
test tests::fill_huge ... ok
test rdrand::tests::fill_huge ... ok
```

I think the main question here is if we should unconditionally run these tests like we did before, or if we should only run them if the `"rdrand"` feature is set.


newpavlov:
As I wrote [here](https://github.com/rust-random/getrandom/pull/473#pullrequestreview-2112011800), this probably can be done a bit better.

josephlr:

PR #477, opened 2024-06-12
CI: run Clippy for all backends, rework the rustdoc job

### LINKED ISSUE #446: Run `cargo clippy` in CI for all targets
Right now `cargo clippy` is only run 64-bit Linux targets. It should be run for all targets so that important lints that statically enforce safety checks are checked on all targets. See #445 in particular.

newpavlov:
Consolidates Clippy runs into one job which uses fixed Nightly version. Moves Clippy, rustdoc, and rustfmt jobs into a separate workflow file. 

Additionally, tweaks the rustdoc job to use only rustdoc with `-Dwarnings` flag. Minimal versions job has nothing to do with docs and should be done separately for each target. `cargo deadlinks` was not used to check external links and internal links are checked by rustdoc.

Closes #446

### DIFF
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -16,7 +16,7 @@ jobs:
     env:
       RUSTFLAGS: "-Dwarnings"
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
     - uses: dtolnay/rust-toolchain@master
       with:
         # We need Nightly for -Zbuild-std.
@@ -70,7 +70,7 @@ jobs:
     name: rustfmt
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: rustfmt
@@ -82,7 +82,7 @@ jobs:
     name: rustdoc
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@master
         with:
           # We need Nightly for doc_auto_cfg
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -70,7 +70,7 @@ jobs:
     name: rustfmt
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: rustfmt
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -1,0 +1,94 @@
+name: Workspace
+
+on:
+  push:
+    branches: master
+  pull_request:
+    branches: master
+
+permissions:
+  contents: read
+
+jobs:
+  clippy:
+    name: Clippy
+    runs-on: ubuntu-latest
+    env:
+      RUSTFLAGS: "-Dwarnings"
+    steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/workspace.yml
```diff
@@ -0,0 +1,81 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: nightly-2024-06-11
+        components: clippy,rust-src
+    - name: Linux (linux_android_with_fallback.rs)
+      run: cargo clippy --target x86_64-unknown-linux-gnu
+    - name: Linux (linux_android.rs)
+      run: cargo clippy --target x86_64-unknown-linux-gnu --features linux_disable_fallback
```
josephlr, line 25: We should probably add checks for `std` and `custom` features here as well.

.github/workflows/workspace.yml
```diff
@@ -0,0 +1,81 @@
...
+    - name: macOS (getentropy.rs)
+      run: cargo clippy -Zbuild-std --target x86_64-apple-darwin
```
josephlr, line 27: Here and elsewhere (except for the `js` implementation), we should just do `build-std=core`, as we don't need to build all of the standard library just for clippy.

.github/workflows/workspace.yml
```diff
@@ -0,0 +1,81 @@
...
+  clippy:
+    name: Clippy
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: nightly-2024-06-11
+        components: clippy,rust-src
+    - name: Linux (linux_android_with_fallback.rs)
```
josephlr, line 22: Should we sort the steps in this job by filename? That might make it easier to tell that all the (non util) files are checked.

.github/workflows/workspace.yml
```diff
@@ -0,0 +1,81 @@
...
+    - name: Fuchsia (fuchsia.rs)
+      run: cargo clippy -Zbuild-std --target x86_64-unknown-fuchsia
+    - name: iOS (apple-other.rs)
+      run: cargo clippy -Zbuild-std --target x86_64-apple-ios
+    - name: WASI (wasi.rs)
+      run: cargo clippy -Zbuild-std --target wasm32-wasip2
+    - name: Hermit (hermit.rs)
+      run: cargo clippy -Zbuild-std --target x86_64-unknown-hermit
+    # - name: VxWorks (vxworks.rs)
+    #   run: cargo clippy -Zbuild-std --target x86_64-wrs-vxworks
```
josephlr, line 45: With `-Zbuild-std=core` this works now.

.github/workflows/workspace.yml
```diff
@@ -0,0 +1,81 @@
+name: Workspace
```
josephlr, line 1: For this file, should we be setting `RUSTFLAGS: "-Dwarnings"`? Clippy issues are warnings, and we are pinning the version, so it seems like we should deny warnings.

josephlr (changes requested):
I like this, just a few minor nits.

.github/workflows/workspace.yml
newpavlov: You are right. But I added it only for the Clippy job, since it's not needed for the other two.

.github/workflows/workspace.yml
newpavlov: I mostly followed the order used in lib.rs. I reordered it alphabetically, but it also may be worth to reorder (in a separate PR) lib.rs as well.

josephlr (approved):

Issue #453, opened 2024-06-04
use_file: Just use `libstd` on targets that we know have libstd, instead of pthreads + libc.


briansmith:
I propose that on all targets where we know libstd is available, we use libstd instead of pthreads and libc for the `use_file` implementation, effectively reverting to the original implementation.

The current proposal for x86_64-unknown-linux-none in #424 is to limit support for that target to Linux kernel versions that have the `getrandom` syscall, so it won't use `use_file`.

Historically, we've tried to make it so `getrandom` avoids using libstd in part so that libstd itself could use `getrandom` as a dependency. However, I think it is clear that libstd is not going to use this crate as its needs are almost completely different; indeed its code has diverged considerably.

Another reason we've avoided libstd is to ensure that we are `#[no_std]` compatible. However, once we add x86_64-unknown-linux-none support, we'll have an actual no-std target that we can build (at least) for to ensure the core parts of this crate remain `no_std` compatible.

`use_file` is a significant chunk of the unsafe code, with the most problematic dependencies (libpthreads). It duplicates functionality that's in libstd, and the functionality it duplicates is exactly the kind of functionality (threading primitives) that we should avoid duplicating, as we're not getting any benefit from the QA and improvements in libstd.

Of all the targets that use the `use_file` implementation, only one doesn't have libstd support: 32-bit x86 QNX Neutrino. I believe @samkearney is working on it, based on their comment in https://github.com/rust-lang/rust/pull/109173#issuecomment-1476495419.

samkearney:
Thanks for the mention @briansmith, just FYI, I have abandoned the work to bring std support to x86 QNX neutrino 7.0. QNX dropped support for 32-bit architectures in 7.1 (released 2020) and the cost-benefit for adding support to a 7-year-old release doesn't make sense anymore.

That being said, I also don't care if no_std x86 QNX 7.0 breaks, and I would wager pretty strongly that virtually nobody else is using it either (it is much more likely that they are using the more recent QNX 7.1 or the in-progress support for QNX 8).

briansmith:
I think we should switch 32-bit x86 QNX to be a target for which we support custom implementations, and drop the built-in implementation for it. That way we don't have to carry around the current implementation forever just for it. Anybody who needs the current implementation could adopt the current `use_file` code in their custom implementation.

samkearney:
Works for me. 👍

briansmith:
I agree with your "even if this crate were used by libstd...".

[...]

PR #490, opened 2024-06-21
use_file: switch to futex on Linux and to `nanosleep` on other targets

newpavlov:
All non-Linux targets only open file in the critical section, so we probably can remove all synchronization and replace it with a sleep-based wait loop.

On Linux initialization thread may spend a lot of time in the critical section because of `wait_until_rng_ready`. On entropy-starved (and especially virtualized) systems we may spend tens of seconds in it, so we need a proper synchronization between initializing and waiting threads. Luckily, the futex API is a great fit in this case and allows us to efficiently handle the synchronization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Breaking Changes
-- Update MSRV to 1.38 [#425]
+- Update MSRV to 1.60 [#472]
 - Remove support of the `wasm32-wasi` target (use `wasm32-wasip1` or `wasm32-wasip2` instead) [#499]
+
+### Changed
+- Switch to `futex` on Linux and to `nanosleep`-based wait loop on other targets
+  in the `use_file` backend [#490]
 
 ### Added
 - `wasm32-wasip1` and `wasm32-wasip2` support [#499]
 
-[#425]: https://github.com/rust-random/getrandom/pull/425
+[#472]: https://github.com/rust-random/getrandom/pull/472
+[#490]: https://github.com/rust-random/getrandom/pull/490
 [#499]: https://github.com/rust-random/getrandom/pull/499
 
 ## [0.2.15] - 2024-05-06
--- a/src/linux_android_with_fallback.rs
+++ b/src/linux_android_with_fallback.rs
@@ -8,7 +8,13 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
     if HAS_GETRANDOM.unsync_init(is_getrandom_available) {
         linux_android::getrandom_inner(dest)
     } else {
-        use_file::getrandom_inner(dest)
+        // prevent inlining of the fallback implementation
+        #[inline(never)]
+        fn inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
+            use_file::getrandom_inner(dest)
+        }
+
+        inner(dest)
     }
 }
 
--- a/src/use_file.rs
+++ b/src/use_file.rs
@@ -4,7 +4,6 @@ use crate::{
     Error,
 };
 use core::{
-    cell::UnsafeCell,
     ffi::c_void,
     mem::MaybeUninit,
     sync::atomic::{AtomicI32, Ordering},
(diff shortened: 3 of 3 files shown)


newpavlov:
cc @briansmith

josephlr:
I don't think we should split how we do synchronization between the different targets. It adds a bunch of complexity for little gain. No contention occurs in the common case.

Also, see https://github.com/rust-random/getrandom/issues/453#issuecomment-2182476778. If we want something better than `pthread_mutex_{lock/unlock}`, we should either use `std::sync::Mutex` or just not use a lock at all (and use libc::poll to handle sleeping/waking the threads).

newpavlov:
All other `use_file` users are relatively minor targets, so I think it makes sense to have a more specialized version of this code for Linux. Especially considering the differences between Linux and the rest (futex availability, `/dev/random` polling, `#[inline(never)]` gating). Moving out the Linux stuff makes `use_file` much cleaner, without any target-specific `cfg`s.

Not only `std` mutex adds unnecessary global state, but its implementation has a bunch of machinery which is not needed in our case (e.g. poisoning). Splitting the Linux code also could help with #401, if we'll decide to implement it.

IIUC implementation for the remaining targets uses pthread mutex under the hood, so with this PR we can continue to use pthread mutex, since it was primarily problematic on Linux and not other targets. Considering that non-Linux targets only open file in the critical section, I think we can even remove locking completely and replace it with a simple sleep-based loop similar to what I drafted in #478, without any `libc::poll` complications.

newpavlov:
Hm, if we are fine with a sleep-based loop for non-Linux targets, I guess we could eliminate the split by using sleep-based implementation of the `wait` function.

josephlr:
> Not only `std` mutex adds unnecessary global state, but its implementation has a bunch of machinery which is not needed in our case (e.g. poisoning).

The main benefit of `std`'s Mutex is that it's well tested an reviewed and not a custom thing we need to maintain. I think using the already present implementation is fine. It doesn't really matter that `Mutex` supports more functionality than we need, it's already implemented as part of Rust.

newpavlov:
The futex wait/wake operations are very straightforward, they are essentially textbook-level examples, so I don't think they require any particular maintenance.

I removed the code split and removed dependency on pthread mutex. What do you think about the new code?

josephlr:
> Hm, if we are fine with a sleep-based loop for non-Linux targets, I guess we could eliminate the split by using sleep-based implementation of the `wait` function.

I think its reasonable to get rid of locking for initializing the file descriptor. It's not clear to me if a `sleep` based approach is better or worse than just opening the file (potentially multiple times) and then closing the file if the thread "loses" the race.

newpavlov:
If we are to close eyes on the `sync` module for now, are you fine with the rest of the code?

>If you don't want to use stds Mutex, what would be your thoughts on the second proposal in https://github.com/rust-random/getrandom/issues/453#issuecomment-2182476778 to just not have any locking at all?

It's essentially what I did in this PR. On non-Linux targets there is no locking, but on Linux we need some kind of locking because `wait_until_rng_ready` may block for a significant amount of time. 

`libc::poll`-based synchronization IMO is more convoluted and hacky than the futex-based code and `LazyFd` will result in unnecessary opening operations, which may hit `ulimit`.

josephlr:
> `libc::poll`-based synchronization IMO is more convoluted and hacky than the futex-based code. `LazyFd` will result in unnecessary opening operations, which may hit `ulimit`. In the worst case scenario, every thread will open `/dev/random` to poll it, and after it's ready, they all will rush to open `/dev/urandom`.


I mean we would be using `libc::poll` regardless, that part would be unchanged. As for potentially opening multiple files, I don't think it would happen in practice. Even if a bunch of threads are `poll`-ing `/dev/random` they will all be polling the _same_ fd (in my proposal). The only way multiple fds get opened is if two threads race on `libc::open` itself, and even in that scenario, the extra FD gets closed immediately.

josephlr:
> it's one of the reasons why I thought that the code split can be a good idea

Looking at this, I actually think that splitting the code is fine, as we can share common abstractions (if any) between `use_file.rs` and `linux_with_fallback.rs`. My main concern is with _how_ the locking/sleeping/waiting is implemented.

josephlr:
> If we are to close eyes on the `sync` module for now, are you fine with the rest of the code?

I would say the contents/complexity of the `sync` module are the main reason I don't prefer this approach. I agree that if we just were opening shared FDs, we could solve this in multiple nice ways without using `std`.

newpavlov:
>I mean we would be using libc::poll regardless, that part would be unchanged. 

Ah, I somewhat misunderstood your idea. I guess it could work, but I think it will be more code and (to my taste) less straightforward. Plus, as you note yourself, this approach will always result in 2 file descriptors being continuously opened instead of 1.

> we can share common abstractions (if any) between use_file.rs and linux_with_fallback.rs

Amount of common abstractions depends on whether we use pthread mutex or the sleep-based loop for non-Linux targets. In the former case the amount is small enough to warrant complete independence. But in the former case I think it makes more sense to keep code in one module like done in the current version.

>I would say the contents/complexity of the sync module are the main reason I don't prefer this approach.

I guess perceived complexity depends on how familiar reader is with futexes. Maybe I can amend this by adding comments explaining what exactly `wake` and `wait` do?

josephlr:
> I guess perceived complexity depends on how familiar reader is with futexes. Maybe I can amend this by adding comments explaining what exactly `wake` and `wait` do?

Looking at the man page for futex I found this:
> To reiterate, bare futexes are not intended as an easy-to-use abstraction for end-users. (There is no wrapper function for this system call in glibc.)

This makes me think that futexes (whike very cool) are not really designed for general use, but rather as a way to allow things like `std` to have fast mutexes.

newpavlov:
>This makes me think that futexes (whike very cool) are not really designed for general use, but rather as a way to allow things like std to have fast mutexes.

Futex is certainly a very low-level API which is usually used as a building block for implementing higher-level synchronization structures. But it's not like it does not get used to implement custom synchronization primitives in production code (e.g. some people use it to implement an io-uring-like IPC API on top of shared memory). IMO it's an extremely good fit solution for our problem.

newpavlov:
@josephlr 
I would like to move forward with this PR. Could you take a look at it again?

newpavlov:
I will merge this for now to make it easier to work on other stuff. We can revert or change it before v0.3 release, since I will  be waiting for other people's review before cutting it.

briansmith:
> I will merge this for now to make it easier to work on other stuff. We can revert or change it before v0.3 release, since I will be waiting for other people's review before cutting it.

Why was it urgent to merge this? This is quite a big change to merge without review and I think the lack of review is mostly because nobody agrees we should do this. I think we should figure out a way to make progress on the other things that this was blocking without having this change merged.

newpavlov:
Neither you, nor Joe were active on GitHub for quite some time. I wanted to work on the PR backlog and had some time for that. I don't want to deal with potential merge conflicts in multiple PRs while a work on other features. As we discussed with Joe, the merges are not final and will be reviewed and maybe changed before v0.3 release.

PR #496, opened 2024-08-06
Update Cargo.toml to declare libc dependency for target_os=emscripten

burakemir:
When targetting wasm32-unknown-emscripten, there is a dependency on libc.
The Cargo.toml should reflect that.


newpavlov:
IIUC Emscripten in included into the UNIX target OS family, so this is redundant. Note that we [test](https://github.com/rust-random/getrandom/blob/master/.github/workflows/tests.yml#L162) Emscripten build in our CI.

burakemir:
Thanks, you are right and there is an issue in our build configuration. Also figured out how to `--print all-target-specs-json`.

Closing this.

Issue #498, opened 2024-09-05
Allow to change default entropy source with configuration flags


newpavlov:
Some platforms provide hardware-based "true" RNG sources which can be faster and have smaller footprint than using OS-provided APIs. For example: the RDRAND instruction (x86), the RNDR register (ARM, see #494), the `seed` CSR (RISC-V). In most cases users should prefer OS-provided APIs because they are more widely available, likely less buggy (see comments in the `rdrand` module), and there are less security concerns about them. But in some controlled cases it could be useful to be able to switch to these entropy sources.

Crate features could be used for that, but they are not a good fit since, strictly speaking, change of entropy backend is not "additive". Some users also misuse those features and enable them unconditionally in library crates. It's probably better to use configuration flags (i.e. `cfg(getrandom = "rdrand")`). In future we may replace them with something like [mutually exclusive features](https://github.com/rust-lang/cargo/issues/2980).

This configuration flag should probably replace the existing `linux_disable_fallback`, `rdrand`, and `js` features. The last one is the most debatable. It's the most used feature compared to the other two, so replacing it with a configuration flag may result in complaints from developers targeting Web WASM.

 We also can gate support of "problematic" targets such as ESP-IDF (see #397) behind such flag to improve visibility of potential issues with such targets.

PR #501, opened 2024-09-23
fix: Override other implementations on custom feature

### LINKED ISSUE #500: Add support for targeting the macOS (Darwin/XNU) kernel
Currently if the target OS is macOS, it assumes that you are building an application and not a kernel extension. This causes getrandom to try using libc, which is obviously not available in the kernel.

VisualEhrmanntraut:
Fixes #500

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -235,7 +235,9 @@ pub use crate::error::Error;
 // The function MUST NOT ever write uninitialized bytes into `dest`,
 // regardless of what value it returns.
 cfg_if! {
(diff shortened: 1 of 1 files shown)


newpavlov:
We intentionally do not allow to overwrite "default" entropy sources with custom ones. This may change with #498, but your current approach is certainly is a no-go.

newpavlov:
Renaming the feature does not solve the root security issue with overriding default entropy sources. As I wrote in the issue, at the moment, your best option is to simply patch the `getrandom` crate with your custom version.

PR #502, opened 2024-09-27
wasi: remove dependency on wasi v0.11

newpavlov:
Having two `wasi` versions in dependencies causes some annoyances in downstream users, e.g. v0.11 gets show as an "outdated" dependency. The dependency is removed by directly linking `random_get` by following the `wasi` code.


newpavlov:
@yoshuawuyts
Could you take a look?

newpavlov:
Huh, it seems we can not "just" link to `get_random_u64`. It looks like it needs a pretty [arcane linking magic](https://docs.rs/wasi/0.13.2+wasi-0.2.1/src/wasi/bindings.rs.html#15466-15784)...

PR #504, opened 2024-10-04
Use configuration flags to switch between backends

### LINKED ISSUE #230: It should be possible to use a custom implementation when RDRAND is available

### LINKED ISSUE #346: Should the "custom" feature take precedence over the "js" feature?
Currently the "custom" feature is last in our precedence order for selecting implementations. The original idea behind this was to ensure that a crate couldn't (accidentally or intentionally) change the RNG source for a program.

Sometimes folks are targeting `wasm32-unknown-unknown`, bu they are not targeting the web, but their deps are enabling the `js` feature. #342 CC: @nickray

I would propose the following new precedence order:
  - If on a supported target, use the standard implementation
  - If `custom` feature is enabled, use the custom implementation
  - If on `wasm32-unknown-unknown` and `js` is enabled, use the `wasm-bindgen` implementation.

### LINKED ISSUE #397: Is the ESPIDF implementation actually correct?
When reviewing the corresponding change to *ring*, https://github.com/briansmith/ring/pull/1944, I noticed that the implementation in `getrandom` for ESPIDF seems really questionable. The ESPIDF documentation is quite vague on what promises its PRNG is making, especially when the OS isn't fully configured. It also encourages people to use a userspace CSPRNG; it isn't clear if this suggestion is intended to just address performance or security or both. 

In particular, should the ESPIDF random API considered more of an entropy source, but not a full CSPRNG?

In *ring* we've taken the approach, temporarily, of using `getrandom` on this OS only if the user opts in with a feature flag that draws attention to this concern.

### LINKED ISSUE #498: Allow to change default entropy source with configuration flags

newpavlov:
This PR removes `linux_disable_fallback`, `rdrand`, `js`, `test-in-browser`, and `custom` crate features. As their replacement two new configuration flags are introduced: `getrandom_browser_test` and `getrandom_backend`. The latter can have the following values: `custom`, `rdrand`, `linux_getrandom`, `wasm_js`, `esp_idf`.

`getrandom_backend` allows to change default backends which resolves issues like #346 and provides more flexibility to users. For example, it allows to use RDRAND or RDRND (see #494) directly instead of syscall-based interfaces. We previously did not allow such overwrites because of security concerns, but they do not apply in this case since configuration flags used by a project can not be controlled by its upstream dependencies.

The `register_custom_getrandom!` macro is removed in favor of explicitly defining the `__getrandom_custom` function. It does not look like the macro was widely used in practice and it's probably easier to explain the `extern fn` approach (especially to embedded developers) than the "magical" registration macro.

The new configuration flags also allow a great simplification of our testing code.

Finally, ESP-IDF support is no longer enabled by default because of the concerns raised in #397. Users can enable it by enabling the `esp_idf` opt-in backend.

Closes #230
Closes #346
Closes #397
Closes #498


src/lib.rs
```diff
@@ -237,7 +250,34 @@ pub use crate::error::Error;
...
+    } else if #[cfg(getrandom_backend = "linux_getrandom")] {
+        #[cfg(not(any(target_os = "android", target_os = "linux")))]
+        compile_error!("`linux_getrandom` backend can be enabled only for Linux/Android targets!");
+        mod util_libc;
+        #[path = "linux_android.rs"] mod imp;
+    } else if #[cfg(getrandom_backend = "rdrand")] {
+        #[cfg(not(any(target_arch = "x86_64", target_arch = "x86")))]
+        compile_error!("`rdrand` backend can be enabled only for x86 and x86-64 targets!");
+        #[path = "rdrand.rs"] mod imp;
+    } else if #[cfg(getrandom_backend = "wasm_js")] {
```
newpavlov, line 264: We could split `wasm_js` feature further into `wasm_node` and `wasm_web` parts. But I don't know whether people need WASM files which can run both on Node and Web.

src/custom.rs
```diff
@@ -1,103 +1,12 @@
...
         fn __getrandom_custom(dest: *mut u8, len: usize) -> u32;
     }
-    // Previously we always passed a valid, initialized slice to
-    // `__getrandom_custom`. Ensure `dest` has been initialized for backward
-    // compatibility with implementations that rely on that (e.g. Rust
-    // implementations that construct a `&mut [u8]` slice from `dest` and
-    // `len`).
-    let dest = uninit_slice_fill_zero(dest);
-    let ret = unsafe { __getrandom_custom(dest.as_mut_ptr(), dest.len()) };
+    let ret = unsafe { __getrandom_custom(dest.as_mut_ptr().cast(), dest.len()) };
```
newpavlov, line 9: Note that in v0.3 we no longer need to zeroize the destination buffer since we can specify that uninitialized buffers may be passed to `__getrandom_custom`.

We also probably could use `NonZeroUsize` for `len` to improve code generation for `__getrandom_custom` provider.

src/lib.rs
```diff
@@ -237,7 +250,34 @@ pub use crate::error::Error;
...
+        #[cfg(not(all(
+            any(target_arch = "wasm32", target_arch = "wasm64"),
+            target_os = "unknown",
+        )))]
+        compile_error!("`wasm_js` backend can be enabled only on OS-less WASM targets!");
+        #[path = "js.rs"] mod imp;
+    } else if #[cfg(getrandom_backend = "esp_idf")] {
+        #[cfg(not(target_os = "espidf"))]
+        compile_error!("`esp_idf` backend can be enabled only for ESP-IDF targets!");
```
newpavlov, line 273: ESP-IDF support is now optional because of the concerns raised in https://github.com/rust-random/getrandom/issues/397.

newpavlov:
cc @briansmith

Issue #503, opened 2024-09-28
Use vDSO on Linux 6.11


ericlagergren:
See https://github.com/golang/go/issues/69577

briansmith:
I think this would be good to have opt-in for a while, while the DSO feature is being tested in the real world.

newpavlov:
This may be interesting in the context of #424, but otherwise I think we should just use `libc::getrandom` (maybe through `dlopen` in the backend with `/dev/urandom` fallback) and let it decide how the syscall is handled.

newpavlov:
Closing in favor of #508 and https://github.com/bytecodealliance/rustix/issues/1185.

PR #508, opened 2024-10-07
linux_android: use `libc::getrandom`

newpavlov:
Use of `libc::getrandom` will automatically give us optimizations like vDSO (#503) and can help with testing of fallback logic (#289).

It was also requested in #285. In that discussion we decided against using this approach, but in the light of the vDSO optimization it may be worth to reconsider it.

In `linux_android_with_fallback` use of `libc::syscall` is replaced by `dlsym`-based code similar to what we use in the `netbsd` backend.

### DIFF
--- a/src/linux_android_with_fallback.rs
+++ b/src/linux_android_with_fallback.rs
@@ -1,37 +1,80 @@
 //! Implementation for Linux / Android with `/dev/urandom` fallback
-use crate::{lazy::LazyBool, linux_android, use_file, util_libc::last_os_error, Error};
-use core::mem::MaybeUninit;
+use crate::{use_file, util_libc, Error};
+use core::{
+    ffi::c_void,
+    mem::{self, MaybeUninit},
+    ptr::{self, NonNull},
+    sync::atomic::{AtomicPtr, Ordering},
+};
 
-pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
-    // getrandom(2) was introduced in Linux 3.17
-    static HAS_GETRANDOM: LazyBool = LazyBool::new();
-    if HAS_GETRANDOM.unsync_init(is_getrandom_available) {
-        linux_android::getrandom_inner(dest)
-    } else {
-        // prevent inlining of the fallback implementation
-        #[inline(never)]
-        fn inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
-            use_file::getrandom_inner(dest)
+type GetRandomFn = unsafe extern "C" fn(*mut c_void, libc::size_t, libc::c_uint) -> libc::ssize_t;
+
+/// Sentinel value which indicates that `libc::getrandom` either not available,
+/// or not supported by kernel.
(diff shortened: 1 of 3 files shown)


src/linux_android_with_fallback.rs
```diff
@@ -1,37 +1,80 @@
...
+fn init() -> NonNull<c_void> {
+    static NAME: &[u8] = b"getrandom\0";
+    let name_ptr = NAME.as_ptr().cast::<libc::c_char>();
+    let raw_ptr = unsafe { libc::dlsym(libc::RTLD_DEFAULT, name_ptr) };
+    let res_ptr = match NonNull::new(raw_ptr) {
+        Some(fptr) => {
+            let getrandom_fn = unsafe { mem::transmute::<NonNull<c_void>, GetRandomFn>(fptr) };
+            let dangling_ptr = ptr::NonNull::dangling().as_ptr();
+            // Check that `getrandom` syscall is supported by kernel
+            let res = unsafe { getrandom_fn(dangling_ptr, 0, 0) };
```
newpavlov, line 28: Maybe we can skip these check and assume that if `libc` exposes `getrandom`, it will be supported by the kernel?

src/linux_android_with_fallback.rs
```diff
@@ -1,37 +1,80 @@
...
-    if HAS_GETRANDOM.unsync_init(is_getrandom_available) {
-        linux_android::getrandom_inner(dest)
-    } else {
-        // prevent inlining of the fallback implementation
-        #[inline(never)]
-        fn inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
-            use_file::getrandom_inner(dest)
+type GetRandomFn = unsafe extern "C" fn(*mut c_void, libc::size_t, libc::c_uint) -> libc::ssize_t;
+
+/// Sentinel value which indicates that `libc::getrandom` either not available,
```
ericlagergren, line 12: This is complicated. Is there really such a big performance penalty for unconditionally calling getrandom each time and then falling back to /dev/urandom when getrandom returns ENOSYS?

ericlagergren: ...or just using an AtomicBool like before?

newpavlov: What do you mean by "complicated"? This value is guaranteed to not be a valid function pointer, so we use it as "uninitialized" marker.

`dlsym` can be a relatively heavy operation, so I would prefer to cache its result.

src/linux_android_with_fallback.rs
ericlagergren: No, getrandom is documented as returning ENOSYS.

tamird:
This new implementation seems to fall back where the old one didn't. I believe that's because I'm statically linking musl.

The result is a crash, because I'm doing this in a very special "distro" which doesn't have /dev/urandom mounted. See https://github.com/aya-rs/aya/blob/665d4f20bb53de0aa10545bb897ab73f0661a337/init/src/main.rs.

Anyway, I'm pretty sure use dlsym in this way isn't going to work for anyone statically linking musl (which is a major reason for using musl).

tamird:
Yeah, exactly - there is no test to check that musl works correctly (i.e. no fallback). I'll leave you to add that test, but I'm sending a PR that I think fixes this by not using dlsym on musl.

PR #510, opened 2024-10-09
Replace `as` casts with safer conversions

newpavlov:
Based on #445.


src/linux_android.rs
```diff
@@ -6,14 +6,18 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
...
             0,
-        ) as libc::ssize_t
-    }
+        )
+    };
+
+    const _: () =
+        assert!(core::mem::size_of::<libc::c_long>() == core::mem::size_of::<libc::ssize_t>());
+    res.try_into()
+        .expect("c_long to ssize_t conversion is lossless")
```
newpavlov, line 22: This cast no longer will be needed if #508 lands.

src/wasi.rs
```diff
@@ -24,10 +24,16 @@ pub fn getrandom_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
     // https://docs.rs/wasi/0.11.0+wasi-snapshot-preview1/src/wasi/lib_generated.rs.html#2046-2062
     // Note that size of an allocated object can not be bigger than isize::MAX bytes.
     // WASI 0.1 supports only 32-bit WASM, so casting length to `i32` is safe.
+    #[allow(clippy::cast_possible_truncation, clippy::cast_possible_wrap)]
```
newpavlov, line 27: This is the only place which needs to disable the lints. Unfortunately, the compiler does not use the requirement that non-ZST allocations can not be bigger than `isize::MAX`, so using `TryInto` for converting length into `i32` will contain a panic. We also could return `Error::UNEXPECTED` if `dest.len() > i32::MAX`, but I don't like such workaround.

PR #515, opened 2024-10-14
ci: enforce that `getrandom` does not contain potential panics

### LINKED ISSUE #435: Document and Test that `getrandom()` never panics.
Spun off of the review of #434 

We should have a way to verify in the CI that our implementation (at least on easy-to-test architectures) does not panic. This could be done via the [`no-panic` crate](https://crates.io/crates/no-panic) as a dev-dependancy.

newpavlov:
The check is done by compiling `cdylib` which exposes `getrandom_uninit` and checking that the result does not contain any strings with "panic".

For now CI only checks `linux_android_with_fallback`, `linux_android`, `rdrand`, `wasi`, and `getentropy` backends. Other backend checks could be added in future PRs.

Closes #435

### DIFF
--- a/.github/workflows/nopanic.yaml
+++ b/.github/workflows/nopanic.yaml
@@ -1,0 +1,58 @@
+name: No panic
+
+on:
+  push:
+    branches: master
+  pull_request:
+    branches: master
+
+permissions:
+  contents: read
+
+defaults:
+  run:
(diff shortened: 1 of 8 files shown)


newpavlov:
The commits above demonstrate that CI successfully detects potential panics.

Issue #516, opened 2024-10-14
No-panic check tracking issue


newpavlov:
Backends list:
- [ ] apple-other
- [ ] custom
- [ ] esp_idf
- [ ] fuchsia
- [x] getentropy
- [ ] getrandom
- [ ] hermit
- [x] linux_android
- [x] linux_android_with_fallback
- [ ] netbsd
- [x] rdrand
- [ ] rndr
- [ ] solaris
- [ ] solid
- [x] use_file (checked via linux_android_with_fallback)
- [ ] vxworks
- [x] wasi
- [x] wasm_js (can not be checked)
- [ ] windows7
- [ ] windows

newpavlov:
Closing this issue since the most important backends are covered and we do not have a clear path forward for the remaining ones.

PR #524, opened 2024-10-16
ci: add nopanic check for apple-other

newpavlov:
Relevant issue: #516


newpavlov:
The two failing commits above demonstrate successful detection of panics in `apple-other` and `getentropy` backends respectively.

PR #526, opened 2024-10-16
ci: add nopanic check for windows

newpavlov:
Relevant issue: #516

### DIFF
--- a/.github/workflows/nopanic.yaml
+++ b/.github/workflows/nopanic.yaml
@@ -136,6 +136,6 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: nightly-2024-10-14
-          # components: rust-src
-      # - run: cargo build --release --target=x86_64-win7-windows-msvc -Zbuild-std=std,panic_abort
+          components: rust-src
       - run: cargo build --release
+      - run: cargo build --release --target=x86_64-win7-windows-msvc -Zbuild-std="std,panic_abort"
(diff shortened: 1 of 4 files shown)


newpavlov:
It looks like `link.exe` does not perform the same dead code elimination as linkers in Linux and macOS, so we need to reconsider our approach towards nopanic checks on Windows.

ChrisDenton:
On my machine `link.exe` does eliminate it but still wants it to exist in the first place. In any case, you could use `lto` to eliminate the symbol before the linker even sees it.

newpavlov:
LTO is irrelevant here. I tried to use the following program:
```rust
#![no_std]

#[panic_handler]
fn panic(_info: &core::panic::PanicInfo) -> ! {
    extern "C" {
        fn panic_nonexistent() -> !;
    }
    unsafe { panic_nonexistent() }
}

#[no_mangle]
pub extern "C" fn getrandom_wrapper(_: *mut u8, _: usize) -> u32 {
    0
}
```
And it still was failing the linking step. I will try to use a slightly different approach to this a bit later.

ChrisDenton:
It should be highly relevant. Link time optimization happens before the linker is given anything (unless you use llvm's linker) so if `panic` is never called then no references to `panic_nonexistent` will exist by the time the linker sees it.

newpavlov:
Huh, you are right. Thank you! TIL, before that I thought LTO applies only during "link time" and has not effect in cases like the snippet above.

PR #541, opened 2024-11-21
Add `wasm32v1-none` support

daxpedda:
Changes:
- Add `default-features = false` to the `js-sys` and `wasm-bindgen-test` dependencies, enabling `no_std` support.
- The crate feature `std` now enabled `std` crate features of `wasm-bindgen`, `js-sys` and `wasm-bindgen-test`.
- Use `any(target_os = "unknown", target_os = "none")` instead of `target_os = "unknown"`.
- `thread_local!` is only supported with `std`, so a fallback to `static` with `once_cell::unsync::Lazy` is added. This is safe because Wasm doesn't support threads without `target_feature = "atomics"`.
- When detecting `target_feature = "atomics"` without `feature = "std"`, use [`#[thread_local]`](https://doc.rust-lang.org/1.82.0/unstable-book/language-features/thread-local.html?highlight=thread_lo#thread_local).
- Now `wasm32-unknown-unknown` is properly supported without pulling in `std`.

I was unable to add actual `no_std` tests without some ugly modifications to it. Especially because of the whole `harness = false` thing. Let me know how you want me to proceed.

~~This depends on https://github.com/rustwasm/wasm-bindgen/pull/4277 and a new `wasm-bindgen` release.~~
Was released.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -245,33 +245,33 @@ jobs:
       - name: Test (Node)
         env:
           RUSTFLAGS: -Dwarnings --cfg getrandom_backend="wasm_js"
-        run: wasm-pack test --node
+        run: wasm-pack test --node -- --features std
       - name: Test (Firefox)
         env:
           WASM_BINDGEN_USE_BROWSER: 1
           RUSTFLAGS: -Dwarnings --cfg getrandom_backend="wasm_js"
-        run: wasm-pack test --headless --firefox
+        run: wasm-pack test --headless --firefox -- --features std
       - name: Test (Chrome)
         env:
           WASM_BINDGEN_USE_BROWSER: 1
           RUSTFLAGS: -Dwarnings --cfg getrandom_backend="wasm_js"
-        run: wasm-pack test --headless --chrome
+        run: wasm-pack test --headless --chrome -- --features std
       - name: Test (dedicated worker)
         env:
           WASM_BINDGEN_USE_DEDICATED_WORKER: 1
           RUSTFLAGS: -Dwarnings --cfg getrandom_backend="wasm_js"
-        run: wasm-pack test --headless --firefox
+        run: wasm-pack test --headless --firefox -- --features std
       - name: Test (shared worker)
         env:
           WASM_BINDGEN_USE_SHARED_WORKER: 1
           RUSTFLAGS: -Dwarnings --cfg getrandom_backend="wasm_js"
-        run: wasm-pack test --headless --firefox
+        run: wasm-pack test --headless --firefox -- --features std
       - name: Test (service worker)
         env:
           WASM_BINDGEN_USE_SERVICE_WORKER: 1
           RUSTFLAGS: -Dwarnings --cfg getrandom_backend="wasm_js"
         # Firefox doesn't support module service workers and therefor can't import scripts
-        run: wasm-pack test --headless --chrome
+        run: wasm-pack test --headless --chrome -- --features std
 
   wasi:
     name: WASI
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,8 @@
 # Allow normal use of "cargo run" and "cargo test" on these wasm32 platforms.
 [target.wasm32-unknown-unknown]
 runner = 'wasm-bindgen-test-runner'
+[target.wasm32v1-unknown]
+runner = 'wasm-bindgen-test-runner'
 [target.wasm32-wasip1]
 runner = 'wasmtime'
 [target.wasm32-wasip2]
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -23,7 +23,7 @@ jobs:
         # Fixed Nigthly version is used to prevent
         # CI failures which are not relevant to PR changes
         # on introduction of new Clippy lints.
-        toolchain: nightly-2024-10-08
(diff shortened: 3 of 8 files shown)


daxpedda:
Current CI failure is unrelated.

daxpedda:
Hm, CI still has an unrelated failure.

newpavlov:
Yes, see https://github.com/rust-lang/rust/issues/133401. We may disable the GNU/Hurd job since it's not the first time the `std` breaks this target.

newpavlov:
How costly the `getrandom_init` call in practice? Maybe we should just skip the thread local dance and get fresh `RngSource` on each call? As a small optimization we could cache the source kind in an atomic static.

Also, is it common to use one (non-trivial) WASM file on both Web and Node environments simultaneously, i.e. maybe we should split `wasm_js` into `wasm_web_js` and `wasm_node_js`?

daxpedda:
> How expensive is the `getrandom_init` call in practice? Maybe we should just skip the thread-local dance and obtain fresh `RngSource` on each call? As a small optimization we could cache the source kind in an atomic static.

For Web that seems reasonable. NodeJS seems to be quite involved and I would argue that a thread-local is justified.

> Also, is it common to use one (non-trivial) WASM file in both Web and Node environments simultaneously, i.e. maybe we should split `wasm_js` into `wasm_web_js` and `wasm_node_js`?

AFAIK it is common to do it together. Whats strange to me is that the implementation is different.

newpavlov:
I've opened #557 to see how such simplification could look like.

.github/workflows/tests.yml
```diff
@@ -245,33 +245,33 @@ jobs:
       - name: Test (Node)
         env:
           RUSTFLAGS: -Dwarnings --cfg getrandom_backend="wasm_js"
-        run: wasm-pack test --node
+        run: wasm-pack test --node -- --features std
```
newpavlov, line 248: Can we add `wasm-pack` test for the `wasm32v1-none` target?

daxpedda: `wasm32v1-none` doesn't support `std`. I noted in the OP that I would need some guidance on how to add `no_std` tests to the project.

> I was unable to add actual `no_std` tests without some ugly modifications to it. Especially because of the whole `harness = false` thing. Let me know how you want me to proceed.

newpavlov: Hm. I guess the best option will be to make code free of `std` items and since the code will be equivalent for both targets we can just rely on `wasm32-unknown-unknown` testing.

newpavlov:
>I didn't dig into the specifics on why the NodeJS implementation is so special, the API should be the same.

It's probably because a full Web Crypto support was [added](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) only in Node.js 19 (2022-10-18) and `getRandomValues()` was techincally supported since 17.4 (2022-01-18). While we had releases with WASM support a fair time before that. So I guess we can just drop the alternative Node.js path completely.

>I will add the implementation to that effect to this PR tonight.

We also can continue with #557.

>Get rid of the Uint8Array when not using shared memory.

Can you demonstrate how this detection could look like?

.github/workflows/tests.yml
daxpedda: The problem is specifically [disabling the default test harness](https://doc.rust-lang.org/1.83.0/cargo/reference/cargo-targets.html#the-harness-field). While `wasm-bindgen-test` is a test harness, it only works for Wasm. So if we disable the default test harness, native tests won't work.

I propose adding an explicit test target just for Web that is disabled by default:
```toml
[[test]]
name = "no_std-test"
path = "tests/mods.rs"
harness = false
test = false
```
Additionally we would have to add:
- `no_main`
- `no_std`
- A global allocator.

Which I propose could be added in a separate `web` module to not pollute the native tests.

WDYT?

daxpedda:
> > I will add the implementation to that effect to this PR tonight.
> 
> We also can continue with #557. I will leave the choice to you.

Sure, I'm a fan of splitting things up.

> > Get rid of the Uint8Array when not using shared memory.
> 
> Can you demonstrate how this detection could look like?

We can simply use [`wasm_bindgen::memory()`](https://docs.rs/wasm-bindgen/0.2.97/wasm_bindgen/fn.memory.html) to determine if the [`WebAssembly.Memory.buffer`](https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/Memory/buffer) is an `ArrayBuffer` or a `SharedArrayBuffer`. Afterwards we can easily cache it in a static atomic.

.github/workflows/tests.yml
daxpedda: Wouldn't we have to copy the same tests to run in Wasm as well? Or are you proposing to create dedicated Wasm-only tests?

I guess you are saying we can do this in a follow-up or is that something you want done in this PR?

newpavlov: Yes, we can leave addition of `wasm32v1-none` tests for a separate PR, but my main point is that we may not need them in the first place if there are no differences in code between `wasm32-unknown-unknown` and `wasm32v1-none`.

daxpedda: Ah, I see!

daxpedda: EDIT: I agree, if we remove the thread-local related code, we don't need to test that.

daxpedda:
Closing in favor of #557.

PR #545, opened 2024-11-25
Impl core::error::Error for Error on no_std

dhardy:
I assume this is an oversight and not deliberate.


dhardy:
Ah, I believe this wasn't stable until Rust 1.81.0: https://github.com/rust-lang/rust/issues/103765

newpavlov:
Yes. Unfortunately, it's a matter of MSRV.

PR #546, opened 2024-11-27
Add "insecure" functions

### LINKED ISSUE #365: Support for non-blocking getrandom
In some scenarios (e.g. for seeding `HashMap`) it's desirable to get potentially less secure numbers than to block indefinitely. The motivation is largely equivalent to motivation behind the `GRND_NONBLOCK` flag used in the [`getrandom`](https://man7.org/linux/man-pages/man2/getrandom.2.html) syscall. We probably should add support for such scenarios, either by adding an additional argument or by introducing a separate set of non-blocking functions.

Relevant PRs: #353 #352

newpavlov:
Adds 4 new functions: `insecure_fill`, `insecure_fill_uninit`, `insecure_u32`, and `insecure_u32`.

On some platforms these function can be less prone to blocking, but may return potentially "insecure" random data. Such data can be used for seeding `HashMap`s and non-security-sensitive PRNGs.

Closes #365

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ library like [`rand`].
 
 [`rand`]: https://crates.io/crates/rand
 
-## Usage
+## Examples
 
 Add the `getrandom` dependency to your `Cargo.toml` file:
 
@@ -35,6 +35,16 @@ fn get_random_u128() -> Result<u128, getrandom::Error> {
     let mut buf = [0u8; 16];
     getrandom::fill(&mut buf)?;
     Ok(u128::from_ne_bytes(buf))
+}
+```
+
+The crate also support direct generation of random `u32` and `u64` values:
+
+```rust
+fn get_rand_u32_u64() -> Result<(u32, u64), getrandom::Error> {
+    let a = getrandom::u32()?;
+    let b = getrandom::u64()?;
+    Ok((a, b))
 }
 ```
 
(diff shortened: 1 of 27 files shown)


dhardy:
> In its current form, yes, it's. But I haven't decided yet on which approach to use.

Then it could be added in v0.4. I avoided putting out many breaking releases for `rand` because moving to a new version can be a big deal for users, but I don't think that should be the case with `getrandom`.

> Seeding of hash maps and non-security-critical PRNGs are both legitimate use-cases. For example, see discussions about issues with systemd which led to the introduction of the GRND_INSECURE flag.

Yes, systemd is a special case because it operates in an early boot environment. But precisely because this is a special case (and also very targeted to a specific environment) it probably isn't going to use a general-purpose library like `getrandom`. I would expect most other applications needing to deal with early boot to be special cases which wouldn't want to use a general-purpose `getrandom` library too (I may be wrong, but until I see evidence I will assume that is not the case).

Hence "systemd needs this" is not good justification for this feature.

dhardy:
Possibly of note is https://github.com/rust-lang/rust/issues/130703: introduction of what is effectively `getrandom` within `std`. The internal code does use insecure seeding for hashmaps but the public API proposed here only exposes secure seeding.

newpavlov:
This PR got stale, so closing it for now. I probably will open a new one after v0.3 release.

PR #551, opened 2024-11-29
Remove faulty support for `wasm64-unknown-unknown`

daxpedda:
This removes the previously in #303 added support for `wasm64-unknown-unknown`.

`wasm-bindgen` does in fact not support the Memory64 proposal. The reason it still compiled is that `wasm-bindgen` emits panicking stubs for any target it doesn't support. However, during runtime this would have failed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,15 +76,15 @@ Pull Requests that add support for new targets to `getrandom` are always welcome
 `getrandom` also provides optional (opt-in) backends, which allow users to customize the source
 of randomness based on their specific needs:
 
-| Backend name      | Target               | Target Triple        | Implementation
-| ----------------- | -------------------- | -------------------- | --------------
-| `linux_getrandom` | Linux, Android       | `*‑linux‑*`          | [`getrandom`][1] system call (without `/dev/urandom` fallback). Bumps minimum supported Linux kernel version to 3.17 and Android API level to 23 (Marshmallow).
-| `linux_rustix`    | Linux, Android       | `*‑linux‑*`          | Same as `linux_getrandom`, but uses [`rustix`] instead of `libc`.
-| `rdrand`          | x86, x86-64          | `x86_64-*`, `i686-*` | [`RDRAND`] instruction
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -76,15 +76,15 @@ Pull Requests that add support for new targets to `getrandom` are always welcome
...
+| Backend name      | Target               | Target Triple            | Implementation
+| ----------------- | -------------------- | ------------------------ | --------------
+| `linux_getrandom` | Linux, Android       | `*‑linux‑*`              | [`getrandom`][1] system call (without `/dev/urandom` fallback). Bumps minimum supported Linux kernel version to 3.17 and Android API level to 23 (Marshmallow).
+| `linux_rustix`    | Linux, Android       | `*‑linux‑*`              | Same as `linux_getrandom`, but uses [`rustix`] instead of `libc`.
+| `rdrand`          | x86, x86-64          | `x86_64-*`, `i686-*`     | [`RDRAND`] instruction
+| `rndr`            | AArch64              | `aarch64-*`              | [`RNDR`] register
+| `esp_idf`         | ESP-IDF              | `*‑espidf`               | [`esp_fill_random`]. WARNING: can return low-quality entropy without proper hardware configuration!
+| `wasm_js`         | Web Browser, Node.js | `wasm32‑unknown‑unknown` | [`Crypto.getRandomValues`] if available, then [`crypto.randomFillSync`] if on Node.js (see [WebAssembly support])
```
daxpedda, line 86: `wasm-bindgen` only supports `wasm32-unknown-unknown` (and now `wasm32v1-none`, see #541), it does not support Emscripten. Even though Emscriptens target is `wasm32-unknown-emscripten`, I'm not what else was being implied here.

newpavlov: `wasm_js` works only on `target_os = "unknown"` targets, so IIUC it can not be used with Emscripten.

newpavlov (approved):

README.md
daxpedda: Yep. I was referring to the fact that it previously said `wasm*-*-unknown`, implying the `target_os` could be something else.

PR #552, opened 2024-11-29
Run Wasm tests in workers as well

### LINKED ISSUE #319: Test that `getrandom` works correctly in a (service) worker in web environments.
Right now the testing for browser environments is "whatever wasm-bindgen-test does." There are several different contexts in which WebAssembly code can be run, and it would be good to be able to test that `getrandom` is working in all those contexts (extensions, workers), not just in the default execution environmet in a web page. I don't know how to do this testing though. cc @oliverdunk and see https://github.com/briansmith/ring/pull/1508 for background.

daxpedda:
This change runs Wasm tests in workers as well.

- Fix Node tests not actually running any tests. It was already broken because of an interaction with the latest Rust version and an old Node.js version.
- Test dedicated, shared and service worker by using the newly added environment variables: https://github.com/rustwasm/wasm-bindgen/pull/4295.

Fixes #319.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -240,16 +240,14 @@ jobs:
 
   web:
     name: Web
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
-      - run: choco install wget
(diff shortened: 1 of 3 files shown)


daxpedda:
Apparently the Node.js was broken already. I unfortunately don't know enough about Node.js, locally it seems to work for me, so I'm going to do some tests with the CI.

PR #554, opened 2024-12-02
wasm_js: remove IE 11 workaround

newpavlov:
Internet Explorer is superseded by Edge (which has 3-5% usage share) and effectively no longer supported. Citing Wikipedia:

>For SAC versions of Windows 10, Internet Explorer 11 support ended on June 15, 2022, Internet Explorer was permanently disabled on February 14, 2023, and any remaining icons or shortcuts were due to be removed on June 13, 2023.

Technically, it's still supported on LTSC, but Microsoft discourages the use of LTSC editions outside of "special-purpose devices" that perform a fixed function and thus do not require new user experience features, so I think we can ignore it.


newpavlov:
cc @daxpedda

daxpedda:
I don't really have an informed opinion on this, but I'm in favor!

PR #555, opened 2024-12-04
netbsd: test fallback using configuration flag

### LINKED ISSUE #289: For platforms that implement runtime fallback logic, all paths should be tested

newpavlov:
Adds the `getrandom_test_netbsd_fallback` configuration flag, which forces the NetBSD backend to use the KERN_ARND fallback, and tweaks CI to use the flag.

Closes #289


newpavlov:
Huh... It looks our NetBSD fallback code was incorrect. `sysctl` returns either 0 or -1, while the code thought that on success it returns number of bytes filled. Surprisingly, this bug was not reported for all this time, so I guess virtually no one uses Rust on older NetBSDs.

PR #557, opened 2024-12-04
wasm_js: remove the separate codepath for Node.js and TLS caching

newpavlov:
Acquire of "global" objects is relatively cheap (adds ~40 ns of overhead), so it make sense to do it on each call instead of trying to cache objects in a TLS variable.

Additionally, removes the separate codepath for Node.js since it fully supports the Web Crypto API since v19 (released 2022-10-18).


newpavlov:
@daxpedda 
Can you take a look? If you agree with the changes and have no suggestions, I will merge it.

I think we can leave additional optimization suggested by you [here](https://github.com/rust-random/getrandom/pull/541#issuecomment-2518445859) for a separate PR since I do not fully understand how to implement them (I am not very familiar with `wasm-bindgen`).

daxpedda:
Ah, I thought you are going to merge #541 first.
Sure, I will take a look.

README.md
```diff
@@ -83,7 +83,7 @@ of randomness based on their specific needs:
 | `rdrand`          | x86, x86-64          | `x86_64-*`, `i686-*`     | [`RDRAND`] instruction
 | `rndr`            | AArch64              | `aarch64-*`              | [`RNDR`] register
 | `esp_idf`         | ESP-IDF              | `*‑espidf`               | [`esp_fill_random`]. WARNING: can return low-quality entropy without proper hardware configuration!
-| `wasm_js`         | Web Browser, Node.js | `wasm32‑unknown‑unknown` | [`Crypto.getRandomValues`] if available, then [`crypto.randomFillSync`] if on Node.js (see [WebAssembly support])
+| `wasm_js`         | Web Browser, Node.js | `wasm32‑unknown‑unknown`, `wasm32v1-none` | [`Crypto.getRandomValues`] if available, then [`crypto.randomFillSync`] if on Node.js (see [WebAssembly support])
```
daxpedda, line 86:
```suggestion
| `wasm_js`         | Web Browser, Node.js | `wasm32‑unknown‑unknown`, `wasm32v1-none` | [`Crypto.getRandomValues`]
```
The corresponding links in the bottom should be removed as well.

newpavlov:
I thought that when you wrote "Sure, I'm a fan of splitting things up." you meant that we should continue with this PR. :)

src/backends/wasm_js.rs
```diff
@@ -1,158 +1,55 @@
...
+    Ok(())
 }
 
 #[wasm_bindgen]
 extern "C" {
     // Return type of js_sys::global()
     type Global;
-
     // Web Crypto API: Crypto interface (https://www.w3.org/TR/WebCryptoAPI/)
     type WebCrypto;
```
daxpedda, line 48:
```suggestion
    type Crypto;
```
No need to differentiate it now. Would also align with what its actually called in JS.

src/backends/wasm_js.rs
```diff
@@ -1,158 +1,55 @@
...
+#[cfg(not(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none"))))]
 compile_error!("`wasm_js` backend can be enabled only for OS-less WASM targets!");
 
-use js_sys::{global, Function, Uint8Array};
+use js_sys::{global, Uint8Array};
 use wasm_bindgen::{prelude::wasm_bindgen, JsCast, JsValue};
 
 // Size of our temporary Uint8Array buffer used with WebCrypto methods
 // Maximum is 65536 bytes see https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
 const WEB_CRYPTO_BUFFER_SIZE: u16 = 256;
```
daxpedda, line 15:
```suggestion
const CRYPTO_BUFFER_SIZE: u16 = 256;
```
Same.

src/backends/wasm_js.rs
```diff
@@ -1,158 +1,55 @@
...
-    })
-}
-
-fn getrandom_init() -> Result<RngSource, Error> {
     let global: Global = global().unchecked_into();
-
-    // Get the Web Crypto interface if we are in a browser, Web Worker, Deno,
-    // or another environment that supports the Web Cryptography API. This
-    // also allows for user-provided polyfills in unsupported environments.
     let crypto = global.crypto();
```
daxpedda, line 19: The `JsValue::is_object()` check was probably still useful to detect non-supporting environments. However we are safe from panicking because `WebCrypto::get_random_values()` should catch that, but we just catch it later and the error message is not as clear.

In this case we should still continue to return `Error::WEB_CRYPTO`.

daxpedda (changes requested):
There are also some Node.js related leftovers on the `Error` type.

> I thought that when you wrote "Sure, I'm a fan of splitting things up." you meant that we should continue with this PR. :)

Apologies, I probably didn't read carefully enough. I understood it just as a follow-up, not a replacement.

README.md
```diff
@@ -83,7 +83,7 @@ of randomness based on their specific needs:
 | `rdrand`          | x86, x86-64          | `x86_64-*`, `i686-*`     | [`RDRAND`] instruction
 | `rndr`            | AArch64              | `aarch64-*`              | [`RNDR`] register
 | `esp_idf`         | ESP-IDF              | `*‑espidf`               | [`esp_fill_random`]. WARNING: can return low-quality entropy without proper hardware configuration!
-| `wasm_js`         | Web Browser, Node.js | `wasm32‑unknown‑unknown` | [`Crypto.getRandomValues`] if available, then [`crypto.randomFillSync`] if on Node.js (see [WebAssembly support])
```
daxpedda, line 86: ``[`crypto.randomFillSync`]`` and `[WebASsembly support]` still has to be removed from the bottom of the README.

daxpedda (approved):

daxpedda:
We entirely forgot the whole `no_std` support.
Will make a PR for that.

PR #559, opened 2024-12-05
Web optimizations

daxpedda:
As discussed in #541.


daxpedda:
There is a strange bug in Firefox where using `instanceof SharedArrayBuffer` doesn't seem to work correctly in shared workers. This is probably related to [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1613912), but I will open a dedicated bug report tomorrow.

In the meantime I implemented a workaround where I check the name of the constructor instead.

daxpedda:
Alright, I just remembered that I encountered this issue before. So the problem is that `SharedArrayBuffer` is not available without COOP & COEP. But it is still possible to create a shared `WebAssembly.Memory` backed by a `SharedArrayBuffer` without COOP & COEP. So the way to check for that is still via the constructor name, which is still in place.

We can detect COOP & COEP via [`crossOriginIsolated`](https://developer.mozilla.org/en-US/docs/Web/API/Window/crossOriginIsolated), in which case we can use the simpler `instanceof SharedArrayBuffer` check.

See https://github.com/w3c/ServiceWorker/pull/1545 and Bugzilla [#1](https://bugzilla.mozilla.org/show_bug.cgi?id=1883145) and [#2](https://bugzilla.mozilla.org/show_bug.cgi?id=1613912).

daxpedda:
Alright, after playing around with the implementation and accounting for various scenarios, I decided its probably best if we just always check the constructor name for simplicities sake. Let me know what you think otherwise.

newpavlov:
This is a fair amount of additional complexity. It would be nice to provide performance measurements to demonstrate that this optimization is worth the trouble.

And wouldn't the unconditional dependency on `std` break the `wasm32v1-none` support?

daxpedda:
> This is a fair amount of additional complexity. It would be nice to provide performance measurements to demonstrate that this optimization is worth the trouble.

The first change, caching the `Crypto` global object, is a fairly minimal change. All it does is reduce the amount of time getting the object from 35-40ns to 0. While I don't think this is a big improvement, the change is basically two lines of code and the code is actually shorter then before. (apart from the workaround fixed by https://github.com/rustwasm/wasm-bindgen/pull/4315)

The second change is not using an `Uint8Array` unless necessary. It is only necessary when the underlying Wasm memory is a `SharedArrayBuffer`. In any other case, the most common case, the random bytes can be copied right into the Wasm memory. This not only reduces the amount of FFI calls from 4 to 1, but it also avoids allocating the necessary memory, which we don't cache anymore, and the byte copying into the Wasm memory.

I did a quick local benchmark:
- Firefox: 1174ns to 695ns, ~41% improvement
- Chrome: 1326ns to 781ns, ~41% improvement
- Node.js v23: 2362ns to 752ns, ~68% improvement

Here is the code (no `black_box()` shenanigans needed because its all FFI calls):
```rust
const RUNS: u32 = 10_000_000;
const CRYPTO_BUFFER_SIZE: u32 = 256;

fn main() {
    let global: Global = js_sys::global().unchecked_into();
    let crypto = global.crypto();

    let mut buffer = [0; CRYPTO_BUFFER_SIZE as usize];
    let start = Instant::now();

    for _ in 0..RUNS {
        crypto.get_random_values_ref(&mut buffer).unwrap();
    }

    console::log_1(&format!("{:?}", start.elapsed() / RUNS).into());

    let mut buffer = [0; CRYPTO_BUFFER_SIZE as usize];
    let start = Instant::now();

    for _ in 0..RUNS {
        let js_buffer = Uint8Array::new_with_length(CRYPTO_BUFFER_SIZE);
        let js_buffer = js_buffer.subarray(0, CRYPTO_BUFFER_SIZE);
        crypto.get_random_values(&js_buffer).unwrap();
        unsafe { js_buffer.raw_copy_to_ptr(buffer.as_mut_ptr()) };
    }

    console::log_1(&format!("{:?}", start.elapsed() / RUNS).into());
}
```
I will let you decide on how representative this is for actual real world usage.

I assume the most offensive part of the code additions is the `SharedArrayBuffer` detection, which basically manually implements a `LazyLock` because we don't have access to one in `no_std`. The MSRV is not necessary a problem because we could use `once_cell`, which is a dependency of `wasm-bindgen` anyway.

I briefly tested just not caching it and doing the detection on every single call. It would still be an improvement, but we loose around 10% of the performance gain.

Let me know what you think.

> And wouldn't the unconditional dependency on `std` break the `wasm32v1-none` support?

This is a workaround for a macro-hygiene bug in `wasm-bindgen` fixed in https://github.com/rustwasm/wasm-bindgen/pull/4315, we can simply gate it behind a `cfg(feature = "std")`. Which I will do now in the rebase.

daxpedda:
I removed the caching of the `SharedArrayBuffer` detection, we can discuss adding that in a follow-up PR.
Also added another optimization: not creating a sub-array unless necessary, saving another FFI call.

Let me know if you want me to split out the `SharedArrayBuffer` detection entirely.

daxpedda:
Alright, I made https://github.com/rustwasm/wasm-bindgen/pull/4325 to solve this issue. Its a dance I've done a couple of times already and its pretty bad. The Wasm module already contains all that information so we will just let `wasm-bindgen` extract it for us.

So this will wait for a new release of `wasm-bindgen`.
Let me know what you think of the changes now.

daxpedda:
I decided to remove the `SharedArrayBuffer` detection entirely. It was discussed in the past to support something like this in `wasm-bindgen`, but it was too niche. If you need `SharedArrayBuffer` you have to compile with `target_feature = "atomics"`. This would also be according to spec.

daxpedda:
This PR is ready again. I updated the OP so all the relevant information can be found there.

.github/workflows/tests.yml
```diff
@@ -229,11 +229,27 @@ jobs:
...
+        rust:
+          - { version: stable, atomics: false }
+          - {
+              description: with Atomics,
+              atomics: true,
+              version: nightly,
+              components: rust-src,
+              flags: '-Ctarget-feature=+atomics,+bulk-memory',
+              args: '-Zbuild-std=panic_abort,std',
+            }
```
josephlr, line 246: I think we can make this `matrix` more readable by doing:
```yaml
- {
    description: Web
    version: stable
    flags: '-Dwarnings --cfg getrandom_backend="wasm_js"'
    args: '--features=std'
  }
- {
    description: Web with Atomics
    version: nightly
    flags: '-Dwarnings --cfg getrandom_backend="wasm_js" -Ctarget-feature=+atomics,+bulk-memory'
    args: '--features=std -Zbuild-std=panic_abort,std',
  }
```

And then having simpler YAML below:
```yaml
name: ${{ matrix.rust.description }}
```
```yaml
env:
  RUSTFLAGS: ${{ matrix.rust.flags }}
```
```yaml
run:
  wasm-pack test --node -- ${{ matrix.rust.args }}
```

This makes it way easier to tell at a glance what's going on and which args/flags interact.

.github/workflows/workspace.yml
```diff
@@ -49,6 +49,10 @@ jobs:
       env:
         RUSTFLAGS: -Dwarnings --cfg getrandom_backend="wasm_js"
       run: cargo clippy -Zbuild-std --target wasm32-unknown-unknown
+    - name: Web WASM with atomics (wasm_js.rs)
+      env:
+        RUSTFLAGS: -Dwarnings --cfg getrandom_backend="wasm_js" -Ctarget-feature=+atomics,+bulk-memory
```
josephlr, line 54: What's the reason for needing `bulk-memory` here?

josephlr (commented):
Thanks @daxpedda for trying to help us clean this code up. I also really like the idea of using `target_feature = "atomics"` to check for if we need to care about shared memory or not. 

I have some suggestions below on how to furthur simplify.

src/backends/wasm_js.rs
```diff
@@ -7,53 +7,68 @@ pub use crate::util::{inner_u32, inner_u64};
...
     type Crypto;
-    // Getters for the Crypto API
-    #[wasm_bindgen(method, getter)]
-    fn crypto(this: &Global) -> Crypto;
+    // Holds the global `Crypto` object.
+    #[wasm_bindgen(thread_local_v2, js_name = crypto)]
+    static CRYPTO: Option<Crypto>;
     // Crypto.getRandomValues()
+    #[cfg(not(target_feature = "atomics"))]
+    #[wasm_bindgen(method, js_name = getRandomValues, catch)]
```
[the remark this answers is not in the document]
daxpedda: You are forfeiting the check if `Crypto` is available though.
But if your intention was just to get rid of the nesting, how about prefacing the function like this:
```rust
if CRYPTO.with(Option::is_none) {
    return Err(Error::WEB_CRYPTO)
}

// use `get_random_values()` without nesting:
get_random_values(&buf);
```
Let me know how you want to proceed.

My recommendation is to keep the check in place, there's all sort of strange JS environments out there, `getrandom` is very foundational and should support these imo. And `CRYPTO` is cached after all so there is no performance overhead.

.github/workflows/workspace.yml
daxpedda: LLVM requires the bulk memory proposal when using the threads (called "atomics") proposal.
However, I can see that LLVM does this check only on the linker level, so its not required for Clippy.

Let me know if you want me to remove it.

.github/workflows/tests.yml
daxpedda: Done.

.github/workflows/workspace.yml
josephlr: Makes sense to keep as-is. We should keep the flags the same between clippy and testing. Thanks for the explanation.

src/backends/wasm_js.rs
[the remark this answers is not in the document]
josephlr: Checking if the global `crypto` object exists doesn't really change what platforms/environments we support. It just changes what error message we get when the call to `crypto.getRandomValues` fails. Before, we were checking for existence of the object as part of our platform detection (and to cache the value in `RngSource`). Now that we don't have two different implementations for Web vs Node, we don't need to do any of that.

So if I'm understanding things correctly, having the check for the existence of the object just lets us return `Error::WEB_CRYPTO` to the user, informing them that `"Web Crypto API is unavailable"` instead of my implementation above which would make us return `Error::WEB_GET_RANDOM_VALUES` in all places.

I think returning good error messages is valuable, so I think keeping some sort of check is OK, provided that we aren't making an additional FFI call every time (which I don't think we do because the result of  `is_none` is cached).

I personally find the un-nested version easier to read, so we might try a version that avoids too much nesting (potentially by using a helper function).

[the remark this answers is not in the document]
daxpedda: > Checking if the global `crypto` object exists doesn't really change what platforms/environments we support. It just changes what error message we get when the call to `crypto.getRandomValues` fails.

Ah, you are relying on the fact that WBG just catches any exception when tagging it with `catch`. While this is currently the case, its not intended to be used like this. Ideally, this would only catch exceptions thrown by the function. However, I guess this behavior has to stay in place because otherwise it would be a breaking change.

> I personally find the un-nested version easier to read, so we might try a version that avoids too much nesting (potentially by using a helper function).

Sure!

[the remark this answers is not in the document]
daxpedda: Alright, let me know if you like this better!

src/backends/wasm_js.rs
```diff
@@ -7,32 +7,48 @@ pub use crate::util::{inner_u32, inner_u64};
...
+#[cfg(target_feature = "atomics")]
+use js_sys::Uint8Array;
+use wasm_bindgen::{prelude::wasm_bindgen, JsValue};
 
-// Size of our temporary Uint8Array buffer used with WebCrypto methods
-// Maximum is 65536 bytes see https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
-const CRYPTO_BUFFER_SIZE: u16 = 256;
+// Maximum buffer size allowed in `Crypto.getRandomValuesSize` is 65536 bytes.
+// See https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
+const MAX_BUFFER_SIZE: u16 = 256;
```
daxpedda, line 16: This is still an artifact of when we used to cache the `Uint8Array`. I assume this was lowered from 65536 so we don't hog 64kB in the background even when that's not needed.

For the happy path, non-atomic, this is pointless ofc because we don't actually create any buffer.

For atomic, the question is if it actually makes sense to create a small buffer and call `getRandomValues()` more often or is it better to create the maximum bytes `getRandomValues()` in favor of reducing the amount of calls.

So I did some local benchmarking (I apologize for the formatting):

| bytes |   base   |       64        |       128       |       256       |       512       |      1024       |      2048       |      4096       |      8192       |      16384      |      32768      |      65536      |
| ----: | -------: | --------------: | --------------: | --------------: | --------------: | --------------: | --------------: | --------------: | --------------: | --------------: | --------------: | --------------: |
|     8 |    435ns |   1.152µs  265% |                 |                 |                 |                 |                 |                 |                 |                 |                 |                 |
|    16 |    482ns |   1.229µs  255% |                 |                 |                 |                 |                 |                 |                 |                 |                 |                 |
|    32 |    555ns |   1.297µs  234% |                 |                 |                 |                 |                 |                 |                 |                 |                 |                 |
|    64 |    601ns |   1.381µs  230% |                 |                 |                 |                 |                 |                 |                 |                 |                 |                 |
|   128 |    687ns |   2.378µs  346% |   1.483µs  216% |                 |                 |                 |                 |                 |                 |                 |                 |                 |
|   256 |    847ns |   4.274µs  505% |   2.527µs  298% |   1.686µs  199% |                 |                 |                 |                 |                 |                 |                 |                 |
|   512 |  1.202µs |   7.994µs  665% |   4.618µs  384% |   2.944µs  245% |   2.171µs  181% |                 |                 |                 |                 |                 |                 |                 |
|  1024 |  1.891µs |  15.772µs  834% |   8.797µs  465% |   5.412µs  286% |   3.824µs  202% |   3.112µs  165% |                 |                 |                 |                 |                 |                 |
|  2048 |   3.28µs |  30.857µs  941% |  17.182µs  524% |   10.23µs  312% |   6.952µs  212% |   5.459µs  166% |   4.973µs  152% |                 |                 |                 |                 |                 |
|  4096 |   6.07µs |  59.789µs  985% |  32.748µs  540% |   19.68µs  324% |   13.07µs  215% |  10.094µs  166% |   8.944µs  147% |  10.097µs  166% |                 |                 |                 |                 |
|  8192 | 11.594µs | 117.864µs 1017% |  64.888µs  560% |  38.477µs  332% |  25.404µs  219% |  19.133µs  165% |   16.29µs  141% |  16.951µs  146% |  17.949µs  155% |                 |                 |                 |
| 16384 | 22.497µs | 236.436µs 1051% | 125.872µs  560% |   75.82µs  337% |  49.249µs  219% |  36.608µs  163% |  31.173µs  139% |  29.977µs  133% |  29.801µs  132% |  32.536µs  145% |                 |                 |
| 32768 | 45.626µs | 456.666µs 1001% | 251.232µs  551% | 151.411µs  332% |  98.836µs  217% |   72.35µs  159% |   59.47µs  130% |  55.756µs  122% |  54.609µs  120% |  56.186µs  123% |   63.44µs  139% |                 |
| 65536 | 92.087µs |  935.13µs 1015% |  512.67µs  557% |  301.09µs  327% | 196.186µs  213% |  144.02µs  156% | 118.095µs  128% |  105.63µs  115% |  102.41µs  111% |   104.2µs  113% |  110.35µs  120% | 118.917µs  129% |

"bytes": how many bytes we are trying to get
"base": baseline time, representing 100%, by using the non-atomic path (with the fixed maximum chunk size of 65536)
all the other columns: atomic path and how many bytes we use for the `Uint8Array` buffer

I skipped testing for cases where we are trying retrieve less bytes than the buffer size, as it should result in the same time.

The result seems clear: taking the maximum amount of bytes and reducing the calls to `getRandomValues()` seems to always yield the best result.

So my proposal is:
- Fix the chunk size for the happy path to be the maximum size: 65536.
- As we don't cache the `Uint8Array`, we should not create one with a fixed size, but `min(dest.len(), FIXED_SIZE)`.
- Increase the maximum size of the buffer in the atomic path to 65536 as well.

---

Let me know if you want me to apply this here or in a follow-up.

newpavlov: I agree that we should use 64K chunk size for the non-atomic code path. But I think we should keep the 256 byte buffer for the atomic path. The main use-case for `getrandom` is generation of seeds and cryptographic material, so it's rarely used for generation of big random buffers. For example, in the [WASI p2 backend](https://github.com/rust-random/getrandom/blob/master/src/backends/wasi_p2.rs#L21-L25) we even use API which generates 64-bit integers instead of the buffer-based one.

src/backends/wasm_js.rs
```diff
@@ -7,32 +7,48 @@ pub use crate::util::{inner_u32, inner_u64};
...
         let chunk_len: u32 = chunk
             .len()
             .try_into()
-            .expect("chunk length is bounded by CRYPTO_BUFFER_SIZE");
+            .expect("chunk length is bounded by MAX_BUFFER_SIZE");
         // The chunk can be smaller than buf's length, so we call to
         // JS to create a smaller view of buf without allocation.
-        let sub_buf = buf.subarray(0, chunk_len);
+        let sub_buf = if chunk_len == u32::from(MAX_BUFFER_SIZE) {
+            buf.clone()
```
newpavlov, line 48: Does this mean that we create a new allocation here instead of reusing the existing one?

src/backends/wasm_js.rs
```diff
@@ -7,32 +7,48 @@ pub use crate::util::{inner_u32, inner_u64};
 #[cfg(not(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none"))))]
 compile_error!("`wasm_js` backend can be enabled only for OS-less WASM targets!");
 
-use js_sys::{global, Uint8Array};
-use wasm_bindgen::{prelude::wasm_bindgen, JsCast, JsValue};
+#[cfg(target_feature = "atomics")]
+use js_sys::Uint8Array;
```
newpavlov, line 11: It's better to move this import to the atomic `inner` function.

Cargo.toml
```diff
@@ -64,8 +64,8 @@ windows-targets = "0.52"
 
 # wasm_js
 [target.'cfg(all(getrandom_backend = "wasm_js", target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))'.dependencies]
-wasm-bindgen = { version = "0.2.96", default-features = false }
-js-sys = { version = "0.3.73", default-features = false }
+wasm-bindgen = { version = "0.2.98", default-features = false }
+js-sys = { version = "0.3.75", default-features = false }
```
newpavlov, line 68: IIUC `js-sys` is used only in the atomic backend.

newpavlov (commented):
Gating on `target_feature = "atomics"` looks a bit surprising to me, but I am not familiar with the `wasm-bindgen` and the wider WASM ecosystem, so I can not make an informed decision here.

src/backends/wasm_js.rs
daxpedda: No, cloning the buffer just clones the reference, not actually the whole buffer.
This is just here to save the FFI call to `subarray()`.

I did it like this because our MSRV of v1.63 doesn't support [propagating a temporary lifetime extension into if](https://github.com/rust-lang/rust/pull/121346), but now that we have gated this behind `target_feature = "atomics"` this isn't MSRV bound anymore.

src/backends/wasm_js.rs
daxpedda: Its required in the bindings on the bottom of the file as well.
I could:
- Move it into the atomic `inner()` function and use the full path when importing.
- Move the atomic only import into the atomic `inner()` function.
- Leave it as is.

Let me know.

src/backends/wasm_js.rs
daxpedda: This is what I'm proposing to account for this:
```rust
    let buf_len = u32::min(
        dest.len().try_into().unwrap_or(MAX_BUFFER_SIZE.into()),
        MAX_BUFFER_SIZE.into(),
    );
    let buf = Uint8Array::new_with_length(buf_len);
```
So we will always use the minimum amount necessary. This is an improvement over the current state already, because right now we allocate 256 bytes even if we only need 8 bytes.

src/backends/wasm_js.rs
daxpedda: Done.

Cargo.toml
daxpedda: Done.

daxpedda:
> Gating on `target_feature = "atomics"` looks a bit surprising to me, but I am not familiar with the `wasm-bindgen` and the wider WASM ecosystem, so I can not make an informed decision here.

I'm happy to answer any questions, but here is some context:

The base problem is that [`Crypto.getRandomValues()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) doesn't support accepting a `SharedArrayBuffer`. You can see this by looking at the  [`Crypto.getRandomValues()` interface specification](https://www.w3.org/TR/2017/REC-WebCryptoAPI-20170126/#dfnReturnLink-0) and not finding a [`[AllowShared]` extended attribute on the `array` parameter](https://webidl.spec.whatwg.org/commit-snapshots/20f6212d8d9c5fdd5a6e52c6127c0c38b32b740a/#AllowShared).

The underlying Wasm memory in browsers is represented by a [`WebAssembly.Memory`](https://www.w3.org/TR/2024/WD-wasm-js-api-2-20241206/#memories) and its underlying buffer is a [`ArrayBuffer`](https://www.w3.org/TR/2024/WD-wasm-js-api-2-20241206/#ref-for-identified-with). To use a `SharedArrayBuffer` instead, the [`shared` property can be used](https://github.com/WebAssembly/threads/blob/b2567bff61ee6fbe731934f0ed17a5d48dc9ab01/proposals/threads/Overview.md#javascript-api-changes).

This is connected to the [Wasm threads proposal](https://github.com/WebAssembly/threads/tree/b2567bff61ee6fbe731934f0ed17a5d48dc9ab01) (toolchains use the target feature `atomics`) because it introduces the [`shared` memory type](https://webassembly.github.io/threads/core/bikeshed/#syntax-memtype). Now, [quoting the threads proposal](https://github.com/WebAssembly/threads/blob/b2567bff61ee6fbe731934f0ed17a5d48dc9ab01/proposals/threads/Overview.md#shared-linear-memory) (this will eventually land in a separate W3C JS specification, but the proposal isn't there yet):
> It is a validation error to attempt to import shared linear memory if the module's memory import doesn't specify that it allows shared memory. Similarly, it is a validation error to import non-shared memory if the module's memory import specifies it as being shared.

This is to say that without marking the Wasm memory as shared, which requires the `atomics` target feature, the underlying memory can never be a `SharedArrayBuffer`.

Lastly, Rust [by default marks the memory as shared when the `atomics` target feature is active](https://github.com/rust-lang/rust/blob/9c707a8b769523bb6768bf58e74fa2c39cc24844/compiler/rustc_codegen_ssa/src/back/linker.rs#L1286-L1287). In conclusion: when the `atomics` target feature is enabled, the memory is marked as shared and therefor MUST be backed by a `SharedArrayBuffer` in JS.

Within the spec, there is one thing we might want to consider: it is allowed to enable the `atomics` target feature without using a shared memory and therefor using a regular `ArrayBuffer`. To account for that, we might want to detect that and run the optimized code path. However, this use case is not supported by Rust or `wasm-bindgen` and users would have to do some post-processing to pull off something like that. If `wasm-bindgen` ever gains support for that we will offer a detection mechanism: see https://github.com/rustwasm/wasm-bindgen/pull/4325.

Outside the spec, the other case was using a `SharedArrayBuffer` without the `atomics` target feature. This isn't possible because browsers require the memory marked as `shared`. So this also requires post-processing and is not supported by `wasm-bindgen` and I don't see a need for `getrandom` to support something hacky like that either.

src/backends/wasm_js.rs
newpavlov: Ah, true. Then we can leave it as-is.

src/backends/wasm_js.rs
newpavlov: Looks good to me!

daxpedda: Done.

daxpedda:
CI failure is unrelated.

src/backends/wasm_js.rs
```diff
@@ -7,34 +7,54 @@ pub use crate::util::{inner_u32, inner_u64};
...
+use wasm_bindgen::{prelude::wasm_bindgen, JsValue};
 
-// Size of our temporary Uint8Array buffer used with WebCrypto methods
-// Maximum is 65536 bytes see https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
-const CRYPTO_BUFFER_SIZE: u16 = 256;
+// Maximum buffer size allowed in `Crypto.getRandomValuesSize` is 65536 bytes.
+// See https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
+const MAX_BUFFER_SIZE: u32 = 65536;
```
newpavlov, line 16: I think it's better to use `usize` here since it will allow us to remove several `as` casts.

newpavlov (approved):

src/backends/wasm_js.rs
daxpedda: Unfortunately it won't save us from casting. We need it in `u32` for all `Uint8Array` methods. So we would just be trading one set of `as` casts with another.

I tried out adding `const MAX_BUFFER_USIZE: usize = MAX_BUFFER_SIZE as usize;` which is checked during compile time. But I found it a bit strange.

Let me know if you like that idea.

newpavlov: You can write the following:
```rust
let buf_len = usize::min(dest.len(), MAX_BUFFER_SIZE);
let buf_len_u32 = buf_len.try_into().expect("...");
let buf = Uint8Array::new_with_length(buf_len_u32);
for chunk in dest.chunks_mut(buf_len) { ... }
```

Number of `TryInto` casts stays the same and you can remove two `as` casts.

daxpedda: Nice! We have 0 `as` casts now!

src/backends/wasm_js.rs
```diff
@@ -46,14 +66,16 @@ pub fn fill_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
...
 extern "C" {
-    // Return type of js_sys::global()
-    type Global;
     // Web Crypto API: Crypto interface (https://www.w3.org/TR/WebCryptoAPI/)
     type Crypto;
-    // Getters for the Crypto API
-    #[wasm_bindgen(method, getter)]
-    fn crypto(this: &Global) -> Crypto;
+    // Holds the global `Crypto` object.
+    #[wasm_bindgen(thread_local_v2, js_name = crypto)]
```
daxpedda, line 72: I initially avoided using `globalThis` here because I have no idea how it works exactly on Node.js. In the meantime I have looked up the documentation and asked around: it should be perfectly fine to use `globalThis` in Node.js.

For context: the current setup just accesses the object directly from the global scope. According to the spec this is perfectly fine. However, this is usually discouraged because global variables can simply be overwritten. While I'm not sure if there is more history behind this, fact is that read-only properties like `crypto` can't be overwritten. So both approaches, accessing `Crypto` through the global `crypto` variable or `globalThis.crypto`, will work the same.

But to stick with "how it should be done" I will add `globalThis` now.

src/backends/wasm_js.rs
newpavlov: BTW you also could just use `js_sys::Uint8Array` in both places without importing it.

daxpedda: Right, let me know if you prefer that route.

src/backends/wasm_js.rs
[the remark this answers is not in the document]
newpavlov: Maybe we can just add a separate check with caching?
```rust
const CRYPTO_PRESENT: u8 = 0;
const CRYPTO_NOT_PRESENT: u8 = 1;
const CRYPTO_UNINIT: u8 = 255;

static CRYPTO_CHECK: AtomicU8 = AtomicU8::new(CRYPTO_UNINIT);

pub fn fill_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
    match CRYPTO_CHECK.load(Relaxed) {
        CRYPTO_PRESENT => {}
        CRYPTO_NOT_PRESENT => return Err(Error::WEB_CRYPTO),
        _ => {
            let global: Global = global().unchecked_into();
            let crypto = global.crypto();
            if crypto.is_object() {
                CRYPTO_CHECK.store(CRYPTO_PRESENT, Relaxed);
            } else {
                CRYPTO_CHECK.store(CRYPTO_NOT_PRESENT, Relaxed);
                return Err(Error::WEB_CRYPTO);
            }
        }
    }
    // ...
}
```
IIUC the atomic types are still present on "non-atomic" WASM targets.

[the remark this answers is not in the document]
daxpedda: If the goal is to cache the check then that's not necessary. `CRYPTO` is already cached in a thread-local, its result resides in Rust memory after the first call and it makes no FFI calls afterwards.
Let me know if I misunderstood you.

src/backends/wasm_js.rs
daxpedda: I prefer the code to be in the style maintainers prefer :)
I went ahead and applied your suggestion.

src/backends/wasm_js.rs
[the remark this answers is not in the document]
newpavlov: I meant it in the context of @josephlr's proposal to remove use of `thread_local_v2` and rely on `js_namespace` instead.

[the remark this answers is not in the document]
daxpedda: Ah, I see!

From my understanding the issue was getting rid of the nesting, but that has been achieved, right? Is there some other motivation to get rid of `thread_local_v2`?

On that note: your proposal will store the check globally. But different threads may have different outcomes. This isn't hypothetical either, some worklets don't have access to `Crypto`. So this needs to be in a thread-local to work correctly.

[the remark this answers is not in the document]
newpavlov: I am not familiar with the `wasm-bindgen` inner workings, but wouldn't `js_namespace`-based code result in a lighter-weight WASM file? Plus, the resulting source code will be slightly smaller (assuming that we do not care about the `WEB_CRYPTO` error).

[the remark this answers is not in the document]
daxpedda: Okay, that fast: [according to the specification](https://www.w3.org/TR/2017/REC-WebCryptoAPI-20170126/#Crypto-method-getRandomValues), `Crypto.getRandomValues()` can only fail if we don't supply the right buffer type or we exceed the maximum byte length of 65536. So instead of getting rid of `WEB_CRYPTO` I will go ahead and get rid of `WEB_GET_RANDOM_VALUES`.

Let me know what you think!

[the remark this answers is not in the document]
daxpedda: Done in https://github.com/rust-random/getrandom/pull/559/commits/8cf68328aa898966233496da384185744dd25d87.

newpavlov:
Thank you!

PR #560, opened 2024-12-05
Add `no_std` support for Web

daxpedda:
In #557 we forgot to port the `no_std` support from #541.

This also adds running Clippy for `wasm32v1-none`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ windows-targets = "0.52"
 # wasm_js
 [target.'cfg(all(getrandom_backend = "wasm_js", target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))'.dependencies]
 wasm-bindgen = { version = "0.2.96", default-features = false }
-js-sys = "0.3.73"
+js-sys = { version = "0.3.73", default-features = false }
 [target.'cfg(all(getrandom_backend = "wasm_js", target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))'.dev-dependencies]
 wasm-bindgen-test = "0.3"
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -131,6 +131,40 @@ jobs:
           RUSTFLAGS: -Dwarnings --cfg getrandom_backend="rdrand"
         run: cargo build --features=std
 
+  web:
+    name: ${{ matrix.target.description }} ${{ matrix.feature.description }} ${{ matrix.atomic.description }}
+    runs-on: ubuntu-24.04
+    env:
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -65,14 +65,14 @@ windows-targets = "0.52"
 # wasm_js
 [target.'cfg(all(getrandom_backend = "wasm_js", target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))'.dependencies]
 wasm-bindgen = { version = "0.2.96", default-features = false }
-js-sys = "0.3.73"
+js-sys = { version = "0.3.73", default-features = false }
```
newpavlov, line 68: I think we only need this line and all other changes can be removed.

Cargo.toml
```diff
@@ -65,14 +65,14 @@ windows-targets = "0.52"
...
 [target.'cfg(all(getrandom_backend = "wasm_js", target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))'.dev-dependencies]
 wasm-bindgen-test = "0.3"
 
 [features]
 # Implement std::error::Error for getrandom::Error and
 # use std to retrieve OS error descriptions
-std = []
+std = ["wasm-bindgen/std", "js-sys/std"]
```
newpavlov, line 75: We do not use any items dependent on `std` from these crates, no? And users should not depend on `getrandom` to enable the `std` feature for these crates.

.github/workflows/workspace.yml
```diff
@@ -48,7 +48,23 @@ jobs:
     - name: Web WASM (wasm_js.rs)
       env:
         RUSTFLAGS: -Dwarnings --cfg getrandom_backend="wasm_js"
-      run: cargo clippy -Zbuild-std --target wasm32-unknown-unknown
+      run: cargo clippy --features std -Zbuild-std=panic_abort,std --target wasm32-unknown-unknown
+    - name: Web WASM no_std (wasm_js.rs)
+      env:
+        RUSTFLAGS: -Dwarnings --cfg getrandom_backend="wasm_js"
+      run: cargo clippy -Zbuild-std=core,alloc --target wasm32-unknown-unknown
+    - name: Web WASM no_std with atomics (wasm_js.rs)
```
newpavlov, line 56: These lines do not check any new code with Clippy. If you want to tests that the crate can be built for `wasm32v1-none`, you should do it in the build workflow.

Cargo.toml
daxpedda: I do really appreciate that you are pushing back on this actually.
Thanks!

newpavlov (approved):
Thank you!

PR #561, opened 2024-12-06
Web: cache `SharedArrayBuffer` detection

daxpedda:
This change caches the `SharedArrrayBuffer` detection. Ideally we would have simply used `once_cell::sync::Lazy` because `std::sync::OnceLock` is not compatible with our MSRV, but both are not compatible with `no_std`.

Follow-up to #559.
Based on #559.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -229,11 +229,27 @@ jobs:
   #         run: cargo test
 
   web:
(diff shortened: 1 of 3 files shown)


daxpedda:
Closed in favor of https://github.com/rustwasm/wasm-bindgen/pull/4325.

PR #562, opened 2024-12-09
Make target-specific errors private

newpavlov:
The main reason for existence of the target-specific errors is nicer error descriptions. Users arguably should not match on those error constants since our inner implementation may change together with possible error codes (for example, like with `IOS_SEC_RANDOM`). Marking the error constants as private we get flexibility to change error codes and names.

Additionally, the `internal_desc` change reduces generated binary size since compiler no longer has to include irrelevant error descriptions. Ideally the target-specific error descriptions would be defined in the relevant modules, but I couldn't find a good way to do it, so, unfortunately, I had to partially duplicate backend `cfg`s in `internal_desc`.

### DIFF
--- a/src/backends/apple_other.rs
+++ b/src/backends/apple_other.rs
@@ -10,6 +10,11 @@ pub fn fill_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
     if ret == libc::kCCSuccess {
         Ok(())
     } else {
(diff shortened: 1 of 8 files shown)


josephlr (approved):
This looks good and removes confusing stuff from the public API. I was also trying to figure out ways to put the descriptions in the platform-specific files, but I also don't know how to do this.

I'll keep thinking about it, but you should merge this in the meantime.

PR #565, opened 2024-12-13
Remove explicit docsrs configuration flag from docs.rs metadata

newpavlov:
The flag is passed automatically for all docs.rs builds. See: https://github.com/rust-lang/docs.rs/pull/2390


josephlr:
Whoops forgot to review this, LGTM.

PR #569, opened 2024-12-18
Allow `Error::raw_os_error` to return non-`i32` codes

### LINKED ISSUE #568: `Error` on targets where `RawOsError` is not equal to `i32`
Previously, all Rust targets with `std` were using `i32` for "raw" OS error codes, i.e. `io::Error::raw_os_error/from_raw_os_error` have accepted `i32`. We have encoded this assumption in `Error::raw_os_error`, `fmt::Debug/Display`, and `From<Error> for io::Error` implementations.

But in the recently added UEFI `std` implementation `usize` is used for error codes instead as per [UEFI spec](https://uefi.org/specs/UEFI/2.10/Apx_D_Status_Codes.html). It's reflected by the new (Nightly-only) type alias [`RawOsError`](https://doc.rust-lang.org/std/io/type.RawOsError.html)

For now, we just added `cfg`-based exceptions for UEFI targets in the relevant internal implementations (see #566). But it may be worth to reconsider our use of `NonZeroU32` for `Error` internal representation and return type of `Error::raw_os_error`.

cc @usamoi

newpavlov:
This allows us to remove some UEFI-specific exceptions from the code.

This PR only partially resolves #568 since we still use `NonZeroU32` on UEFI targets, but it should be sufficient to future-proof ourselves for the v0.3 release.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -3,6 +3,16 @@ extern crate std;
 
 use core::{fmt, num::NonZeroU32};
 
+// This private alias mirrors `std::io::RawOsError`:
+// https://doc.rust-lang.org/std/io/type.RawOsError.html)
+cfg_if::cfg_if!(
+    if #[cfg(target_os = "uefi")] {
+        type RawOsError = usize;
+    } else {
+        type RawOsError = i32;
+    }
+);
+
 /// A small and `no_std` compatible error type
 ///
 /// The [`Error::raw_os_error()`] will indicate if the error is from the OS, and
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -57,20 +67,27 @@ impl Error {
...
-            } else {
-                errno
-            }
-        })
+            #[cfg(target_os = "solid_asp3")]
+            let res = res.map(|errno| -errno);
+            res
+        } else {
+            None
+        }
```
josephlr, line 90: I like the simplification, might be a little more readable to do:
```rust
let code = self.0.get();
if code >= Self::INTERNAL_START {
    return None;
}
let errno = RawOsError::try_from(code).ok()?;
#[cfg(target_os = "solid_asp3")]
let errno = -errno;
return Some(errno)
```
just to reduce the indenting and fully get rid of the `map`.

josephlr (approved):

PR #571, opened 2024-12-18
Detect use of MemorySanitizer without using Nightly-only features

josephlr:
This allows msan detection to "just-work" whenever someone passes `-Zsanitizer=memory`. Users no longer need to do any `getrandom`-specific configuration.

This will also continue working once https://github.com/rust-lang/rust/issues/123615 is merged, which stabilizes some sanitizers (but not MemorySanitizer).

This is the approch taken by other low-level crates:
  - [`parking_lot_core`](https://github.com/Amanieu/parking_lot/blob/ca920b31312839013b4455aba1d53a4aede21b2f/core/build.rs)
  - [`crossbeam-utils`](https://github.com/crossbeam-rs/crossbeam/blob/00283fb1818174c25b02d7f1c883c5e19f8506a4/crossbeam-utils/build.rs#L42)

The only downside is that this adds a build-script, but it's as small as possible, doesn't seem to impact build times, and is only a temporary workaround.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,15 +267,14 @@ our code should correctly handle it and return an error, e.g.
 
 ## Sanitizer support
 
-If your code uses [`fill_uninit`] and you enable memory sanitization
-(i.e. `-Zsanitizer=memory`), you need to pass the `getrandom_sanitize`
-configuration flag to enable unpoisoning of the destination buffer
-filled by `fill_uninit`.
+If your code uses [`fill_uninit`] and you enable
+[MemorySanitizer](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html#memorysanitizer)
+(i.e. `-Zsanitizer=memory`), we will automatically handle unpoisoning
+of the destination buffer filled by `fill_uninit`.
 
-For example, it can be done as follows (requires a Nightly compiler):
+You can run sanitizer tests for your crate dependent on `getrandom` like this:
 ```sh
-RUSTFLAGS="-Zsanitizer=memory --cfg getrandom_sanitize" \
-    cargo test -Zbuild-std --target=x86_64-unknown-linux-gnu
+RUSTFLAGS="-Zsanitizer=memory" cargo test -Zbuild-std --target=x86_64-unknown-linux-gnu
 ```
 
 ## Minimum Supported Rust Version
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -140,9 +140,9 @@ jobs:
           toolchain: nightly-2024-10-08
           components: rust-src
       - env:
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -267,15 +267,16 @@ our code should correctly handle it and return an error, e.g.
...
+of the destination buffer filled by `fill_uninit`.
 
-For example, it can be done as follows (requires a Nightly compiler):
+The following command runs all our tests with MemorySanitizer
+(requires a Nightly compiler):
 ```sh
-RUSTFLAGS="-Zsanitizer=memory --cfg getrandom_sanitize" \
-    cargo test -Zbuild-std --target=x86_64-unknown-linux-gnu
+RUSTFLAGS="-Zsanitizer=memory" cargo test \
+    -Zbuild-std --target=x86_64-unknown-linux-gnu --all-targets
```
newpavlov, line 279: This line break can be probably removed.

newpavlov (approved):
Addition of a build script is quite unfortunate, but since this approach is used by other fundamental crates I guess it's acceptable.

README.md
josephlr: I initially had this because the doc tests weren't working, but I realized this can be fixed by just setting `RUSTDOCFLAGS`. I've updated the docs and the CI invocation.

With the increased environment variables, I decided to keep the line-break as otherwise the wrapping ends up being hard to read.

newpavlov: This line is intended for users, no? They probably do not need the `RUSTDOCFLAGS` flag.

newpavlov: It may be worth to write this part like this:
> You can run sanitizer tests for your crate dependent on `getrandom` like this:
> ```sh
> RUSTFLAGS="-Zsanitizer=memory" cargo test -Zbuild-std --target=x86_64-unknown-linux-gnu
> ```

josephlr: > It may be worth to write this part like this:
> 
> > You can run sanitizer tests for your crate dependent on `getrandom` like this:
> > ```shell
> > RUSTFLAGS="-Zsanitizer=memory" cargo test -Zbuild-std --target=x86_64-unknown-linux-gnu
> > ```

That invocation fails without also setting `RUSTDOCFLAGS="-Zsanitizer=memory"`, which is why I updated the CI to set that environment variable as well.

josephlr: Oh I'm being dumb, I understand what you mean. Our crate needs `RUSTDOCFLAGS`, but other crates may not.

josephlr: Fixed the README to use your suggestion.

newpavlov: Hm, it's a bit weird that we have to pass `RUSTDOCFLAGS` sedately. In that case, I am fine with both variants.

newpavlov:
It;s worth to update the changelog and change the configuration flag mentioned in it.

newpavlov (approved):

Issue #433, opened 2024-05-26
Release 0.3?


josephlr:
We have a few changes coming up (#425, #414 , #432) which will make the following changes to the compatibility of this crate:
  - Increase the Minimum Supported Rust Version (MSRV) from 1.36 (released July 4, 2019) to 1.56 (released October 21, 2021).
  - Increase the supported OS version of `*-pc-windows-{gnu,msvc}` from Windows Vista (released November 30, 2006) to Windows 10 (released July 15, 2015).
     - Rust increased its minimum Windows OS version from Windows 7 to Windows 10 in [February 26th, 2024](https://blog.rust-lang.org/2024/02/26/Windows-7.html).
     - Older versions of Windows are still supported under the `x86_64-win7-windows-msvc` and `i686-win7-windows-msvc` targets
     - We would intend to continue to support the `*-win7-*` targets (via [`RtlGenRandom`](https://learn.microsoft.com/en-us/windows/win32/api/ntsecapi/nf-ntsecapi-rtlgenrandom))
  - We are **not** making any changes to the API of this crate
  - We are **not** making any changes to the features of this crate

The [Rust Semver Compatibility Guide](https://doc.rust-lang.org/cargo/reference/semver.html) does not specify if such actions would be breaking changes. Both:
  - [changing the minimum version of Rust required](https://doc.rust-lang.org/cargo/reference/semver.html#env-new-rust)
  - [changing the platform and environment requirements](https://doc.rust-lang.org/cargo/reference/semver.html#env-change-requirements)

are listed as "Possibly-breaking" changes.

My personal preference would be to release these changes as `0.2.16` (not consider this a breaking change) because:
  - `getrandom` is widely used throughout the Rust ecosystem so updating all callers can be somewhat tedious.
  - It's not clear how a user can have a `custom` implementation work with both versions of `getrandom`.
  - Both the MSRV increases and the Windows platform increases are less disruptive than an Semver-breaking change.

However, if we do decide to have this be a breaking change, we should discuss if there are any other semver-breaking changes we wish to make (see #346 and #230). Another option would just be to get consensus on #232 and release these changes as 1.0

josephlr:
CC: @dhardy @vks for how this might affect `rand`

newpavlov:
>We are not making any changes to the API of this crate

[...]

Issue #424, opened 2024-05-17
Support `x86_64-unknown-linux-none`


briansmith:
The target was added in https://github.com/rust-lang/rust/pull/125023 and is available in the latest Nightly builds. Use `cargo +nightly build --target=x86_64-unknown-linux-none -Zbuild-std`.

The `libc` crate does NOT support this target, so `libc::syscall` and `libc::getrandom` cannot be used on this target.

I am planning to support this target in *ring* soon.

newpavlov:
I think #401 is relevant here. We need a way of doing raw syscalls for this. Ideally, it would be handled by the [`linux-raw-sys`](https://github.com/sunfishcode/linux-raw-sys) crate, but @sunfishcode is [hesitant](https://github.com/bytecodealliance/rustix/issues/1055) to add raw syscall functions to it.

josephlr:
I think if we want to support this target, we should just write the inline assembly to invoke the `x86_64` `syscall` instruction, and hardcode the raw syscall number (318).

I agre with @sunfishcode that supporting a general mechanism for invoking _any_ syscall on _any_ Linux CPU architecture is _very_ complicated (vDSO, some syscalls may or may not return, `vfork`, inline assembly isn't stable on all architectures supported by Rust, etc...).

But the situation for `getrandom(2) + x86_64` specifically is _much_ simpler, Linux only uses `syscall` on `x86_64`, and  `x86_64` inline assembly is stable. So the implementation can just be:
```rust
const NR_GETRANDOM: u32 = 318;

pub fn getrandom_syscall(buf: &mut [MaybeUninit<u8>], flags: u32) -> isize {
    unsafe {
        let ret;
        asm!(
            "syscall",
            inlateout("rax") NR_GETRANDOM as isize => ret,
            in("rdi") buf.as_mut_ptr(),
            in("rsi") buf.len(),
            in("rdx") flags,
            lateout("rcx") _,
            lateout("r11") _,
            options(nostack, preserves_flags)
        );
        ret
    }
}
```

newpavlov:
@josephlr 
Hm, you are right. If the raw target is limited to x86-64 it would be the easiest way to do this.

We probably can do it for other target arches as well (on x86 we can just use `int x80` for simplicity) and use gate it behind a hypothetical `linux_raw_syscall` feature.

sunfishcode:
[...]

PR #572, opened 2024-12-23
Add `linux_raw` opt-in backend

newpavlov:
This PR replaces the `linux_rustix` opt-in backend introduced in #520 with `linux_raw` opt-in backend based raw syscalls implemented using `asm!`. It noticeably reduces dependency footprint of the crate (see concerns raised in #433). Unfortunately, it means that we have to implement the raw syscalls ourselves.

This PR implements `linux_raw` support  only for target arches with stable `asm!`. Nightly-only support for other targets can be added in later PRs.


josephlr (commented):
Personally, I think this seems like a lot of added complexity for a backend that is not enabled by default. I think that we should try to find some other way to solve the issues raised in #433, or just not have a rustix backend right now.

Getting raw syscalls correct can be very complex if we want to support a range of targets, and I don't think we are well situated to maintain such things.

newpavlov:
>just not have a rustix/raw-syscall backend right now

Yes, it's a reasonable option. We can delay this backend until a later release.

But I believe we should have a "raw Linux syscall" backend as it was requested several times (#401, #424).

>Getting raw syscalls correct can be very complex if we want to support a range of targets, and I don't think we are well situated to maintain such things.

I do not agree with the "very complex" assessment. As mentioned in the comments, I've mostly copied the battle-tested code from `rustix`, intentionally limited the set of supported targets and did not include any vDSO complexity.

Ideally, I would prefer to have the syscall stuff in a separate lean crate, but, unfortunately, `rustix`/`linux-raw-sys` developers weren't open to such changes. I could create my own fork of `linux-raw-sys`, but would it be really better than the current solution?

newpavlov:
>this seems like a lot of added complexity for a backend that is not enabled by default

Note that I plan to use this backend to support the `*-linux-none` targets by default. I think it's better to do it in a separate PR.

newpavlov:
@josephlr
I would like to add this backend and support of `*-linux-none` in the next release. WDYT?

src/backends/linux_raw.rs
```diff
@@ -0,0 +1,127 @@
+//! Implementation for Linux / Android using `asm!`-based syscalls.
+use crate::{Error, MaybeUninit};
+
+pub use crate::util::{inner_u32, inner_u64};
+
+#[cfg(not(any(target_os = "android", target_os = "linux")))]
```
josephlr, line 6: I'm trying to figure out what the constraints are around directly invoking syscalls on Android. I can try to check with some people at work, but it's unclear if bionic libc (the libc shipped with Android) does any special handling for syscalls. If the only demand for this is on `linux` targets, it might just be best to initially exclude Android.

src/backends/linux_raw.rs
```diff
@@ -0,0 +1,127 @@
...
+            const __NR_getrandom: u32 = 278;
+            core::arch::asm!(
+                "ecall",
+                in("a7") __NR_getrandom,
+                inlateout("a0") buf => r0,
+                in("a1") buflen,
+                in("a2") flags,
+                options(nostack, preserves_flags)
+            );
+        } else if #[cfg(target_arch = "x86")] {
```
josephlr, line 65: Give that there is no `i686-unknown-linux-none` target and invoking raw syscalls on x86 is a notorious foot-gun, could we just not support 32-bit x86? It seems easier and then we don't have to have warnings saying "you probably shouldn't use this backend"

src/backends/linux_raw.rs
```diff
@@ -0,0 +1,127 @@
...
+            // In thumb-mode, r7 is the frame pointer and is not permitted to be used in
+            // an inline asm operand, so we have to use a different register and copy it
+            // into r7 inside the inline asm.
+            // Theoretically, we could detect thumb mode in the build script, but several
+            // register moves are cheap enough compared to the syscall cost, so we do not
+            // bother with it.
+            core::arch::asm!(
+                "mov {tmp}, r7",
+                "mov r7, {nr}",
+                "svc 0",
```
josephlr, line 26: Looking at the [`syscall(2)` documentation](https://man7.org/linux/man-pages/man2/syscall.2.html) it seems like there (might) be a difference between EABI / OABI targets for 32-bit arm. Also, that documentation suggests that 32-bit arm should use the `swi` instruction.

Personally, I would be fine omitting 32-bit arm support (similar to `x86` support). It's more complicated, and there isn't a 32-bit ```-unknown-linux-none``` target.

README.md
```diff
@@ -110,6 +111,14 @@ WILL NOT have any effect on its downstream users.
 
 [`.cargo/config.toml`]: https://doc.rust-lang.org/cargo/reference/config.html
 
+### Raw Linux syscall support
+
+Currently the `linux_raw` backend supports only targets with stabilized `asm!` macro,
+i.e. `arm`, `aarch64`, `loongarch64`, `riscv32`, `riscv64`, `x86`, and `x86_64`.
```
josephlr, line 117: Is there a reason s390x (the only other stable `asm!` target) is excluded? I'm totally fine not supporting it (similar to `x86` and 32-bit `arm`) if it's challenging in any way.

src/backends/linux_raw.rs
```diff
@@ -0,0 +1,127 @@
...
+            core::arch::asm!(
+                "ecall",
+                in("a7") __NR_getrandom,
+                inlateout("a0") buf => r0,
+                in("a1") buflen,
+                in("a2") flags,
+                options(nostack, preserves_flags)
+            );
+        } else if #[cfg(target_arch = "x86")] {
+            const __NR_getrandom: isize = 355;
```
josephlr, line 66: We use `isize` here while above we use u32, is there any particular reason?

src/backends/linux_raw.rs
```diff
@@ -0,0 +1,127 @@
...
+                in("ebx") buf,
+                in("ecx") buflen,
+                in("edx") flags,
+                options(nostack, preserves_flags)
+            );
+        } else if #[cfg(target_arch = "x86_64")] {
+            #[cfg(target_pointer_width = "64")]
+            const __NR_getrandom: isize = 318;
+            #[cfg(target_pointer_width = "32")]
+            const __NR_getrandom: isize = 1073742142;
```
josephlr, line 82: Could we write this as `(1 << 30) + 318`?

josephlr (commented):
Sorry for the delay in reviewing this. I think this is a fine idea, but I'm not sure if we want to bother supporting the more "difficult" targets where syscalls are more complicated than a single instruction.

src/backends/linux_raw.rs
josephlr: Looks like `swi` is the same as `svc` (https://developer.arm.com/documentation/ddi0406/cb/Application-Level-Architecture/Instruction-Details/Alphabetical-list-of-instructions/SVC--previously-SWI-?lang=en) I'm still confused about the EABI vs OABI stuff though.

src/backends/linux_raw.rs
newpavlov: Done.

src/backends/linux_raw.rs
newpavlov: It's because `inlateout("eax") __NR_getrandom => r0`  requires for both variables to have the same type. I guess we could use this instead:
```rust
in("rax") __NR_getrandom,
lateout("rax") r0,
```

README.md
newpavlov: `asm!` was stabilized for it only in Rust 1.84. I will try to add it a bit later.

src/backends/linux_raw.rs
newpavlov: IIRC the only problem with `int 0x80` is its performance, otherwise it works fine. So I think it's fine to have support for it for completeness sake despite its suboptimality.

src/backends/linux_raw.rs
newpavlov: Yeah, it would be nice to verify that raw syscalls work on Android without any surprises.

Here is a relevant `rustix` discussion: https://github.com/bytecodealliance/rustix/issues/1095 No one has mentioned any fundamental technical obstacles to using raw syscalls and it looks like some people [use them in practice](https://github.com/bytecodealliance/rustix/pull/1090#issuecomment-2248330606).

src/backends/linux_raw.rs
newpavlov: This code is based on `rustix` and it does not look like they differentiate between EABI / OABI: https://github.com/bytecodealliance/rustix/blob/main/src/backend/linux_raw/arch/mod.rs#L27-L28

After a brief search I couldn't find any relevant information about OABI support in Rust. Are you sure that Rust supports it in the first place? For example, it looks like GCC has dropped OABI support completely.

src/backends/linux_raw.rs
newpavlov: I tweaked the warning a bit. Now it notes that `linux_raw` may be slower than `libc::getrandom`-based backends in general, not just on x86.

src/backends/linux_raw.rs
newpavlov: Since users have to explicitly enable the backend, I think we can leave it as-is and potentially disable it later if it will be found problematic on Android for some reason.

PR #573, opened 2024-12-27
Remove `linux_rustix` opt-in backend

newpavlov:
As suggested [here](https://github.com/rust-random/getrandom/pull/572#pullrequestreview-2521131415) it may be worth to remove `linux_rustix` from the v0.3.0 release.

We probably will either return it in v0.3.1, or will introduce a different `linux_raw` opt-in backend (see #572).


josephlr:
LGTM, I think having fewer backends to deal with will help with resolving the lock file issues raised in #433 (however we choose to handle things).

josephlr (approved):

PR #576, opened 2025-01-02
Remove the link to PS Vita `getentropy` from the README

lukateras:
Hi ^^

Currently, PS Vita entry in the README links to the Emscripten issue tracker, likely by mistake. This pull request unlinks `getentropy` in the PS Vita entry.

Thank you for gathering platform-specific sources of entropy under one library! You're doing God's work. <3

PR #577, opened 2025-01-03
Add a link to PS Vita `getentropy` to the README

lukateras:
This is a follow-up to #576. Just found the link!


newpavlov:
Ideally, we would use a link to official docs, but in its absence I guess such link is fine. Though I think it's better to use [this link](https://github.com/vitasdk/newlib/blob/2d869fe47aaf02b8e52d04e9a2b79d5b210fd016/newlib/libc/sys/vita/getentropy.c) instead.

The CI failure is not relevant, so we can ignore it.

newpavlov (approved):

lukateras:
Aaa thank you for such a quick turnaround! ^^ (I do somewhat prefer the previous link since it shows the exact revision where it was added, similar to the 3DS link, but the reference is clear one way or the other. )

PR #579, opened 2025-01-03
Remove unnecessary error branch of ProcessPrng

silverstillisntgold:
ProcessPrng is guaranteed to never fail during runtime, so there's no need to handle that case. Also found that windows_targets declares all extern functions as pub, so wrapped the link! macro in a local module to prevent it from being accessible outside of windows.rs.

When getrandom increases the MSRV to at least 1.71, the windows-targets crate can be ditched entirely in favor of linking with raw-dylib.

### DIFF
--- a/src/backends/windows.rs
+++ b/src/backends/windows.rs
@@ -27,21 +27,20 @@ pub use crate::util::{inner_u32, inner_u64};
 
 // Binding to the Windows.Win32.Security.Cryptography.ProcessPrng API. As
 // bcryptprimitives.dll lacks an import library, we use the windows-targets
-// crate to link to it. The link! macro always declares external functions
-// as pub, so we wrap it in a private module to keep it local.
+// crate to link to it.
 //
 // TODO(MSRV 1.71): Migrate to linking as raw-dylib directly.
 // https://github.com/joboet/rust/blob/5c1c72572479afe98734d5f78fa862abe662c41a/library/std/src/sys/pal/windows/c.rs#L119
 // https://github.com/microsoft/windows-rs/blob/0.60.0/crates/libs/targets/src/lib.rs
-mod internal {
-    windows_targets::link!("bcryptprimitives.dll" "system" fn ProcessPrng(pbdata: *mut u8, cbdata: usize) -> i32);
-}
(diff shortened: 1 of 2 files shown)


src/backends/windows.rs
```diff
@@ -27,21 +27,21 @@ pub use crate::util::{inner_u32, inner_u64};
...
 
-impl Error {
-    /// Calling Windows ProcessPrng failed.
-    pub(crate) const WINDOWS_PROCESS_PRNG: Error = Self::new_internal(10);
+pub fn fill_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
+    // Since Windows 10, calls to the user-mode RNG are guaranteed
+    // to never fail during runtime (rare windows W).
+    // See the bottom of page 6 of the aforementioned Windows RNG
+    // whitepaper for more information.
+    unsafe { internal::ProcessPrng(dest.as_mut_ptr().cast::<u8>(), dest.len()) };
```
newpavlov, line 45: I think it's worth to add a debug assert:
```rust
let ret = unsafe { ... };
// Since Windows 10, calls to the user-mode RNG are guaranteed
// to never fail during runtime, i.e. they return `TRUE` which is equal to 1.
debug_assert!(ret, 1);
```

josephlr: Agreed

src/backends/windows.rs
```diff
@@ -27,21 +27,21 @@ pub use crate::util::{inner_u32, inner_u64};
...
-        TRUE => Ok(()),
-        _ => Err(Error::WINDOWS_PROCESS_PRNG),
-    }
+// crate to link to it. The link! macro always declares external functions
+// as pub, so we wrap it in a private module to keep it local.
+//
+// TODO(MSRV 1.71): Migrate to linking as raw-dylib directly.
+// https://github.com/joboet/rust/blob/5c1c72572479afe98734d5f78fa862abe662c41a/library/std/src/sys/pal/windows/c.rs#L119
+// https://github.com/microsoft/windows-rs/blob/0.60.0/crates/libs/targets/src/lib.rs
+mod internal {
```
josephlr, line 36: Personally, I don't think this `mod internal` is needed, this is part of an internal `backends` module, so `ProcessPrng` is not exported as part of the external API of this crate.

josephlr (dismissed):

newpavlov (approved):

PR #580, opened 2025-01-09
Add `opt_in` crate feature

newpavlov:
This feature resolves the lock file bloat issue at the cost of less ergonomic use of opt-in backends.

TODO: update docs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,13 +121,13 @@ jobs:
       - run: cargo build --target=${{ matrix.target }} --features=std
       - env:
           RUSTFLAGS: -Dwarnings --cfg getrandom_backend="linux_getrandom"
-        run: cargo build --target=${{ matrix.target }} --features=std
+        run: cargo build --target=${{ matrix.target }} --features=std,opt_in
       - env:
           RUSTFLAGS: -Dwarnings --cfg getrandom_test_linux_fallback
         run: cargo build --features=std
       - env:
           RUSTFLAGS: -Dwarnings --cfg getrandom_backend="rdrand"
-        run: cargo build --features=std
+        run: cargo build --features=std,opt_in
 
   web:
     name: ${{ matrix.target.description }} ${{ matrix.feature.description }} ${{ matrix.atomic.description }}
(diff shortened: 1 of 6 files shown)


dhardy:
I think it would make more sense to call this feature `wasm_backends`. In case another backend were added / made optional, we'd then add another feature for that backend.

I also don't see much point in `opt_in` affecting other backends like `rdrand` which aren't adding dependencies.

josephlr:
I think the ideal situation would be to get @dhardy's idea in https://github.com/rust-random/getrandom/issues/433#issuecomment-2563787171 to work somehow. That way an external crate (say `getrandom-js`) could have code like:
```rust
use getrandom::{register_backend, Error};

register_backend!("js", web_crypto_fill);

unsafe extern "C" fn web_crypto_fill(ptr: *mut u8, len: size) -> Option<Error> {
  // ...
}
```

And then passing `--cfg getrandom_backend="js"` would "just work". This would also remove the concept of a "custom" backend, as any custom backend could be declared and used in the same way. The "rdrand", "rndr", and "linux_getrandom" would stay as-is being "builtin" alternative backends.

dhardy:
@josephlr: I don't think `register_backend!` can work like that: IIUC the only real mechanism we have is to define a `#[no_mangle]` fn with a precise signature exactly once in the binary.

newpavlov:
>I think it would make more sense to call this feature wasm_backends.

The `opt_in` feature is not WASM-specific. It allows users to enable optional backends, including the ability to overwrite the default backends. The idea here is to have a unified approach which would be the same for any potential future opt-in backend which may bring additional dependencies (e.g. like `linux_rustix`).

@josephlr 
I am not sure how idea would work. Can you explain it in a more detail?

As I wrote [here](https://github.com/rust-random/getrandom/issues/433#issuecomment-2563644348), one alternative approach is to move WASM JS implementation into a separate crate and rely on the custom backend.

dhardy:
> The `opt_in` feature is not WASM-specific. It allows users to enable optional backends, including the ability to overwrite the default backends. The idea here is to have a unified approach which would be the same for any potential future opt-in backend which may bring additional dependencies (e.g. like `linux_rustix`) without proliferation of opt-in-specific features.

Using a separate feature flag for each backend (or group of backends for some target platform) makes more sense IMO, so e.g. `wasm_backends` and `rustix_backends` (or just `js` and `rustix`).

Unless your intention is that `--cfg getrandom_backend="X"` only be usable when `opt_in` is enabled... but I'm not sure we can lint that properly (`lints.rust.unexpected_cfgs` can't be affected by feature flags I think).

newpavlov:
Closing in favor of #574.

PR #581, opened 2025-01-10
Remove uses of #[path = "../lazy.rs"]

josephlr:
Now that the MSRV is >= 1.60, we can use `#[cfg(target_has_atomic = "ptr")]` to conditionally compile `lazy.rs`. We should probably also move other atomic/lazy code into lazy.rs, but that can wait for a followup PR.

I also slightly cleaned up the `rndr.rs` code to make things a bit more readable and to have the `std` implementation take precedence over the `asm`-based implementation. Let me know if you want me to split it off into its own PR.

### DIFF
--- a/src/backends/rndr.rs
+++ b/src/backends/rndr.rs
@@ -60,10 +60,19 @@ unsafe fn rndr_fill(dest: &mut [MaybeUninit<u8>]) -> Option<()> {
     Some(())
 }
 
+#[cfg(target_feature = "rand")]
 fn is_rndr_available() -> bool {
+    true
+}
+#[cfg(not(target_feature = "rand"))]
+fn is_rndr_available() -> bool {
+    use crate::lazy::LazyBool;
+    static RNDR_GOOD: LazyBool = LazyBool::new();
+
     cfg_if::cfg_if! {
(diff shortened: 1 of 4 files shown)


src/backends/rndr.rs
```diff
@@ -60,10 +60,19 @@ unsafe fn rndr_fill(dest: &mut [MaybeUninit<u8>]) -> Option<()> {
     Some(())
 }
 
+#[cfg(target_feature = "rand")]
 fn is_rndr_available() -> bool {
+    true
+}
+#[cfg(not(target_feature = "rand"))]
```
newpavlov, line 67: Add an empty line here for better readability.

newpavlov (commented):
Considering that `lazy` is used only in two opt-in modules, I think `#[path = "../lazy.rs"]` is fine, especially considering that it also allows us to remove `#![allow(dead_code)]`. But it could be a different story if other atomic code is moved into the lazy module.

Could you move the RNDR cleanup into a separate PR?

newpavlov:
I applied the suggested rndr tweaks in #586.

josephlr:
Closing as the remaining changes to the lazy module should wait until if we start moving other atomic code into the lazy module.

PR #584, opened 2025-01-14
Web: use safe method to copy to `MaybeUninit`

daxpedda:
This removes the last `unsafe` call in the Web implementation.
See [`Uint8Array::copy_to_uninit`](https://docs.rs/js-sys/0.3.77/js_sys/struct.Uint8Array.html#method.copy_to_uninit).


josephlr (approved):
Thanks for helping us fix this.

PR #586, opened 2025-01-24
rndr: minor tweaks

newpavlov:
The most notable change is that `is_aarch64_feature_detected!` now takes precedence over `asm!`-based detection.

Based on #581


josephlr:
LGTM, thanks!!

PR #602, opened 2025-02-05
linux_android_with_fallback: do not use dlsym on MUSL targets

### LINKED ISSUE #600: linux_android_with_fallback never uses getrandom on musl
This regressed in #508.

`dlsym` can't look up a statically linked symbol, and the result is _always_ falling back to /dev/urandom.

newpavlov:
Fixes #600

### DIFF
--- a/src/backends/linux_android_with_fallback.rs
+++ b/src/backends/linux_android_with_fallback.rs
@@ -65,7 +65,6 @@ fn init() -> NonNull<c_void> {
 }
 
 // Prevent inlining of the fallback implementation
-#[cold]
 #[inline(never)]
 fn use_file_fallback(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
     use_file::fill_inner(dest)
--- a/src/backends/linux_android_with_fallback.rs
+++ b/src/backends/linux_android_with_fallback.rs
@@ -65,6 +65,7 @@ fn init() -> NonNull<c_void> {
 }
 
 // Prevent inlining of the fallback implementation
+#[cold]
 #[inline(never)]
 fn use_file_fallback(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
     use_file::fill_inner(dest)
--- a/src/backends/linux_android_with_fallback.rs
+++ b/src/backends/linux_android_with_fallback.rs
@@ -3,8 +3,8 @@ use super::use_file;
 use crate::Error;
 use core::{
     ffi::c_void,
-    mem::{self, MaybeUninit},
-    ptr::{self, NonNull},
+    mem::{transmute, MaybeUninit},
+    ptr::NonNull,
     sync::atomic::{AtomicPtr, Ordering},
 };
 use use_file::util_libc;
@@ -17,18 +17,28 @@ type GetRandomFn = unsafe extern "C" fn(*mut c_void, libc::size_t, libc::c_uint)
 /// or not supported by kernel.
 const NOT_AVAILABLE: NonNull<c_void> = unsafe { NonNull::new_unchecked(usize::MAX as *mut c_void) };
 
-static GETRANDOM_FN: AtomicPtr<c_void> = AtomicPtr::new(ptr::null_mut());
+static GETRANDOM_FN: AtomicPtr<c_void> = AtomicPtr::new(core::ptr::null_mut());
 
 #[cold]
 #[inline(never)]
 fn init() -> NonNull<c_void> {
-    static NAME: &[u8] = b"getrandom\0";
-    let name_ptr = NAME.as_ptr().cast::<libc::c_char>();
-    let raw_ptr = unsafe { libc::dlsym(libc::RTLD_DEFAULT, name_ptr) };
+    // Use static linking to `libc::getrandom` on MUSL targets and `dlsym` everywhere else
+    #[cfg(not(target_env = "musl"))]
+    let raw_ptr = {
+        static NAME: &[u8] = b"getrandom\0";
+        let name_ptr = NAME.as_ptr().cast::<libc::c_char>();
+        unsafe { libc::dlsym(libc::RTLD_DEFAULT, name_ptr) }
+    };
+    #[cfg(target_env = "musl")]
+    let raw_ptr = {
+        let fptr: GetRandomFn = libc::getrandom;
+        unsafe { transmute::<GetRandomFn, *mut c_void>(fptr) }
+    };
+
     let res_ptr = match NonNull::new(raw_ptr) {
         Some(fptr) => {
-            let getrandom_fn = unsafe { mem::transmute::<NonNull<c_void>, GetRandomFn>(fptr) };
-            let dangling_ptr = ptr::NonNull::dangling().as_ptr();
+            let getrandom_fn = unsafe { transmute::<NonNull<c_void>, GetRandomFn>(fptr) };
+            let dangling_ptr = NonNull::dangling().as_ptr();
             // Check that `getrandom` syscall is supported by kernel
             let res = unsafe { getrandom_fn(dangling_ptr, 0, 0) };
             if cfg!(getrandom_test_linux_fallback) {
@@ -54,7 +64,7 @@ fn init() -> NonNull<c_void> {
     res_ptr
 }
 
-// prevent inlining of the fallback implementation
+// Prevent inlining of the fallback implementation
 #[inline(never)]
 fn use_file_fallback(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
     use_file::fill_inner(dest)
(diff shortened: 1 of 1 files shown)


src/backends/linux_android_with_fallback.rs
```diff
@@ -17,18 +17,28 @@ type GetRandomFn = unsafe extern "C" fn(*mut c_void, libc::size_t, libc::c_uint)
...
+    #[cfg(not(target_env = "musl"))]
+    let raw_ptr = {
+        static NAME: &[u8] = b"getrandom\0";
+        let name_ptr = NAME.as_ptr().cast::<libc::c_char>();
+        unsafe { libc::dlsym(libc::RTLD_DEFAULT, name_ptr) }
+    };
+    #[cfg(target_env = "musl")]
+    let raw_ptr = {
+        let fptr: GetRandomFn = libc::getrandom;
+        unsafe { transmute::<GetRandomFn, *mut c_void>(fptr) }
```
tamird, line 35: This cast to a pointer and back smells like an indirect call.

src/backends/linux_android_with_fallback.rs
```diff
@@ -54,7 +64,8 @@ fn init() -> NonNull<c_void> {
     res_ptr
 }
 
-// prevent inlining of the fallback implementation
+// Prevent inlining of the fallback implementation
+#[cold]
```
tamird, line 68: Is this wise? I'm not sure what it does here but the fallback is not cold if it is used.

newpavlov: Since we never inline the fallback function, this annotation just influences how compiler treats the `if fptr == NOT_AVAILABLE { .. } else { .. }` branch. Since we inline the `libc::getrandom` code, I think it makes sense to treat it as a more likely branch.

tamird: Thinking about this a bit more, not inlining the fallback also seems like a mistake. The only thing that shouldn't be inlined is the *check* for whether or not fallback should be used. Otherwise we're paying a penalty for the fallback on every call.

I don't think we should hint the compiler about likelihood of fallback vs not.

newpavlov: Maybe. I would be fine with removing `use_file_fallback` and using `use_file::fill_inner` directly in the if branch.

@josephlr WDYT?

tamird: Maybe this discussion can happen in a later PR. I suggest removing the second commit for now and merging this to fix the breakage.

tamird:
@newpavlov just force push to remove the two spurious commits - you're going to clutter the git history this way =/

newpavlov:
We always squash commits before merge, so local commit history in PR does not matter much.

tamird:
OK. If this were my project I'd ask you to back out all the unnecessary import churn and random comment changes - but it's not!

src/backends/linux_android_with_fallback.rs
```diff
@@ -17,18 +17,28 @@ type GetRandomFn = unsafe extern "C" fn(*mut c_void, libc::size_t, libc::c_uint)
 /// or not supported by kernel.
 const NOT_AVAILABLE: NonNull<c_void> = unsafe { NonNull::new_unchecked(usize::MAX as *mut c_void) };
 
-static GETRANDOM_FN: AtomicPtr<c_void> = AtomicPtr::new(ptr::null_mut());
+static GETRANDOM_FN: AtomicPtr<c_void> = AtomicPtr::new(core::ptr::null_mut());
```
tamird, line 20: could we avoid this noise in the diff? doesn't seem related to the fix being made.

src/backends/linux_android_with_fallback.rs
```diff
@@ -17,18 +17,28 @@ type GetRandomFn = unsafe extern "C" fn(*mut c_void, libc::size_t, libc::c_uint)
...
+    let raw_ptr = {
+        let fptr: GetRandomFn = libc::getrandom;
+        unsafe { transmute::<GetRandomFn, *mut c_void>(fptr) }
+    };
+
     let res_ptr = match NonNull::new(raw_ptr) {
         Some(fptr) => {
-            let getrandom_fn = unsafe { mem::transmute::<NonNull<c_void>, GetRandomFn>(fptr) };
-            let dangling_ptr = ptr::NonNull::dangling().as_ptr();
+            let getrandom_fn = unsafe { transmute::<NonNull<c_void>, GetRandomFn>(fptr) };
```
tamird, line 40: the diff on these 2 lines is also not necessary.

src/backends/linux_android_with_fallback.rs
```diff
@@ -54,7 +64,7 @@ fn init() -> NonNull<c_void> {
     res_ptr
 }
 
-// prevent inlining of the fallback implementation
+// Prevent inlining of the fallback implementation
```
tamird, line 67: not necessary

src/backends/linux_android_with_fallback.rs
```diff
@@ -78,7 +88,7 @@ pub fn fill_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
         use_file_fallback(dest)
     } else {
         // note: `transmute` is currently the only way to convert a pointer into a function reference
-        let getrandom_fn = unsafe { mem::transmute::<NonNull<c_void>, GetRandomFn>(fptr) };
+        let getrandom_fn = unsafe { transmute::<NonNull<c_void>, GetRandomFn>(fptr) };
```
tamird, line 91: not necessary

tamird (commented):
@newpavlov can we please merge this one way or the other to fix the breakage and cut 0.3.2? Thank you.

newpavlov:
I think it's fine to keep minor code tweaks in this PR.

Before releasing v0.3.2 I would like to also merge the `linux_raw` and `efi_rng` PRs.

newpavlov:
@sbwml
Please use English here, even if it's somewhat broken.

Rust officially [uses](https://doc.rust-lang.org/stable/rustc/platform-support.html) musl v1.2.3 for the `mips-unknown-linux-musl` target. Meanwhile, `getrandom` was added in musl v1.1.20. So it may be a problem with the `libc` crate, which for some reason does not expose this symbol. It may be worth to open an issue in [its repository](https://github.com/rust-lang/libc).

For now you could downgrade `getrandom` to an older version as a temporary workaround.

tamird:
Indeed, `fn getrandom` is missing on `mips-unknown-linux-musl` even though `SYS_getrandom` is defined: https://github.com/rust-lang/libc/blob/4159080fc4dcec769703d79994ab725a4e6d26b4/src/unix/linux_like/linux/musl/b32/mips/mod.rs#L727

Looks like that was somewhat intentional: https://github.com/rust-lang/libc/commit/457c02d7b5a517cd7499e98582300058ebc71ade

newpavlov:
>Looks like that was somewhat intentional

Either way, it's worth to clarify it with `libc` developers. The reason for that may not be longer applicable and, at the very least, it should be fine to add it for the MUSL target.

tamird:
The discussion in that PR (https://github.com/rust-lang/libc/pull/1399) suggests that it was failing CI at the time. The musl version was bumped in https://github.com/rust-lang/rust/pull/107129, though. I'll send a patch and see what happens.

tamird:
https://github.com/rust-lang/libc/pull/4346

eareimu:
> 好尴尬 🙃，当我为 mips musl 编译它时，遇到了一些奇奇怪怪的错误：
> 
> https://github.com/sbwml/shadowsocks-rust-mips/actions/runs/13956208010/job/39067818527#step:5:404
> 
> ```
> error[E0425]: cannot find value `getrandom` in crate `libc`
>   --> /home/runner/work/shadowsocks-rust-mips/shadowsocks-rust-mips/install/registry/src/index.crates.io-1949cf8c6b5b557f/getrandom-0.3.2/src/backends/linux_android_with_fallback.rs:34:39
>    |
> 34 |         let fptr: GetRandomFn = libc::getrandom;
>    |                                       ^^^^^^^^^ not found in `libc`
> 
> For more information about this error, try `rustc --explain E0425`.
> ```
> 
> 给它定义一个 [getrandom](https://github.com/sbwml/shadowsocks-rust-mips/blob/c4247856138f14cd303217ca1a54e4f4eab3b4d3/patches/getrandom-0.3-musl.patch) 函数它才可以编译 https://github.com/sbwml/shadowsocks-rust-mips/actions/runs/13971234091/job/39113320608#step:5:419

Same issue, the problem was temporarily solved by specifying the getramdom version in Cargo.toml:

``` toml
getrandom = "=0.3.1"
```

Experienced supply poisoning :(

josephlr:
The fix has been merged in https://github.com/rust-lang/libc/pull/4346, once it's released on `libc`'s stable 0.2 branch, the issue should be resolved.

PR #603, opened 2025-02-05
Deduplicate {linux_android,getrandom}.rs

tamird:
The former is identical to the latter. Remove the former.

r? newpavlov


newpavlov (approved):
Looks good!

Could you also remove lines 101-104 from the nopanic.yaml? Similar edits are also needed for the Clippy checks in workspace.yml.

tamird:
Done.

PR #614, opened 2025-03-03
Tweak `Error` representation

### LINKED ISSUE #568: `Error` on targets where `RawOsError` is not equal to `i32`

newpavlov:
On UEFI targets `Error` is now represented as `NonZeroUsize` and on other targets as `NonZeroI32`. In the former case OS errors have the highest bit set to 1, while in the latter case OS errors are represented as negative integers.

Additionally, this PR removes `Error::INTERNAL_START` and `Error::CUSTOM_START` associated constants since the constants are no longer relevant and the inner representation is an internal detail. This is technically a breaking change, but I highly doubt that downstream users rely on them since they can not be used for anything useful, so we can consider it as a "fix" of unintentionally exported constants.

Closes #568

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -62,6 +62,7 @@ impl Error {
 
     /// Creates a new instance of an `Error` from an UEFI error code.
     #[cfg(target_os = "uefi")]
+    #[allow(dead_code)]
     pub(super) fn from_uefi_code(code: RawOsError) -> Self {
         if code & UEFI_ERROR_FLAG != 0 {
             let code = NonZeroRawOsError::new(code).expect("The highest bit of `code` is set to 1");
--- a/src/error.rs
+++ b/src/error.rs
@@ -15,19 +15,6 @@ cfg_if::cfg_if!(
     } else {
         type RawOsError = i32;
         type NonZeroRawOsError = core::num::NonZeroI32;
-
-        /// Constant which defines whether the target uses negative or positive codes for
-        /// system errors.
-        ///
-        /// In general, only `libc`-based targets with its infamous `errno` use positive error codes.
-        ///
-        /// This constant must cover all backends which rely on `Error::from_os_error`.
-        const NEGATIVE_OS_CODES: bool = cfg!(any(
-            target_os = "solid_asp3",
-            target_os = "hermit",
-            getrandom_backend = "linux_raw",
-            all(target_arch = "wasm32", target_os = "wasi",  target_env = "p1"),
-        ));
     }
 );
 
@@ -80,29 +67,13 @@ impl Error {
     /// [1]: https://doc.rust-lang.org/std/io/struct.Error.html#method.from_raw_os_error
     #[allow(dead_code)]
     pub(super) fn from_os_error(code: RawOsError) -> Self {
-        #[cfg(target_os = "uefi")]
-        if code & UEFI_ERROR_FLAG != 0 {
-            let code = NonZeroRawOsError::new(code).expect("The highest bit is set to one");
-            return Self(code);
+        match NonZeroRawOsError::new(code) {
+            #[cfg(target_os = "uefi")]
+            Some(code) if code.get() & UEFI_ERROR_FLAG != 0 => Self(code),
+            #[cfg(not(target_os = "uefi"))]
+            Some(code) if code.get() < 0 => Self(code),
+            _ => Self::UNEXPECTED,
         }
-
-        #[cfg(not(target_os = "uefi"))]
-        {
-            if NEGATIVE_OS_CODES && (code < 0) {
-                let code = NonZeroRawOsError::new(code).expect("The code is not equal to zero");
-                return Self(code);
-            }
-
-            if !NEGATIVE_OS_CODES && (code > 0) {
-                let code = code
-                    .checked_neg()
-                    .expect("Positive numbers can be always negated");
-                let code = NonZeroRawOsError::new(code).expect("The code is not equal to zero");
-                return Self(code);
-            }
-        }
-
-        Self::UNEXPECTED
     }
 
     /// Extract the raw OS error code (if this error came from the OS)
@@ -132,12 +103,21 @@ impl Error {
 
         #[cfg(not(target_os = "uefi"))]
         {
+            // On most targets `std` expects positive error codes while retrieving error strings:
+            // - `libc`-based targets use `strerror_r` which expects positive error codes.
+            // - Hermit defers to the `hermit-abi` crate and it expects positive error codes:
+            //   https://docs.rs/hermit-abi/0.4.0/src/hermit_abi/errno.rs.html#400-532
+            // - WASIp1 uses the same conventions as `libc`:
+            //   https://github.com/rust-lang/rust/blob/1.85.0/library/std/src/sys/pal/wasi/os.rs#L57-L67
+            //
+            // The only excpetion is Solid, `std` expects negative system error codes, see:
+            // https://github.com/rust-lang/rust/blob/1.85.0/library/std/src/sys/pal/solid/error.rs#L5-L31
             if code >= 0 {
                 None
-            } else if NEGATIVE_OS_CODES {
-                Some(code)
-            } else {
+            } else if cfg!(not(target_os = "solid_asp3")) {
                 code.checked_neg()
+            } else {
+                Some(code)
             }
         }
     }
--- a/src/error.rs
+++ b/src/error.rs
@@ -92,6 +92,9 @@ impl Error {
     pub fn raw_os_error(self) -> Option<RawOsError> {
         let code = self.0.get();
 
+        // note: in this method we need to cover only backends which rely on `Error::from_os_error`,
+        // on all other backends this method always returns `None`.
+
         #[cfg(target_os = "uefi")]
         {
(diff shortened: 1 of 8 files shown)


src/error.rs
```diff
@@ -79,27 +89,38 @@ impl Error {
...
+            }
+        }
+
+        #[cfg(not(target_os = "uefi"))]
+        {
+            if code >= 0 {
+                return None;
+            }
+            #[cfg(not(target_os = "solid_asp3"))]
+            let code = code.checked_neg()?;
```
newpavlov, line 107: I am not sure whether we should do this negation on WASIp1 and Hermit.

newpavlov:
@josephlr 
Note that I plan to release v0.3.2. after merging this (with a separate release PR), so if you have any objections or suggestions about the recently merged PR it's worth to sort them out before that.

josephlr:
> @josephlr Note that I plan to release v0.3.2. after merging this (with a separate release PR), so if you have any objections or suggestions about the recently merged PR it's worth to sort them out before that.

That sounds like a good plan, I'm currently working through some of the recently merged PRs. I should hopefully be done by tomorrow. Thanks for all the fixes and reviews. Sorry I haven't had a huge amount of time for this recently.

src/error.rs
```diff
@@ -38,28 +45,33 @@ impl Error {
...
-    /// `1..Error::INTERNAL_START`. Thus, for the result `r`,
-    /// `r == Self::UNEXPECTED || r.raw_os_error().unsigned_abs() == code`.
+    /// replaced with `Error::UNEXPECTED` in unexpected cases.
     ///
     /// [1]: https://doc.rust-lang.org/std/io/struct.Error.html#method.from_raw_os_error
     #[allow(dead_code)]
-    pub(super) fn from_os_error(code: u32) -> Self {
-        match NonZeroU32::new(code) {
-            Some(code) if code.get() < Self::INTERNAL_START => Self(code),
+    pub(super) fn from_os_error(code: RawOsError) -> Self {
```
josephlr, line 69: If we want to handle all the error negating/checking in here, this could be:

On UEFI:
```rust
if code & UEFI_ERROR_FLAG  == 0 {
  Self::UNEXPECTED
} else {
  Self(NonZeroRawOsError::new(code).unwrap())
}
```

On platforms where `RawOsError` is expected to be positive:
```rust
match code.checked_neg() {
  None => Self::ERRNO_NOT_POSITIVE,
  Some(e) if e >= 0 => Self::ERRNO_NOT_POSITIVE,
  Some(e) => Self(NonZeroRawOsError::new(e).unwrap())
}
```
and on platforms where `RawOsError` is expected to be negative:
```rust
if code >= 0 {
  Self::UNEXPECTED
} else {
  Self(NonZeroRawOsError::new(code).unwrap())
}
```
then we just undo things in `raw_os_error`

src/error.rs
josephlr: I think we should figure out where they define their error constants, and use whatever convention is in place there.

josephlr (commented):
I like this change, i have some suggestions on how we might make this easier to follow

josephlr:
> > @josephlr Note that I plan to release v0.3.2. after merging this (with a separate release PR), so if you have any objections or suggestions about the recently merged PR it's worth to sort them out before that.
> 
> That sounds like a good plan, I'm currently working through some of the recently merged PRs. I should hopefully be done by tomorrow. Thanks for all the fixes and reviews. Sorry I haven't had a huge amount of time for this recently.

I didn't quite get to all of the PRs today, but a quick skim shows nothing glaringly bad. I have a few ideas for PRs that might clean things up, but nothing that should block release.

src/error.rs
```diff
@@ -38,28 +45,33 @@ impl Error {
...
-    /// Codes at or above this point can be used by users to define their own
-    /// custom errors.
+    #[deprecated]
+    #[doc(hidden)]
     pub const CUSTOM_START: u32 = (1 << 31) + (1 << 30);
 
+    /// Internal errors can be in the range of 2^16..2^17
+    const INTERNAL_START2: RawOsError = 1 << 16;
+    /// Custom errors can be in the range of 2^17..(2^17 + 2^16)
+    const CUSTOM_START2: RawOsError = 1 << 17;
```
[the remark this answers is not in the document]
newpavlov: It's not only about the value change, but also about the constants type (i.e. instead of `u32` it's now `i32` or `usize`).

src/error.rs
newpavlov: We probably should look into what is done by `std`. After all, the primary use case for the error codes is to get error strings which are included with `std`/`libc`.

src/util_libc.rs
```diff
@@ -34,14 +34,16 @@ cfg_if! {
...
 
-    // c_int-to-u32 conversion is lossless for nonnegative values if they are the same size.
-    const _: () = assert!(core::mem::size_of::<libc::c_int>() == core::mem::size_of::<u32>());
-
-    match u32::try_from(errno) {
-        Ok(code) if code != 0 => Error::from_os_error(code),
-        _ => Error::ERRNO_NOT_POSITIVE,
+    if errno > 0 {
+        let code = errno
+            .checked_neg()
```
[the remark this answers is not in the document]
newpavlov: The problem is that different targets have different conventions around error codes. We could use different `Error` representations for targets which use negative and positive error codes, but it would be... extremely annoying. Assuming that we use the same representation for both cases, we now need to decide whether we want to store OS errors as negative or positive codes.

I think that the negative approach is better since we can always negate positive codes, while with negative codes possibility of `i32::MIN` causes issues. Yes, it means that in `raw_os_error` we have to "un-negate" the code while accounting for `i32::MIN`, but since it's a relatively rarely used method, I think it's better than to introduce additional checking branches to backends.

Arguably, `libc` with its `errno` is the weird one (personally, I hate it with passion). Linux kernel and most other targets use the negative error code convention for obvious reasons.

src/error.rs
newpavlov: Hm. The problem with this approach is that we would have to manually list all targets which use positive (or negative) codes in the `error` module and do it twice.

But we kind of already do it with `#[cfg(not(target_os = "solid_asp3"))]`...

newpavlov: I tweaked the code a bit. I left the `ERRNO_NOT_POSITIVE` check in the `last_os_error` function, but otherwise it's close to your proposal.

src/error.rs
newpavlov: I added the comment for all targets which rely on `from_os_error`: https://github.com/rust-random/getrandom/blob/error_tweak/src/error.rs#L106-L114

src/error.rs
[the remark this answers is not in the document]
newpavlov: I marked the old constants private and removed `INTERNAL/CUSTOM_START2`.

src/error.rs
```diff
@@ -38,28 +43,25 @@ impl Error {
...
-    /// custom errors.
-    pub const CUSTOM_START: u32 = (1 << 31) + (1 << 30);
+    /// Internal errors can be in the range of 2^16..2^17
+    const INTERNAL_START: RawOsError = 1 << 16;
+    /// Custom errors can be in the range of 2^17..(2^17 + 2^16)
+    const CUSTOM_START: RawOsError = 1 << 17;
 
     /// Creates a new instance of an `Error` from a particular OS error code.
     ///
     /// This method is analogous to [`std::io::Error::from_raw_os_error()`][1],
```
josephlr, line 53: This documentation now doesn't match the implementation, as we now:
  - On UEFI or `solid_asp3`: we take the raw error (which is already negative in error cases)
  - On all other platforms: we take the negative of the raw error

We should either:
  - Update the docs
  - have `Error::from_os_error` work like `Error::from_raw_os_error`

as this method isn't public, we could even change the name of `Error::from_os_error` to be something like `Error::from_neg_os_error`. I'm fine considering the UEFI error codes to be "negative" as they have the high-bit set.

josephlr (approved):
Looks good to me, do we want to make the changes to `efi_rng.rs` in this PR, or in a followup PR.

src/error.rs
newpavlov: I introduced a separate method gated on UEFI targets to make it a bit more clearer.

newpavlov:
>do we want to make the changes to efi_rng.rs in this PR

I added commits which change `efi_rng` appropriately (I thought I already did it, but I guess It has slipped my mind for some reason).

PR #618, opened 2025-03-07
Add built-in support for `*-linux-none` targets

newpavlov:


src/backends.rs
```diff
@@ -38,6 +38,9 @@ cfg_if! {
                 );
             }
         }
+    } else if #[cfg(all(target_os = "linux", target_env = ""))] {
```
newpavlov, line 41: Technically, we only have the `x86_64-unknown-linux-none` target, but to future-proof ourselves I think it's fine to use a more general `cfg`.

josephlr (commented):
This looks reasonable, but should we add tests for this? I'm not sure if it's possible to build a `x86_64-unknown-linux-none` binary yet, but we could at least add a tier3 build test

josephlr:
We should also update the supported targets list in lib.rs, if you think this target deserves its own entry.

josephlr (approved):
Looks good, thanks!

I'd assume at this point there's no way to create a binary for `x86_64-unknown-linux-none` using rustix, so it's not possible to test right now?

newpavlov:
IIUC it should be possible to create an executable binary, but we would need to deal with the entry point ourselves. I think the existing `linux_raw` tests and the build-only job should be sufficient.

PR #626, opened 2025-03-08
Add cygwin support

Berrysoft:
Need https://github.com/rust-lang/rust/pull/137621 to test.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -94,6 +94,9 @@ jobs:
           x86_64-wrs-vxworks,
           x86_64-unknown-dragonfly,
           x86_64-unknown-haiku,
+          # TODO: once libstd support for cygwin is added
+          # https://github.com/rust-lang/rust/pull/137621
+          # x86_64-pc-cygwin,
         ]
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -94,8 +94,8 @@ jobs:
           x86_64-wrs-vxworks,
           x86_64-unknown-dragonfly,
           x86_64-unknown-haiku,
-          # TODO: once libstd support for cygwin is added
-          # https://github.com/rust-lang/rust/pull/137621
+          # TODO: once libc support for cygwin is released
+          # https://github.com/rust-lang/libc/pull/4308
           # x86_64-pc-cygwin,
         ]
     steps:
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -91,6 +91,7 @@ check-cfg = [
   'cfg(getrandom_test_linux_fallback)',
   'cfg(getrandom_test_linux_without_fallback)',
   'cfg(getrandom_test_netbsd_fallback)',
+  'cfg(target_os, values("cygwin"))'
```
josephlr, line 94: Please remove this line

josephlr (changes requested):
Thanks!!

Given that this target currently has libcore support, can you add it to `build.yml`'s `tier3` list with a TODO to test it once libstd support is added?

Also, can you update the documentation in `lib.rs` to add the target and a link to that target's documentation.

newpavlov:
It's also worth to add a changelog entry.

Cargo.toml
Berrysoft: Will the compiler warn or error if this line is removed?

josephlr: It will not, note that we don't have any other entries for `target_os` despite it being used in this crate. 

For why this is the case, see https://doc.rust-lang.org/rustc/check-cfg.html#well-known-names-and-values. `target_os` is a "well-known name"

Berrysoft: `target_os` is a well known name, but "cygwin" is not a well known value before rust 1.86. I'll try to remove thia line.

josephlr: I was absolutely wrong here, you were correct this line is necessary. The `unexpected_cfgs` is very annoying and confusing.

josephlr, line 94:
```suggestion
  'cfg(target_os, values("cygwin"))', # TODO(MSRV 1.86): Remove this.
```

josephlr (changes requested):

.github/workflows/build.yml
```diff
@@ -94,6 +94,9 @@ jobs:
           x86_64-wrs-vxworks,
           x86_64-unknown-dragonfly,
           x86_64-unknown-haiku,
+          # TODO: once libstd support for cygwin is added
+          # https://github.com/rust-lang/rust/pull/137621
```
josephlr, line 98:
```suggestion
          # TODO: once libc support for cygwin is released
          # https://github.com/rust-lang/libc/pull/4308
```
This test only needs libcore support, not libstd support.

josephlr (approved):
Except for a small comment, this looks good to me. Thanks!!

PR #627, opened 2025-03-09
Use "raw-dylib" for non-win7 Windows Targets

josephlr:
Now that we only use the standard Windows backend on Windows 1.78 or later, we can just use `kind = "raw-dylib"` to link to `bcryptprimitives.dll`.

This eliminates the `windows-targets` dependency and reduces the `--target=all` dependancy graph for `getrandom` to:
```
└── getrandom v0.3.1
    ├── cfg-if v1.0.0
    ├── libc v0.2.170
    ├── r-efi v5.2.0
    └── wasi v0.14.2+wasi-0.2.4
        └── wit-bindgen-rt v0.39.0
            └── bitflags v2.9.0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   is not triggered in the `linux_android_with_fallback` backend [#605]
 
 ### Changed
-- Update `windows-targets` dependency to v0.53 [#593]
+- Remove `windows-targets` dependency and use [`raw-dylib`] directly [#627]
 - Update `wasi` dependency to v0.14 [#594]
 - Add `#[inline]` attribute to the inner functions [#596]
 - Update WASI and Emscripten links in the crate-level docs [#597]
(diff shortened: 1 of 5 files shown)


newpavlov (approved):
Could you also add a changelog entry for this PR?

josephlr:
> Could you also add a changelog entry for this PR?

Done, and removed the note about #593 as the dependency is now fully removed.

glandium:
Unfortunately, there are issues with raw-dylibs before 1.84.0 on i686-pc-windows-msvc, and still to this day on i686-pc-windows-gnu.

newpavlov:
@glandium 
Does it mean that `getrandom` fails to compile on `i686-pc-windows-gnu`? Could you create a new issue for that with more information?

glandium:
The issue is actually hard to reproduce in a visible way. I haven't managed to reproduce it outside the Firefox build so far. It requires several stars being aligned, the minimum of which is:
- raw-dylib
- crate-type = staticlib
- linking with something else that also uses the symbol with lld
- delayload the library containing the symbol
and I'm not sure what else because trying to come up with a test case with only the above doesn't reproduce the issue (I can observe something weird, though, that could be related). But I can totally reproduce in the Firefox build.
This leads to ldd complaining about duplicate symbols for `__load_$symbol` on i686-pc-windows-{gnu,msvc}, where the issue was fixed in 1.84.0 for i686-pc-windows-msvc, but still exists with i686-pc-windows-gnu.

So, I'm not surprised you don't see any issues on your CI.

newpavlov:
As a potential workaround we could expose the "windows legacy" backend as an opt-in backend. Would it work for you?

We also could modify our build.rs to account for this bug, i.e. switch to the "legacy" backend on pre-1.84 compilers and on i686 targets until the bug is fixed. But personally I would like to not introduce such workarounds for somewhat exotic compiler bugs.

PR #628, opened 2025-03-09
Test Linux fallback paths without use of cfg()s

josephlr:
This PR changes our Linux fallback tests to test the "real" fallback paths rather than using various `cfg()`s within the implementation.

As these tests require being run sequentally and only work on Linux, they are configured to require manual invocation.

Also renames `tests/mod.rs` to `tests/common.rs` to reuse our tests.

Actually testing the netbsd fallback is harder, so that's best left for a future PR.


.github/workflows/tests.yml
```diff
@@ -53,6 +53,8 @@ jobs:
           targets: ${{ matrix.target }}
       - uses: Swatinem/rust-cache@v2
       - run: cargo test --target=${{ matrix.target }} --features=std
+      - run: cargo test --test linux_no_fallback -- --test-threads 1
+      - run: cargo test --test linux_force_fallback -- --test-threads 1
```
newpavlov, line 57: Are you sure that the order of tests is guaranteed?

newpavlov:
Overall, it's an interesting approach, but I am not sure if it's _simpler_ than the `cfg`-based one, plus it relies on several non-trivial things like overriding `getrandom` symbol linking, execution order of tests, and  reading `/proc/self/fd`.

newpavlov:
@josephlr 
Do you plan to work on this or should we close it?

josephlr:
> @josephlr Do you plan to work on this or should we close it?

I plan to work on this, but it's not the type of thing which should block releases.

Sorry for the delay in response. I've recently switched roles inside Google, so things have been a bit chaotic. On the plus side I should now have more time to work on Rust stuff.

newpavlov:
This PR is quite stale and has a lot of conflicts with the master branch, so if we are to do it, I think it's better to start from scratch.

josephlr:
> This PR is quite stale and has a lot of conflicts with the master branch, so if we are to do it, I think it's better to start from scratch.

Agreed, thanks for cleaning this up!

PR #639, opened 2025-03-26
Add crate version to docs.rs links used in `compile_error!`s

### LINKED ISSUE #638: small change to docs to lessen feature flag change confusion
In 0.2.x versions of getrandom, the feature flag "js" was used to setup wasm32-unknown-unkown.  This changed to "wasm_js" in 0.3.x versions.  I recently ran into a problem where I had 3 different sub-dependencies in a project on getrandom across these 2 versions.  The confusion was because the web URL given in the error always shows the latest docs, I wasted some time not realizing the feature flag had changed.  

I would propose to have a small change to the docs to list the previous feature flag(s) in the current docs.  If it is only js -> wasm_js that is affected, I would be happy to pull a PR, otherwise, it might be better to have someone more knowledgeable complete this.

Shows a demo of how to handle this:
https://github.com/holderl5/rust-sub-dependency-feature/tree/main

newpavlov:
Closes #638

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.3.3] - UNRELEASED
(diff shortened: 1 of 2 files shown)


josephlr (approved):
This (and the other doc changes) looks good

PR #640, opened 2025-03-26
ci: enable backtrace in tests

newpavlov:
This would give use more information from panicking tests, e.g. https://github.com/rust-lang/rust/issues/138984.


josephlr:
LGTM!

PR #666, opened 2025-05-17
Add no_op backend

dsherret:
There are times when I go to use a crate in Wasm and some dependency of a dependency uses this crate for randomness, but I'm not using the functionality that needs the randomness. Providing a custom backend doesn't work because then it creates a wasm export and I want my wasm interface clean. For scenarios like this, it's useful to be able to just have this crate not provide any randomness at all.

### DIFF
--- a/src/backends/no_op.rs
+++ b/src/backends/no_op.rs
@@ -1,0 +1,12 @@
+//! Implementation that's not random and does a no-op.
+use crate::Error;
+use core::mem::MaybeUninit;
+
+pub use crate::util::{inner_u32, inner_u64};
+
+pub fn fill_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
+    for byte in dest {
+        byte.write(0);
+    }
+    Ok(())
+}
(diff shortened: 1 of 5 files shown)


src/backends/no_op.rs
```diff
@@ -0,0 +1,12 @@
...
+use core::mem::MaybeUninit;
+
+pub use crate::util::{inner_u32, inner_u64};
+
+pub fn fill_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
+    for byte in dest {
+        byte.write(0);
+    }
+    Ok(())
+}
```
dhardy, line 12: This should fail, reliably.

dhardy (changes requested):
The given impl isn't a no-op, and pretending that we randomized `dest` is not okay for any other dependency needing a securely-generated random key. It is a security hazard.

dsherret:
> The given impl isn't a no-op

Do you have a suggestion for a better name?

> and pretending that we randomized dest is not okay for any other dependency needing a securely-generated random key. It is a security hazard.

There are cases where it's completely fine.

dsherret:
Maybe a better solution would be to have an implementation that does a runtime panic or error?

dsherret:
I opened https://github.com/rust-random/getrandom/pull/667 as an alternative.

PR #667, opened 2025-05-17
`unsupported` opt-in backend

dsherret:
An alternative to https://github.com/rust-random/getrandom/pull/666/ that errors at runtime.

This is useful when needing the crate to compile when targeting wasm32, but this code is never actually hit.


.github/workflows/build.yml
```diff
@@ -273,3 +273,17 @@ jobs:
       - env:
           RUSTFLAGS: -Dwarnings --cfg getrandom_backend="custom"
         run: cargo build --target riscv32i-unknown-none-elf
+
+
```
newpavlov, line 277: Remove one empty line.

README.md
```diff
@@ -87,6 +87,7 @@ of randomness based on their specific needs:
 | `wasm_js`         | Web Browser, Node.js | `wasm32‑unknown‑unknown`, `wasm32v1-none` | [`Crypto.getRandomValues`]. Requires feature `wasm_js` ([see below](#webassembly-support)).
 | `efi_rng`         | UEFI                 | `*-unknown‑uefi`         | [`EFI_RNG_PROTOCOL`] with `EFI_RNG_ALGORITHM_RAW` (requires `std` and Nigthly compiler)
 | `custom`          | All targets          | `*`                      | User-provided custom implementation (see [custom backend])
+| `runtime_error`   | All targets          | `*`                      | Errors when providing randomness. Useful when only needing this crate to compile in wasm32, but it's not actually used.
```
newpavlov, line 90: I would write the description a bit shorter:
>Always returns `Err(Error::UNSUPPORTED)` (see [unsupported backend](link))

You can remove the `Error::UNSUPPORTED` example from the custom backend section and add a new section for this backend with description of when it should be used.

newpavlov (commented):
Looks mostly good, but I am not sure about the backend name. How about `unsupported` since it returns `Error::UNSUPPORTED`?

README.md
dsherret: I updated it. There was actually an example in the "custom backend" section about creating a custom backend that returns Unsupported, but it's not an ideal recommendation because it creates a Wasm module import in the case of compiling for Wasm. I moved that paragraph into the new "Unsupported backend" section

dsherret:
> I am not sure about the backend name. How about unsupported since it returns Error::UNSUPPORTED?

I wasn't sure about it either. Yeah, I like that better. Updated.

README.md
```diff
@@ -204,20 +204,14 @@ unsafe extern "Rust" fn __getrandom_v03_custom(
...
-#[no_mangle]
-unsafe extern "Rust" fn __getrandom_v03_custom(
-    dest: *mut u8,
-    len: usize,
-) -> Result<(), Error> {
-    Err(Error::UNSUPPORTED)
-}
-```
+In some rare scenarios you might be compiling this crate in a constrained
+environment (ex. `wasm32-unknown-unknown`), but this crate's functionality
```
newpavlov, line 210: In the rest of this document we use "e.g.", not "ex.".

newpavlov (approved):
I will leave this PR open for some time for @josephlr to take a look on it. Plus we need to fix the failing CI jobs (it looks like a Nightly toolchain issue).

newpavlov:
Also, please add a changelog entry. You can add the following section:
```
## [0.3.4] - UNRELEASED

### Added
- Your entry

// At the end of the file:
[0.3.4]: https://github.com/rust-random/getrandom/compare/v0.3.3...HEAD
```

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.3.4] - UNRELEASED
+
+### Added
+- `unsupported` opt-in backend [#667]
+
+[#667]: https://github.com/rust-random/getrandom/pull/667
+
+[0.3.4]: https://github.com/rust-random/getrandom/compare/v0.3.3...HEAD
```
newpavlov: This link should be moved to the end of the file.

newpavlov:
Thank you!

We probably will wait for some time before releasing v0.3.4 to accumulate other potential changes. Meanwhile, you could temporarily patch `getrandom` for your project.

PR #668, opened 2025-05-19
Fix Web CI

josephlr:


newpavlov (approved):

newpavlov:
As for the PowerPC failure, we probably can temporarily disable the job (after reporting the issue) since it looks like a toolchain problem in the latest stable release.

josephlr:
> As for the PowerPC failure, we probably can temporarily disable the job (after reporting the issue) since it looks like a toolchain problem in the latest stable release.

I'm going to replace it with a RISC-V Target for now, so we are at least testing something that isn't arm/x86.

PR #699, opened 2025-07-23
Use getrandom syscall on riscv32/riscv64 linux

kraj:
Minimum kernel needed on RISCV is fairly new (4.15+) so we are sure to have getrandom syscall, on glibc there is mimimal ABI kernel to denote it but musl does not have any other way to indicate it, so add it as a condition here to choose getrandom backend for rv32/rv64 on linux when using musl.


josephlr (approved):
Looks reasonable to me!

I think we should probably go through and clean up that large case statement, to more systematically consider particular libc being used, but that shouldn't block this PR.

PR #707, opened 2025-08-19
Add optee backend

DemesneGH:
This adds support for the {arm,aarch64}-unknown-optee targets, enabling `getrandom` to work in OP-TEE environments.

For more information about OP-TEE, see:
https://optee.readthedocs.io/en/latest/


DemesneGH:
The `{arm,aarch64}-unknown-optee` target is currently a custom target and not part of upstream Rust. Because of this, #[cfg(target_os = "optee")] is rejected during compilation.
Would it be possible to merge a backend for a custom target like this? Could anyone provide guidance? Thanks!

newpavlov:
You should use a [custom backend](https://docs.rs/getrandom/latest/getrandom/#custom-backend) for such targets.

PR #714, opened 2025-09-08
ci: disable NetBSD job

newpavlov:


josephlr (commented):
@newpavlov, if the NetBSD job is broken, does that mean we can also remove the following lines from `Cargo.toml`?:
https://github.com/rust-random/getrandom/blob/3983e0fbec37516bd5e1b32eb4bdacb612f36f8b/Cargo.toml#L91-L100

newpavlov:
It was a temporary issue with their CDN. It seems to be fixed now, so we can re-enable the job.

PR #715, opened 2025-09-08
Update Cargo.lock

newpavlov:


josephlr:
LGTM (assuming that this just ran `cargo update`, might be good to add a description if anything else was done).

PR #716, opened 2025-09-14
feat: Add support for Ariel OS

SimonIT:
https://github.com/ariel-os/ariel-os


newpavlov:
We do not have built-in support for targets which do not have a Rust target triplet assigned to them (`wasm-js` is the only exception because of its huge relative popularity). You should use the custom backend in such cases. The support may be even included as part of a crate developed by the Ariel team.

PR #720, opened 2025-09-15
ci: tweak Clippy job for WASIp2

newpavlov:


newpavlov:
The issue should be fixed on the `wasi` side: https://github.com/bytecodealliance/wasi-rs/pull/121

Issue #723, opened 2025-09-19
Does not build out of the box on `x86_64-pc-windows-gnu` (dlltool.exe: program not found)


ColinFinck:
Attempting to build latest `getrandom` on a pristine Windows installation with the `x86_64-pc-windows-gnu` toolchain results in an `error calling dlltool 'dlltool.exe': program not found` failure. This is a known Rust bug of the `-gnu` targets when using `raw-dylib` (see https://github.com/rust-lang/rust/issues/103939): The `-gnu` targets add a call to `dlltool` but don't ship `dlltool`.

Using the new `-gnullvm` target is no alternative either, because it currently lacks even more components: https://github.com/rust-lang/rust/issues/146634.
I cannot use the `-msvc` targets, because I need a freely redistributable toolchain. I also cannot install the missing `dlltool.exe`. To sum it up: `getrandom` needs to work for me on a pristine unmodified `x86_64-pc-windows-gnu` toolchain.

This was no problem until the merge of #627.

I see several options how to solve this problem:

1. Reverting #627 and using `windows-targets` again.
2. Retrieving the address of `ProcessPrng` at runtime using `LoadLibraryW` and `GetProcAddress` (can be put in a `OnceLock` to not do that on every call).
3. Adding a feature to use `RtlGenRandom` instead.

As a quick fix, I'll go with option 3 for now. What do you think could be a long-term solution?

CC @josephlr @newpavlov

ColinFinck:
This is how I fixed it for me now: https://github.com/osdyne/getrandom/commit/23f5f0df1beec1e3b4790e6abd9eb62f02f58765
Currently, I'm only using `getrandom` as a dependency of `tempfile`, so the quality of the generated random numbers is negligible. Your mileage may vary.

If this is a fix you would accept, I could open a PR.

newpavlov:
Shouldn't it be fixed on the language level? The same linking approach based on `kind = "raw-dylib"` is also [used](https://github.com/rust-lang/rust/blob/2f4dfc753fd86c672aa4145940db075a8a149f17/library/std/src/sys/pal/windows/c.rs#L101-L110) by `std`, so just changing `getrandom` would not help you.

newpavlov:
You can enable the legacy backend using the `getrandom_windows_legacy` configuration flag (we [use](https://github.com/rust-random/getrandom/blob/master/build.rs#L51C43-L51C67) it in our build.rs).

ColinFinck:
[...]

PR #724, opened 2025-09-19
Add `windows-legacy` opt-in backend

newpavlov:
Some users may want to use the legacy backend (e.g. see #723), so I think it's worth to introduce a documented configuration flag for it to replace the undocumented `getrandom_windows_legacy` flag.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,9 +10,9 @@ checksum = "46c5e41b57b8bba42a04676d81cb89e9ee8e859a1a66f80a5a72e1cb76b34d43"
 
 [[package]]
 name = "cc"
-version = "1.2.37"
+version = "1.2.38"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


build.rs
```diff
@@ -48,7 +48,7 @@ fn main() {
 
         match rustc_minor_version() {
             Some(minor_ver) if minor_ver < WIN7_INTRODUCED_MINOR_VER => {
-                println!("cargo:rustc-cfg=getrandom_windows_legacy");
+                println!("cargo:rustc-cfg=getrandom_backend=\"windows_legacy\"");
```
josephlr, line 51: One thing that concerned me about this line is if it would override users who used `--getrandom_backend="whatever"` on Windows when using an older compiler.

Turns out, I can be an issue, but depends on the value of `"whatever"`.  Opened #728 to fix.

josephlr:
Other than #728, this looks good. Thanks for taking care of this. It makes things a lot clearer to the user.

Issue #633, opened 2025-03-21
__getrandom_v03_custom Not Preventing wasm-bindgen Import on wasm32-unknown-unknown with Internet Computer


e274426380:
**Description**
Hi, I'm encountering an issue where defining __getrandom_v03_custom as per the documentation (https://docs.rs/getrandom/0.3.2/getrandom/#opt-in-backends) does not prevent wasm-bindgen from being imported when targeting wasm32-unknown-unknown for deployment on the Internet Computer (IC). Despite disabling default features and enabling the custom backend via --cfg getrandom_backend="custom", the resulting WASM file still contains __wbindgen_describe imports, causing deployment failures on IC.

**Expected Behavior**
According to the documentation, defining __getrandom_v03_custom and enabling the custom backend should override the default wasm_js backend, avoiding the dependency on wasm-bindgen and its associated imports (e.g., __wbindgen_describe).

**Actual Behavior**
Even with the custom function defined and the configuration applied, wasm-bindgen is still linked into the final WASM binary, as verified by wasm-objdump -x. This leads to deployment errors on the Internet Computer, which does not support wasm-bindgen imports.

According to my query, the dependency chain is: burn -> rand -> getrandom

```
[package]
name = "burn_price_backend"
version = "0.1.0"
edition = "2021"

[lib]
crate-type = ["cdylib"]

[dependencies]
candid = "0.10"
ic-cdk = "0.16"
ic-cdk-macros = "0.13"
bincode = "1.3.3"
burn = { git = "https://github.com/burn-rs/burn", default-features = false, features = ["std", "ndarray", "autodiff"] }
serde = { version = "1.0", features = ["derive"] }
getrandom = { version = "0.3.2", default-features = false }
```

.cargo/config.toml

```
[build]
rustflags = ["--cfg", "getrandom_backend=\"custom\""]
target = "wasm32-unknown-unknown"
```

This is my code repository:
[https://github.com/e274426380/burn_price](https://github.com/e274426380/burn_price )

e274426380:
> We will add a stronger warning for `wasm_js` (see [#633](https://github.com/rust-random/getrandom/issues/633)). Since it's a downstream issue, we can not do much here, so I will close this issue.

But as far as I know, the getrandom project using version 0.2.9 works fine on Internet computers.
And use this configuration: 
`features = [“custom”] `

newpavlov:
[...]

Issue #176, opened 2020-12-05
A way to avoid wasm-bindgen dependency?


RazrFalcon:
Currently, a vendored `getrandom` archive includes 40! dependencies and 20MB on disk. This is not a problem when people simply using it as a crate, because all of those dependencies will be omitted, but in case of a vendored archive use case it ends up as pretty serious bloat.

Is there a way to reduces the number of indirect dependencies?

newpavlov:
Optional dependencies by default do not get vendored-in. In a new crate dependent only on `getrandom` I get 4 crates in the `vendor` folder: `cfg-if`, `getrandom`, `libc`, and `wasi`.

RazrFalcon:
Ok. My mistake. Then we left with `wasm-bindgen` which has like 10 dependencies.

RazrFalcon:
Wait, so is there a way to remove `wasm-bindgen`?

newpavlov:
Yes, if the crate feature which does not enable `wasm-bindgen` is not enabled, then it and all its dependencies should not get vendored (I assume you use `cargo vendor`). Probably one of your dependencies unconditionally enables this feature (e.g. via `rand` which in v0.7 forwards the `wasm-bindgen` feature), so you need to find it and ask for a fix.

RazrFalcon:
To be more precise, I'm building the `uuid` crate with `wasm-bindgen` feature, to make it work in web. Which pulls `wasm-bindgen`. Maybe there is a way to avoid it altogether?

newpavlov:
If you don't need UUIDv4 (random UUIDS), simply do not enable `wasm-bindgen` and `v4` features. If you need randomness from browser, then `wasm-bindgen` is a required dependency. Without it you will not be able to talk with browser environment while using the `wasm32-unknown-unknown` target.

RazrFalcon:
>then wasm-bindgen is a required dependency.

I see. This is what I wanted to hear. Thanks.

PR #725, opened 2025-09-19
Add `custom-fallback` crate feature

newpavlov:
cc @daxpedda

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,10 @@ exclude = [".*"]
 # use std to retrieve OS error descriptions
 std = []
 
+# Use the custom backend on unsupported targets
+# WARNING: this feature SHOULD be enabled only by fallback implementation crates!
+custom-fallback = []
+
 # Optional backend: wasm_js
 # This flag enables the backend but does not select it. To use the backend, use
 # this flag *and* set getrandom_backend=wasm_js (see README).
--- a/src/backends.rs
+++ b/src/backends.rs
@@ -182,6 +182,9 @@ cfg_if! {
     } else if #[cfg(all(target_arch = "x86_64", target_env = "sgx"))] {
         mod rdrand;
         pub use rdrand::*;
+    } else if #[cfg(feature = "custom-fallback")] {
+        mod custom;
+        pub use custom::*;
     } else if #[cfg(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))] {
         compile_error!(concat!(
             "The wasm32-unknown-unknown targets are not supported by default; \


dhardy:
This has the same [issue as #672](https://github.com/rust-random/getrandom/pull/672#issuecomment-2948708077): any crate which adds a custom backend for wasm-js (as a convenience to wasm-web users) precludes any user from providing their own custom backend.

The only way this is "superior" to the "js" feature of v0.2 is that it moves wasm-js code outside of `getrandom` proper.

dhardy:
I feel like any further points here would just be re-hashing prior discussions. My preference is still #675 (possibly with a long and very specific feature name and allow override via `getrandom_backend` cfg).

newpavlov:
>any crate which adds a custom backend for wasm-js (as a convenience to wasm-web users) precludes any user from providing their own custom backend.

As I wrote [here](https://github.com/wasm-bindgen/wasm-bindgen/issues/4667#issuecomment-3312926970), I believe it's much less likely to happen in practice `js-sys` implementing the custom fallback backend. Adding a new direct dependency is usually perceived as a heavier action than enabling a feature in an existing transitive dependency (despite it potentially pulling new transitive dependencies).

dhardy:
I don't think that matters though:

1. We can allow non-web WASM users to override the backend when a `wasm-js` feature flag is enabled anyway, using the `getrandom_backend` cfg
2. It isn't really our problem to solve in any case when crate `foo` depends on `getrandom` with feature `wasm-js` in cases where it shouldn't. Non-web users can file a bug report against the crate.

newpavlov:
>We can allow non-web WASM users to override the backend when a wasm-js feature flag is enabled anyway, using the getrandom_backend cfg

IIRC it does not work in practice. Pulling `js-sys`/`wasm-bindgen` automatically screws up non-Web WASM developers and we can not disable/enable a dependency using `cfg`s.

It's much easier to see which dependency pulls `js-sys`/`wasm-bindgen` in your dependency tree, than to search for a crate which enables the `wasm_js` feature in `getrandom`.

>It isn't really our problem to solve in any case when crate foo depends on getrandom with feature wasm-js in cases where it shouldn't. Non-web users can file a bug report against the crate.

Theoretically, you are right. But as our history with v0.2 has shown, in practice it does not work well. My aim with this PR (and the `wasm-bindgen` proposal) is to discourage such behavior as much as possible while minimally hurting ergonomics.

dhardy:
> It's also a matter of consistency. We reject PRs which add built-in support for environments without proper Rust targets associated with it (e.g. https://github.com/rust-random/getrandom/pull/716) and wasm_js is the only exception here.

Regarding consistency, firstly this indicates that there possibly should be a wasm-web target (off topic), and secondly I do not see the consistency as mattering much. We've had *far* more requests for WASM-web support than Ariel OS.

Allowing targets like Ariel to more easily support `getrandom` would be an argument for this PR, but when it comes to WASM-web I disagree because I think mis-use (wrong dependency on a getrandom-for-WASM-JS custom backend) is rather likely.

We could do both #675 and this PR if you like (in which case the custom backend should probably take priority where both are enabled). This would leave `getrandom_backend` mostly as a tool of last resort; acceptable?

dhardy:
> we can not disable/enable a dependency using cfgs

Unless the dependency itself is causing a problem this doesn't matter? (Obviously there could be Cargo.lock bloat still.)

> It's much easier to see which dependency pulls js-sys/wasm-bindgen in your dependency tree

True, though there shouldn't be many dependencies on `getrandom` anyway.

I just think the possibility of defaulting to WASM-JS but still allowing custom backends is the best option. Is there a reason this wouldn't work?

newpavlov:
>I just think the possibility of defaulting to WASM-JS but still allowing custom backends is the best option. Is there a reason this wouldn't work?

As I wrote above, IIRC simply having `wasm-bindgen` in your dependency tree breaks non-Web WASM builds even if we do not use it. We have the same problem with the current `wasm_js` feature, but since users are unlikely to enable it for "convenience" (because we require `cfg` on top of it), it's much less impactful in practice.

This is why I think it's advantageous to not have `js-sys` as a `getrandom` dependency in the first place.

>We could do both https://github.com/rust-random/getrandom/issues/675 and this PR if you like

I don't think it makes much sense to do both. It looks like we are at an impasse. So I guess @josephlr could act as a tie breaker here.

dhardy:
> As I wrote above, IIRC simply having `wasm-bindgen` in your dependency tree breaks non-Web WASM builds even if we do not use it.

I found #176 (`Cargo.lock` bloat: a minor issue) and eventually #633 (maybe a more serious issue, though could be further investigated: `wasm-bindgen` should not need to be linked if a `cfg` is used to select another backend).

I'm happy enough with the idea of placing the JS impl in `js-sys` or a third-party crate, though I think it might be worth using a different `extern "Rust"` fn name and feature name to still allow a custom backend to override the JS one — assuming that the override can actually work. Testing required.

Pauan:
> @newpavlov As I wrote above, IIRC simply having wasm-bindgen in your dependency tree breaks non-Web WASM builds even if we do not use it.

It shouldn't, if it does that's a bug.

Especially if you [mark the `wasm-bindgen` dependency as `optional`](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies):

```toml
[features]

[dependencies]
wasm-bindgen = { version = "0.2.103", optional = true }
js-sys = { version = "0.3.80", optional = true }
```

Now the `wasm-bindgen` and `js-sys` dependencies will only be included if you enable the `wasm_js` feature.

newpavlov:
@Pauan 
We discuss the case when `js-sys` *is* a dependency (i.e. when we enable the `wasm_js` feature), but is not used (e.g. because it was overwritten using `cfg`).

dhardy:
> Testing required.

About that: does anyone know of non-web WASM targets we can actually test against? I think most uses are probably some type of block-chain compute platform, but we need something we can test in our CI runners.

It's not really hard to guess why people are happy to assume that WASM-means-web.

I don't care enough to argue this further; @newpavlov's preference of moving the wasm-js impl to another crate via a custom backend is good enough.

dhardy (approved):

Cargo.toml
```diff
@@ -16,6 +16,10 @@ exclude = [".*"]
 # use std to retrieve OS error descriptions
 std = []
 
+# Use the custom backend on unsupported targets
```
dhardy, line 19: Should say:
```toml
# Use the custom backend when getrandom_backend cfg is not set
```

dhardy:
**Concern: security.**

Currently `getrandom` either "just works" or needs some form of configuration. Custom backends must be explicitly configured by whoever builds binaries.

This change would allow `getrandom` to "just work" on WASM via inclusion of an external dependency. Intermediate crates would have the ability to supply some random source.

Implication: verifying security of `getrandom` output on WASM-web is harder since the backend might be imported anywhere in the tree. (Though since at most one backend can be provided, this is not such a big attack vector.)

Ultimately since Rust is not strictly memory safe and therefore every crate in the build must be verified, this is not strictly worse than the alternative, though I would still prefer in-crate implementation (#675).

Pauan:
> @dhardy About that: does anyone know of non-web WASM targets we can actually test against? I think most uses are probably some type of block-chain compute platform, but we need something we can test in our CI runners.

The best options right now are `wasm32-wasip1` and `wasm32-wasip2`

dhardy:
@Pauan those targets already have support. This issue is about `wasm32-unknown-unknown`.

Pauan:
@dhardy Non-JS usages have generally moved away from `wasm32-unknown-unknown` and instead switched to WASI (e.g. [Cloudflare](https://blog.cloudflare.com/announcing-wasi-on-workers/)).

EWasm potentially would have used `wasm32-unknown-unknown`, but it has stalled.

[CosmWasm](https://book.cosmwasm.com/basics/building-contract.html) does use `wasm32-unknown-unknown`.

[Casper](https://docs.casper.network/developers/writing-onchain-code/getting-started) uses `wasm32-unknown-unknown`.

It's actually quite difficult to find non-JS uses of `wasm32-unknown-unknown`.

dhardy:
@Pauan that's helpful thanks. My opinion is therefore that we should care less about non-js `wasm32-unknown-unknown` and adopt #675, but @newpavlov disagrees. @josephlr can we ask you to weigh in on this, please?

newpavlov:
It also makes sense for experimental targets (e.g. see the recent PR for Ariel OS). Such targets could use this feature in their respective "sys" crates.

Technically, I believe we should view Web WASM as such experimental target, despite its popularity.

Pauan:
> @newpavlov Technically, I believe we should view Web WASM as such experimental target, despite its popularity.

In a very strict technical sense, yes... but it is a very stable target, things are not breaking on a weekly basis (like they used to back when it was bleeding edge). In a practical sense it is stable, well supported, and widely used.

The lack of a dedicated `wasm32-web` target is an unfortunate historical quirk,  but it has *mostly* worked out okay thanks to the existence of WASI.

PR #726, opened 2025-09-22
Add assume_wasm_js feature

dhardy:
Essentially this is #675. Partial alternative to #725.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,10 @@ std = []
 # i.e. avoid unconditionally enabling it in library crates.
 wasm_js = ["dep:wasm-bindgen", "dep:js-sys"]
 
+# This flag enables the wasm_js backend and uses it by default on wasm32 where
+# the target_os is unknown. The getrandom_backend cfg may override this.
+assume_wasm_js = ["wasm_js"]
+
 [dependencies]
 cfg-if = "1"
 
--- a/README.md
+++ b/README.md
@@ -132,20 +132,26 @@ which JavaScript interface should be used (or if JavaScript is available at all)
 
 To enable `getrandom`'s functionality on `wasm32-unknown-unknown` using the Web
 Crypto methods [described above][opt-in] via [`wasm-bindgen`], do
-*both* of the following:
+one of the following:
 
--   Use the `wasm_js` feature flag, i.e.
-    `getrandom = { version = "0.3", features = ["wasm_js"] }`.
-    On its own, this only makes the backend available. (As a side effect this
-    will make your `Cargo.lock` significantly larger if you are not already
-    using [`wasm-bindgen`], but otherwise enabling this feature is harmless.)
--   Set `RUSTFLAGS='--cfg getrandom_backend="wasm_js"'` ([see above][opt-in]).
+-   Use the `wasm_js` feature flag to enable the backend (which has the side
+    effect of making `Cargo.lock` significantly larger if you are not already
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -23,6 +23,10 @@ std = []
 # i.e. avoid unconditionally enabling it in library crates.
 wasm_js = ["dep:wasm-bindgen", "dep:js-sys"]
 
+# This flag enables the wasm_js backend and uses it by default on wasm32 where
+# the target_os is unknown. The getrandom_backend cfg may override this.
+assume_wasm_js = ["wasm_js"]
```
newpavlov, line 28: As I wrote in #725, while you technically can override the backend (as any other built-in backend), `wasm-bindgen`/`js-sys` will stay in the dependency tree, thus breaking non-Web WASM builds.

So I don't think it's worth to keep the last sentence.

README.md
```diff
@@ -132,20 +132,26 @@ which JavaScript interface should be used (or if JavaScript is available at all)
...
-    `getrandom = { version = "0.3", features = ["wasm_js"] }`.
-    On its own, this only makes the backend available. (As a side effect this
-    will make your `Cargo.lock` significantly larger if you are not already
-    using [`wasm-bindgen`], but otherwise enabling this feature is harmless.)
--   Set `RUSTFLAGS='--cfg getrandom_backend="wasm_js"'` ([see above][opt-in]).
+one of the following:
+
+-   Use the `wasm_js` feature flag to enable the backend (which has the side
+    effect of making `Cargo.lock` significantly larger if you are not already
+    using [`wasm-bindgen`], but is otherwise harmless), **and**
```
newpavlov: No, it's not harmless.

Cargo.toml
dhardy: To which [I tried to find evidence](https://github.com/rust-random/getrandom/pull/725#issuecomment-3316200364), but I'm not convinced by it. Do you have more evidence?

newpavlov: You literally linked #633. It may be possible to fix it somehow, but in practice today it does break non-Web builds. (well, technically I should use past tense since I did not check it with the latest Rust/wasm-bindgen/js-sys versions, but I assume nothing has changed since then)

dhardy: #633 is not easy to reproduce (the reporter has removed his code) and is not using recent tooling. Is it an issue today? Because one unverifiable report is not sufficient evidence of a problem in my book.

Feature `wasm_js` enables `wasm-bindgen` but does not by itself import anything (nor does `assume_wasm_js` if another `getrandom_backend` is specified), so whether there is an issue in practice depends on the build.

README.md
dhardy: Revised wording. "Harmless" was from the old version.

Cargo.toml
newpavlov: Ok, it seems I was wrong about build breakage. Pulling and compiling `wasm-bindgen` is still not a great look, but it's not as a serious issue as I thought.

Nevertheless, I still prefer to move the Web WASM backend to `js-sys`.

dhardy: > Nevertheless, I still prefer to move the Web WASM backend to `js-sys`.

How would the integration work?

- `getrandom` depending on `js-sys` is fine but doesn't change much vs this solution
- `js-sys` providing a custom backend requires #725 or build-time configuration

newpavlov: Yes, I meant the custom fallback approach.

PR #730, opened 2025-09-29
Implement #675: let wasm_js enable the backend by default

dhardy:
Implements #675. Closes #726.

Should we remove support for `getrandom_backend="wasm_js"` since it is redundant? That would be a breaking change, and I'm not 100% sure it would never be wanted (though the `wasm_js` module currently throws an error of `target_os` is not either "unknown" or "none").

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,8 +17,8 @@ exclude = [".*"]
 std = []
 
 # Optional backend: wasm_js
-# This flag enables the backend but does not select it. To use the backend, use
-# this flag *and* set getrandom_backend=wasm_js (see README).
+# This flag enables the wasm_js backend and uses it by default on wasm32 where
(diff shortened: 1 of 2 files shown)


dhardy:
Pinging @newpavlov @josephlr for feedback please.

newpavlov:
I will leave it to @josephlr to accept, merge, and release this PR. Implementation-wise, I don't see any issues.

josephlr:
> Should we remove support for `getrandom_backend="wasm_js"` since it is redundant? That would be a breaking change, and I'm not 100% sure it would never be wanted (though the `wasm_js` module currently throws an error of `target_os` is not either "unknown" or "none").

I think we should keep it for now, both to avoid a breaking change, and to give people a better error message when they try to select the backend without enabling the required feature.

PR #732, opened 2025-09-29
Update all nightly toolchains to latest nightly

josephlr:
Update all the nightly toolchains to `nightly-2025-09-28`. I also included the following fixes:
  - Don't use the `doc_auto_cfg` feature
    - It was removed in a recent nightly, see https://github.com/rust-lang/rust/pull/138907
    - If we just use `doc_cfg`, everything gets generated correctly:
      <details>
      <summary>Example only using cfg(doc_cfg)</summary>
      Screenshot 2025-09-29 at 3 40 25 PM
      </details>
  - Stop running doctests with `-Zsanitizer=memory`
    - Passing the correct flags to doctests is tricky with `-Zbuild-std` and the doc tests don't really gain us much.
    - See: https://github.com/rust-lang/rust/issues/134172#issuecomment-3349433658
    - I use `--all-targets` to skip the doctests but include everything else.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -156,7 +156,7 @@ jobs:
       - uses: actions/checkout@v5
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly-2025-06-01
+          toolchain: nightly-2025-09-28
           components: rust-src
       - env:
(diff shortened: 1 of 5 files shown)


newpavlov (approved):

apoelstra:
Would it be possible to backport the doc_auto_cfg removal to 0.2?

PR #739, opened 2025-10-15
Let RawOsError be a pub type

dhardy:
See https://github.com/rust-random/rand/pull/1665

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -10,16 +10,18 @@ cfg_if::cfg_if!(
 
         /// Raw error code.
         ///
-        /// This alias mirrors unstable `std::io::RawOsError`:
-        /// <https://doc.rust-lang.org/std/io/type.RawOsError.html>
+        /// This alias mirrors unstable [`std::io::RawOsError`].
+        ///
+        /// [`std::io::RawOsError`]: https://doc.rust-lang.org/std/io/type.RawOsError.html
         pub type RawOsError = usize;
         type NonZeroRawOsError = core::num::NonZeroUsize;
         const UEFI_ERROR_FLAG: RawOsError = 1 << (RawOsError::BITS - 1);
     } else {
         /// Raw error code.
         ///
-        /// This alias mirrors unstable `std::io::RawOsError`:
-        /// <https://doc.rust-lang.org/std/io/type.RawOsError.html>
+        /// This alias mirrors unstable [`std::io::RawOsError`].
+        ///
+        /// [`std::io::RawOsError`]: https://doc.rust-lang.org/std/io/type.RawOsError.html
         pub type RawOsError = i32;
         type NonZeroRawOsError = core::num::NonZeroI32;
     }


src/error.rs
```diff
@@ -9,11 +9,14 @@ cfg_if::cfg_if!(
     if #[cfg(target_os = "uefi")] {
         // See the UEFI spec for more information:
         // https://uefi.org/specs/UEFI/2.10/Apx_D_Status_Codes.html
-        type RawOsError = usize;
+        pub type RawOsError = usize;
```
newpavlov, line 16: With this `RawOsError` will not have a docstring on UEFI targets.

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- New `pub type RawOsError` [#739]
```
```suggestion
- `RawOsError` type alias [#739]
```

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
newpavlov, line 7: Missing link.

src/error.rs
dhardy: Will anyone ever generate docs for the UEFI target? docs.rs certainly won't, and I don't think any dev environment will.

CHANGELOG.md
dhardy: It's not a link.

newpavlov: Yes, but it should be. Before cutting the v0.3.4 release we had `[0.3.4]: https://github.com/rust-random/getrandom/compare/v0.3.3...HEAD` and we should do the same here.

dhardy: Come to think of it, this is where a custom CHANGELOG validator would be useful. (Or even contributor docs.)

src/error.rs
dhardy: Interesting that CI fails because of this though. Should we require docs anyway or change the CI to only require docs on dev/doc platforms?

newpavlov: Let's maintain docstrings properly regardless whether we think users will generate docs for such corner case or not.

newpavlov: Moreover, it triggers the `missing_docs` warning.

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
...
+
+### Added
+- `RawOsError` type alias [#739]
+
+[Unreleased]: https://github.com/rust-random/getrandom/compare/v0.3.4...master
```
newpavlov: We keep diff links in the end of the file.

CHANGELOG.md
newpavlov: NVM my previous (deleted) comment. The recommended [format](https://keepachangelog.com/en/1.0.0/) uses `[Unreleased]` links.

src/error.rs
newpavlov: I suggest a small tweak:
```rust
// This alias mirrors unstable `std::io::RawOsError`:
// https://doc.rust-lang.org/std/io/type.RawOsError.html
cfg_if::cfg_if!(
    if #[cfg(target_os = "uefi")] {
        /// Raw error code
        // See the UEFI spec for more information:
        // https://uefi.org/specs/UEFI/2.10/Apx_D_Status_Codes.html
        pub type RawOsError = usize;
        type NonZeroRawOsError = core::num::NonZeroUsize;
        const UEFI_ERROR_FLAG: RawOsError = 1 << (RawOsError::BITS - 1);
    } else {
        /// Raw error code
        pub type RawOsError = i32;
        type NonZeroRawOsError = core::num::NonZeroI32;
    }
);
```

newpavlov (approved):

PR #740, opened 2025-10-15
Add release.yml CI action

dhardy:
This is the repo-side configuration of https://crates.io/docs/trusted-publishing

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Create release body
       id: create_release_body
       run: |
-        RELEASEVERSION="[${{ '[' + github.ref_name.slice(1) + ']' }}]"
+        RELEASEVERSION="[${{ github.ref_name.slice(1) }}]"
         echo "Version: $RELEASEVERSION"
         RELEASEBODY=$(awk -v ver="$RELEASEVERSION" '/^## / { if (p) { exit }; if ($2 == ver) { p=1; next } } p && NF' CHANGELOG.md)
         {


dhardy:
Added code from https://medium.com/@usman_qb/from-git-tag-to-github-release-automating-changelog-extraction-9cfcad1b13c4 (modified slightly due to our tag format).

I don't know how to test this other than to make a release.

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+      id: auth
+    - run: cargo publish
+      env:
+        CARGO_REGISTRY_TOKEN: ${{ steps.auth.outputs.token }}
+
+    # Source: https://medium.com/@usman_qb
+    - name: Create release body
+      id: create_release_body
+      run: |
+        RELEASEVERSION="[${{ '[' + github.ref_name.slice(1) + ']' }}]"
```
dhardy: Assumption: we can evaluate JavaScript here. GH docs seem to imply this.

dhardy:
I believe this should work as intended, or at least not cause problems:

- The `cargo publish` part is copied from the crates.io trusted-publishing docs above. It should either work or fail to do anything. If we want to automate publishing using GitHub actions this is the best solution that I am aware of.
- The GitHub release script is a little more flaky, but the worst it can do is create a poor draft release on GH. I have tested what I can outside of the GitHub actions script.

Testing needed, *somewhere*.

newpavlov:
LGTM on the first glance, but note that I do not have experience of using Trusted Publishing.

dhardy:
Me neither, and while not giving GH Actions long-lived tokens is a good thing, this still fundamentally gives Actions the ability to publish crates. Whether or not we want that is another question.

(My main motivation here was exploring possibilities. It may be that we want the GH release stuff but not the crates.io publishing.)

What I do like is that this approach associates crates.io releases with git tags (assuming one does not publish to crates.io first).

dhardy:
Closing: #748 replaces the crates.io part; the GH part is trickier and should probably use a separate job (refer to our experiments with https://github.com/rust-random/core/).

PR #742, opened 2025-10-24
chore: make some documents clearer

reddaisyy:
make some documents clearer

PR #743, opened 2025-10-24
windows: tweak comment about ProcessPrng

newpavlov:
The comment was not clear enough and triggered LLM checker in #742.

### DIFF
--- a/src/backends/windows.rs
+++ b/src/backends/windows.rs
@@ -51,9 +51,9 @@ const TRUE: BOOL = 1;
 #[inline]
 pub fn fill_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
     let result = unsafe { ProcessPrng(dest.as_mut_ptr().cast::<u8>(), dest.len()) };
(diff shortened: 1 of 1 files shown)


src/backends/windows.rs
```diff
@@ -51,9 +51,7 @@ const TRUE: BOOL = 1;
 #[inline]
 pub fn fill_inner(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
     let result = unsafe { ProcessPrng(dest.as_mut_ptr().cast::<u8>(), dest.len()) };
-    // Since Windows 10, calls to the user-mode RNG are guaranteed to never
-    // fail during runtime (rare windows W); `ProcessPrng` will only ever
-    // return 1 (which is how windows represents TRUE).
+    // `ProcessPrng` is documented to always return TRUE.
```
newpavlov: I couldn't find anything about `ProcessPrng` on Windows 8, so it may not be technically guaranteed on it, but I think we can omit this part from the comment.

PR #748, opened 2025-11-08
Use trusted-publishing on crates.io

dhardy:
Lets use https://crates.io/docs/trusted-publishing to not require local crates.io keys.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,17 @@
+name: Publish to crates.io
+on:
+  push:
+    tags: ['v*']  # Triggers when pushing tags starting with 'v'
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    environment: release  # Optional: for enhanced security
+    permissions:
+      id-token: write     # Required for OIDC token exchange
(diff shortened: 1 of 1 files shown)


newpavlov (approved):
You forgot to setup the GitHub environment (I did it now). You also need to update the crate settings.

newpavlov:
Do we also want a GitHub release job similar to `rand_core`?

dhardy:
> You forgot to setup the GitHub environment

Should it be using `v*` tags instead of the master branch?

> Do we also want a GitHub release job similar to `rand_core`?

Personally I don't care much (or even if we make GH releases at all). For a new PR anyway.

newpavlov:
>Should it be using v* tags instead of the master branch?

Ideally, we would use an AND condition, but, unfortunately, IIUC GitHub does not currently support them. So I think publishing from the master branch is a more important restriction.

>Personally I don't care much (or even if we make GH releases at all)

I feel the same. But it would be nice to be consistent with other crates in the org (i.e. we should add GH release job for all repos or remove it from `rand_core`).

Issue #722, opened 2025-09-19
Tracking issue for v0.4 release


newpavlov:
List of potential breaking changes:

- [ ] Bump MSRV to 1.85 and edition to 2024
- [ ] Relax MSRV policy to allow MSRV bumps in patch releases, i.e. we would rely on the MSRV-aware resolver going forward
- [ ] Remove the `wasm_js` backend in favor of the custom fallback crate feature (also see https://github.com/wasm-bindgen/wasm-bindgen/issues/4667)

It's arguably a good time to synchronize v0.4 release with the [planned](https://github.com/rust-random/rand/issues/1643) `rand` v0.10 release (which also bumps MSRV to 1.85).

briansmith:
> Bump MSRV to 1.85 and edition to 2024

Why? I see no significant advantage of doing this.

> Relax MSRV policy to allow MSRV bumps in patch releases

+1. But this could be done even in 0.3. No need for 0.4.

newpavlov:
>But this could be done even in 0.3. No need for 0.4.

No, it can not be according to the MSRV policy used in v0.3.

briansmith:
> No, it can not be according to the MSRV policy used in v0.3.

The MSRV policy for 0.3 can be changed.

newpavlov:
We also can break semver compatibility. It does not mean we should do it.

I think it's better to just make a new breaking release and clearly inform downstream users about the change.

briansmith:
OK. The proposed 0.4 release seems to only have negative effects, so I would expect many projects to keep using older releases, especially since I think ports and whatnot get back ported to older releases now.

dhardy:
@briansmith your arguments would make sense for an interface crate like `rand_core` but the only interface for foreign code in `getrandom` is that to support custom backends, which IIUC are uncommon. I therefore fail to see why it matters much whether the next release is semver breaking? Is it just because inevitably some binaries will end up using multiple copies of `getrandom` (v0.3 + v0.4) for a while?

briansmith:
The MSRV for my crate is 1.66. Some versions of 0.3 have an MSRV less than that. If all versions of getrandom 0.4 have an MSRV greater than that, then why bother switching from 0.3 to 0.4? (FWIW, currently I am using 0.2 for similar but different reasons.)

tarcieri:
One nice thing about the higher MSRV is `getrandom::Error` now impls `core::error::Error`.

[...]

PR #749, opened 2025-11-08
Use Edition 2024, MSRV 1.85

dhardy:
The MSRV bump is required *if* we want to make `getrandom` depend on `rand_core` (see https://github.com/rust-random/rand/issues/1675). *Unless* we use a split MSRV depending on whether `rand_core` is a dependency.

It is however a significant bump, so we might prefer not to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `RawOsError` type alias [#739]
 
-### Changes
+### Changed
 - Use Edition 2024 and MSRV 1.85 [#749]
 
 [#739]: https://github.com/rust-random/getrandom/pull/739


newpavlov:
I am in favor of bumping MSRV and releasing v0.4 (see #722) regardless of whether `OsRng` will be part of `getrandom` or not.

(I will wait for CI to be fixed before approving)

.github/workflows/tests.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-24.04, windows-2022]
-        toolchain: [nightly, beta, stable, "1.63"]
+        toolchain: [nightly, beta, stable, "1.85"]
```
newpavlov, line 26: Ideally, we should extract MSRV from the `rust-version` field instead of manually providing it.

dhardy:
Releasing v0.4 works for me. Okay, I'll try to get this fixed.

dhardy:
So yesterday CI showed this error:
```
error[E0133]: call to unsafe function `core::arch::x86_64::_rdrand64_step` is unsafe and requires unsafe block
  --> src/backends/rdrand.rs:34:12
   |
34 |         if rdrand_step(&mut val) == 1 {
   |            ^^^^^^^^^^^^^^^^^^^^^ call to unsafe function
```
Today after adding `unsafe`, the RDRAND UEFI job fails:
```
error: unnecessary `unsafe` block
  --> src/backends/rdrand.rs:34:12
   |
34 |         if unsafe { rdrand_step(&mut val) } == 1 {
   |            ^^^^^^ unnecessary `unsafe` block
   |
   = note: `-D unused-unsafe` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(unused_unsafe)]`
```
This is using nightly rustc. So why does it not think unsafe is necessary here?

dhardy:
Note: `core::error::Error` is stable since 1.81 so we may be able to drop the `std` feature after this. (New PR.)

src/backends/rdrand.rs
```diff
@@ -28,10 +28,11 @@ static RDRAND_GOOD: lazy::LazyBool = lazy::LazyBool::new();
...
 #[target_feature(enable = "rdrand")]
+#[cfg_attr(target_os = "uefi", allow(unused_unsafe))] // HACK: Rust lint gives false positive on uefi
 unsafe fn rdrand() -> Option<Word> {
     for _ in 0..RETRY_LIMIT {
         let mut val = 0;
-        if rdrand_step(&mut val) == 1 {
+        if unsafe { rdrand_step(&mut val) } == 1 {
             return Some(val);
         }
     }
```
dhardy, line 38: Got a better idea?

At least `allow(unused_unsafe)` is fairly harmless.

CHANGELOG.md
```diff
@@ -9,7 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `RawOsError` type alias [#739]
 
+### Changes
```
```suggestion
### Changed
```

src/backends/rdrand.rs
newpavlov: It's not quite a false positive... On Nightly the RDRAND intrinsics are marked as safe (e.g. see [here](https://rust.godbolt.org/z/Txb8jrKes)). I am not sure why it's still marked as unsafe on the stable toolchain, since most intrinsics were changed to safe in 1.87 (after 1.86 relaxed unsafety requirements on functions marked with `#[target_feature(enable="..")]`).

For now, I would just mark the whole function with `allow(unused_unsafe)` without gating it on the target with a comment stating that in new Rust versions RDRAND intrinsics are safe.

PR #750, opened 2025-11-10
Implement core::error::Error without std

dhardy:
There are several other places that `std` is used.

### DIFF
--- a/src/error_std_impls.rs
+++ b/src/error_std_impls.rs
@@ -11,5 +11,3 @@ impl From<Error> for io::Error {
         }
     }
 }
-
-impl std::error::Error for Error {}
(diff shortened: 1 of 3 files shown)


src/error_std_impls.rs
```diff
@@ -1,8 +1,11 @@
+#[cfg(feature = "std")]
 extern crate std;
 
 use crate::Error;
+#[cfg(feature = "std")]
 use std::io;
 
+#[cfg(feature = "std")]
 impl From<Error> for io::Error {
```
newpavlov, line 6:
```suggestion
impl From<Error> for std::io::Error {
```
and remove `use std::io`.

src/error_std_impls.rs
```diff
@@ -1,8 +1,11 @@
+#[cfg(feature = "std")]
 extern crate std;
```
newpavlov, line 1: Move it to lib.rs?

dhardy:
Updated

PR #751, opened 2025-11-10
Add `SysRng` behind new feature `sys_rng`

dhardy:
This was mentioned elsewhere: https://github.com/rust-random/rand/issues/1675

Adds `rand_core` as a dependency, which is one small crate.

This attempts to enable testing of `rng` in all CI tests.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,7 +46,7 @@ mod sys_rng;
 #[cfg(feature = "sys_rng")]
 pub use rand_core;
 #[cfg(feature = "sys_rng")]
-pub use sys_rng::{SysRng, UnwrappedSysRng};
+pub use sys_rng::SysRng;
 
 pub use crate::error::{Error, RawOsError};
 
--- a/src/sys_rng.rs
+++ b/src/sys_rng.rs
@@ -52,10 +52,3 @@ impl TryRngCore for SysRng {
 }
 
 impl TryCryptoRng for SysRng {}
-
-/// An [`RngCore`](rand_core::RngCore) interface over the system's preferred random number source
-///
-/// # Panics
-///
-/// This wrapper will panic on error (see [crate docs](crate#error-handling)).
-pub type UnwrappedSysRng = rand_core::UnwrapErr<SysRng>;
--- a/tests/sys_rng.rs
+++ b/tests/sys_rng.rs
@@ -1,7 +1,7 @@
 #![cfg(feature = "sys_rng")]
 
-use getrandom::rand_core::{RngCore, TryRngCore};
-use getrandom::{SysRng, UnwrappedSysRng};
+use getrandom::SysRng;
+use getrandom::rand_core::TryRngCore;
 
 #[test]
 fn test_sys_rng() {
@@ -14,11 +14,4 @@ fn test_sys_rng() {
 #[test]
 fn test_construction() {
     assert!(SysRng.try_next_u64().unwrap() != 0);
-}
-
-#[test]
-fn test_unwrapped_sys_rng() {
-    let mut buf = [0u8; 128];
-    UnwrappedSysRng::default().fill_bytes(&mut buf);
-    assert!(buf != [0; 128]);
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `RawOsError` type alias [#739]
-- `SysRng` and `UnwrappedSysRng` behind new feature `sys_rng` [#751]
+- `SysRng` behind new feature `sys_rng` [#751]
 
 ### Changed
 - Use Edition 2024 and MSRV 1.85 [#749]


dhardy:
@newpavlov your opinion on this please?

There is an alternative: the `rand_os` crate. I slightly prefer fewer crates but either is acceptable.

newpavlov:
As I wrote previously, personally, I have a slight preference for a separate crate. But I would be fine with this PR as well, if others prefer it.

P.S.: Note that `rand_os` recently got an "unmaintained" advisory: https://github.com/rustsec/advisory-db/pull/2462

.github/workflows/tests.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo test
       # Make sure enabling the std feature doesn't break anything
-      - run: cargo test --features=std
+      - run: cargo test --features=std,rng
```
josephlr: Should we have a test somewhere that `rng` works without the `std` feature? Maybe here we could have:

```suggestion
      - run: cargo test --features=std
      - run: cargo test --features=rng
      - run: cargo test --features=std,rng
```

I agree that having most of the testing use `--features=std,rng` is correct.

Cargo.toml
josephlr: This should be updated to include `"rng"` so that users can see the generated docs for stuff gated behind `rng`

src/rng.rs
```diff
@@ -0,0 +1,60 @@
...
+///
+/// let mut key = [0u8; 32];
+/// OsRng.try_fill_bytes(&mut key).unwrap();
+///
+/// let mut rng = OsRng.unwrap_err();
+/// let random_u64 = rng.next_u64();
+/// ```
+///
+/// [rand]: https://crates.io/crates/rand
+/// [`rand::rngs::OsRng`]: https://docs.rs/rand/latest/rand/rngs/struct.OsRng.html
```
josephlr: Similar to the docs for `getrandom::Error`, we should mention that this will require a crate to enable the `rng` feature. We should also mention here if the `rng` feature will have different MSRV or stability guarantees going forward. Are we allowed to increase the minor version of `rand_core` (from say `0.10` to `0.11`) when moving from `getrandom 0.4.0` to `getrandom 0.4.1`?

Not sure what our policy should be, but we should document it.

src/rng.rs
```diff
@@ -0,0 +1,60 @@
...
+    let x = OsRng.try_next_u64().unwrap();
+    let y = OsRng.try_next_u64().unwrap();
+    assert!(x != 0);
+    assert!(x != y);
+}
+
+#[test]
+fn test_construction() {
+    assert!(OsRng.try_next_u64().unwrap() != 0);
+}
```
josephlr: These should be in a `mod test` with `#[cfg(test)]`

src/lib.rs
```diff
@@ -39,6 +39,13 @@ mod util;
 #[cfg(feature = "std")]
 mod error_std_impls;
 
+#[cfg(feature = "rng")]
+mod rng;
+#[cfg(feature = "rng")]
+pub extern crate rand_core;
```
josephlr:
```suggestion
pub use rand_core;
```

This is necessary for rustdoc to correctly:
  - Render descriptions for all the other items
  - Mark this re-export as gated behind the `rng` feature.

src/rng.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// An RNG over the operating-system's random data source
+///
+/// This is a zero-sized struct. It can be freely constructed with just `OsRng`.
+///
+/// This struct is also available as [`rand::rngs::OsRng`] when using [rand].
+///
+/// # Usage example
+/// ```
+/// use getrandom::{rand_core::{TryRngCore, RngCore}, OsRng};
```
josephlr:
```suggestion
/// # Usage example
/// 
/// The functionality of this struct is only accessible though its
/// [`TryRngCore`] implementation:
/// ```
/// use getrandom::{OsRng, rand_core::TryRngCore};
```

We should make it clear that the only way to use `OsRng` is via `TryRngCore`.

src/lib.rs
josephlr: Is there a reason to re-export all of `rand_core`? Would it make more sense to only re-export `TryRngCore` as that's the only non-marker trait `OsRng` implements?

We could then also add additional documentation explaining the re-export:
```suggestion
#[cfg(feature = "rng")]
/// Re-exported base RNG trait from [`rand_core`]
///
pub use rand_core::TryRngCore;
```

src/rng.rs
dhardy: `doc-auto-cfg` was stabilised so docs will note that `OsRng` requires feature `rng`, if that's what you mean.

> Are we allowed to increase the minor version of `rand_core` (from say `0.10` to `0.11`) when moving from `getrandom 0.4.0` to `getrandom 0.4.1`?

That would break semver since `rand_core_0_10::TryRngCore` and `rand_core_0_11::TryRngCore` are considered separate traits, so it should be clear that we cannot do that.

> We should also mention here if the `rng` feature will have different MSRV or stability guarantees going forward.

I don't see why it would.

src/lib.rs
dhardy: > * Mark this re-export as gated behind the `rng` feature.

It's not, but it does appear to cause render issues (all following text is fixed width / `code`). Weird.

dhardy: @newpavlov convinced me in the past that this is a good general policy, and I agree:

- It allows downstream crates to access `rand_core` functionality without explicitly depending on that
- It does not pretend that `TryRngCore` is part of `getrandom`
- It reduces the chance of people hitting "multiple versions of `rand_core`" errors

dhardy:
Thanks for moving this forward @josephlr. I think I addressed your concerns (except as commented).

src/lib.rs
newpavlov: IMO re-exporting `rand_core` is a must have and should be done by all RNG implementation crates.

src/rng.rs
```diff
@@ -0,0 +1,74 @@
...
+///
+/// let mut rng = OsRng.unwrap_err();
+/// let random_u64 = rng.next_u64();
+/// ```
+///
+/// [rand]: https://crates.io/crates/rand
+/// [`rand::rngs::OsRng`]: https://docs.rs/rand/latest/rand/rngs/struct.OsRng.html
+/// [`RngCore`]: rand_core::RngCore
+#[derive(Clone, Copy, Debug, Default)]
+pub struct OsRng;
```
newpavlov: I would prefer for it to be named as `SysRng`, since `OsRng` can be a confusing name on `no_std` targets. To make migration easier we could add `type OsRng = SysRng;` alias in `rand` (with or without `#[deprecated]` attribute).

Cargo.toml
```diff
@@ -23,8 +26,12 @@ std = []
 # i.e. avoid unconditionally enabling it in library crates.
 wasm_js = ["dep:wasm-bindgen", "dep:js-sys"]
 
+# Provide OsRng over rand_core
+rng = ["dep:rand_core"]
```
newpavlov: How about naming the feature `os_rng` (or `sys_rng` if the type will be renamed)?

src/rng.rs
```diff
@@ -0,0 +1,74 @@
...
+    #[inline]
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), Error> {
+        crate::fill(dest)
+    }
+}
+
+impl TryCryptoRng for OsRng {}
+
+#[cfg(test)]
+mod test {
```
newpavlov: I think it's better to define these tests in `tests/rng.rs` or integrate them into the doctest.

src/rng.rs
newpavlov: >doc-auto-cfg was stabilised 

Are you sure? I thought it was just merged into `doc_cfg`. In RustCrypto we still use `#![cfg_attr(docsrs, feature(doc_cfg))]`.

src/rng.rs
dhardy: Motivation? Having tests close to the feature definition is slightly preferable in my opinion (where tests are not too extensive at least).

src/rng.rs
dhardy: Makes some sense... some backends like RDRAND are not OS features.

@josephlr?

src/rng.rs
newpavlov: My general preference is to keep public API tests in integration tests (i.e. in the tests folder) and/or doctest, while using test modules (i.e. unit tests) only for private API testing.

Since these tests are quite small, I think they can be simply merged into the `OsRng` doctest, potentially with `# ` used to hide asserts.

dhardy: If the motivation is to improve visibility of integration tests, then moving to `tests` makes sense.

I'm disinclined to push these into a doc-test since (a) it hides it, (b) I would normally assume that doc-tests are only for documentation purposes and (c) failure reporting for doc tests is worse (slower and worse messages) than for other types of test.

So I don't understand why you prefer *both* integration tests *and* doc tests over a local `test` module.

newpavlov: It's just a matter of not duplicating tests which are already covered by doctests. I am fine with keeping the example as-is and moving the `test` module into the `tests/` folder.

dhardy:
@newpavlov I addressed your concerns

src/rng.rs
```diff
@@ -0,0 +1,74 @@
...
+    #[inline]
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), Error> {
+        crate::fill(dest)
+    }
+}
+
+impl TryCryptoRng for SysRng {}
+
+#[cfg(test)]
+mod test {
```
newpavlov: This module is no longer needed since you moved the tests to `tests/rng.rs`.

tests/rng.rs
```diff
@@ -0,0 +1,17 @@
+#![cfg(feature = "sys_rng")]
+
+use getrandom::SysRng;
+use getrandom::rand_core::TryRngCore;
+
+#[test]
+fn test_os_rng() {
```
newpavlov: Rename to `test_sys_rng`.

src/lib.rs
```diff
@@ -39,6 +39,13 @@ mod util;
 #[cfg(feature = "std")]
 mod error_std_impls;
 
+#[cfg(feature = "sys_rng")]
+mod rng;
```
newpavlov: Rename the module to `sys_rng`?

.github/workflows/tests.yml
```diff
@@ -273,14 +276,14 @@ jobs:
...
-              args: '--features=std,wasm_js',
+              args: '--features=std,rng,wasm_js',
             }
           - {
               description: Web with Atomics,
               version: nightly,
               components: rust-src,
               flags: '-Dwarnings --cfg getrandom_backend="wasm_js" -Ctarget-feature=+atomics,+bulk-memory',
-              args: '--features=std,wasm_js -Zbuild-std=panic_abort,std',
+              args: '--features=std,rng,wasm_js -Zbuild-std=panic_abort,std',
```
newpavlov: `rng` -> `sys_rng`

src/sys_rng.rs
```diff
@@ -0,0 +1,74 @@
...
+///
+/// `SysRng` implements [`TryRngCore`]:
+/// ```
+/// use getrandom::{rand_core::TryRngCore, SysRng};
+///
+/// let mut key = [0u8; 32];
+/// SysRng.try_fill_bytes(&mut key).unwrap();
+/// ```
+///
+/// Using it as an [`RngCore`] is possible using [`TryRngCore::unwrap_err`]:
```
newpavlov, line 20: Maybe add `type UnwrapSysRng = UnwrapErr<SysRng>;` for convenience?

dhardy: The grammatically correct name would be `UnwrappedSysRng`. Should we? I think maybe not.

It is default-constructible, otherwise I would definitely say no.

newpavlov: I am fine with either name. I can't say I am happy with `Unwrap(ed)SysRng`, but I don't have better ideas right now.

I think such alias can be convenient when dealing with APIs which expect `&mut impl RngCore`.

src/sys_rng.rs
```diff
@@ -0,0 +1,56 @@
+//! rand_core adapter
```
newpavlov: I think it's better to move this module description to `lib.rs`:
```rust
/// `rand_core` adapter
#[cfg(feature = "sys_rng")]
mod sys_rng;
```

src/rng.rs
newpavlov: @dhardy 
I checked doc generation on the latest Nightly and it looks like we still need `#![cfg_attr(docsrs, feature(doc_cfg))]` to properly render the feature-gating information.

newpavlov:
@josephlr
WDYT? It looks like the only unresolved question is whether we need to introduce `UnwrappedSysRng` or not.

@dhardy 
Please add `#![cfg_attr(docsrs, feature(doc_cfg))]` as per the [comment above](https://github.com/rust-random/getrandom/pull/751#discussion_r2549870824) and add a changelog entry.

dhardy:
@newpavlov see `lib.rs:12`.

src/sys_rng.rs
dhardy: The only other reasonable naming I can think of is `SysRng` (with `TrySysRng` for `TryRngCore`). I don't like this though: it promotes the panicking interface as the default one.

`PanickingSysRng` isn't a great name. `InfallibleSysRng` is incorrect.

So `UnwrappedSysRng` is still the best name if we need it (for use in generics).

dhardy:
A consequence of renaming the feature to `sys_rng` is:
image
Would it be better just to name the feature `rng`?

dhardy:
I added `UnwrappedSysRng` (with a slight doc tweak)

newpavlov:
What about changelog?

It could worth to add `test_unwrapped_sys_rng`.

>Would it be better just to name the feature rng?

Personally, I don't see a problem with the provided rendering.

dhardy:
Added CHANGELOG entry and test case.

tests/sys_rng.rs
```diff
@@ -0,0 +1,24 @@
...
+
+#[test]
+fn test_construction() {
+    assert!(SysRng.try_next_u64().unwrap() != 0);
+}
+
+#[test]
+fn test_unwrapped_sys_rng() {
+    let mut buf = [0u8; 128];
+    UnwrappedSysRng::default().fill_bytes(&mut buf);
```
newpavlov: Hm... I haven't considered that we have to use `UnwrappedSysRng::default()` or `UnwrapErr(SysRng)` to initialize the "unwrapped" variant.

Maybe it's better to define `pub struct UnwrappedSysRng;` and implement the traits manually instead of relying on `UnwrapErr`?

dhardy: No because then `SysRng.unwrap_err()` won't return an item of type `UnwrappedSysRng`.

newpavlov: Granted, it may be mildly confusing, but it would it matter in practice? I think most users who don't care about `SysRng` error handling would simply use `UnwrappedSysRng` without touching the `unwrap_err` method.

dhardy: But what's the benefit? Being able to just write `UnwrappedSysRng` for construction? Not worth the potential confusion IMO.

newpavlov: >Being able to just write `UnwrappedSysRng` for construction?

Yes. I think `UnwrappedSysRng` could be more popular in practice than `SysRng`, since most users probably do not care about potential errors and fine with panic if they are encountered.

I think we still need some "salt" for the unwrapped variant, but IMO `UnwrappedSysRng::default()` goes a bit too far.

dhardy: I disagree: most users will ultimately use something higher-level like `rand` or some cipher crate. `getrandom` should not care about sugar.

newpavlov: I thought `SysRng` and `UnwrappedSysRng` will be re-exported from `rand`, no? I think there is plenty of `OsRng` uses in practice and it's worth to consider user convenience here.

dhardy: `SysRng`, yes.

`UnwrappedSysRng`, I don't know. I don't feel like it's a bad thing making `SysRng` slightly harder to use where an `RngCore` is wanted.

I still feel like it doesn't belong here though; shall I remove the typedef?

newpavlov: I guess we can remove it from this PR and leave potential addition of `UnwrappedSysRng` for a separate PR.

Personally, I think `pub struct UnwrappedSysRng;` would be useful to have and the long name is enough salt to gently discourage its use in favor of `SysRng`.

PR #754, opened 2025-11-24
Add `UnwrappingSysRng`

newpavlov:
The type implements `RngCore`/`CryptoRng` traits by unwrapping potential errors. It could be useful for users who want to use the `RngCore` trait, do not care about potential (extremely unlikely errorr) errors, and fine with the panicking behavior.

The `rand_core::UnwrapErr` wrapper is not used to simplify the type initialization (just `UnwrappingSysRng` vs. `UnwrappingSysRng::default()` or `UnwrapErr(SysRng)`).

The intentionally long name is probably sufficient to discourage its use in favor of `SysRng`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `RawOsError` type alias [#739]
 - `SysRng` behind new feature `sys_rng` [#751]
+- `UnwrappingSysRng` behind new feature `sys_rng` [#754]
 
 ### Changed
 - Use Edition 2024 and MSRV 1.85 [#749]
@@ -16,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#739]: https://github.com/rust-random/getrandom/pull/739
 [#749]: https://github.com/rust-random/getrandom/pull/749
 [#751]: https://github.com/rust-random/getrandom/pull/751
+[#754]: https://github.com/rust-random/getrandom/pull/754
 
 ## [0.3.4] - 2025-10-14
 


newpavlov:
I think `UnwrappingSysRng` is a bit more correct than `UnwrappedSysRng` since it's "an RNG which unwraps" and not "an unwrapped RNG". Plus, it's one char longer.

src/sys_rng.rs
```diff
@@ -52,3 +50,49 @@ impl TryRngCore for SysRng {
...
+///
+/// let x: u32 = UnwrappingSysRng.next_u32();
+/// let y: u64 = UnwrappingSysRng.next_u64();
+/// ```
+///
+/// [`rand`]: https://crates.io/crates/rand
+/// [`rand::rngs::UnwrappingSysRng`]: https://docs.rs/rand/latest/rand/rngs/struct.UnwrappingSysRng.html
+/// [`RngCore`]: rand_core::RngCore
+#[derive(Clone, Copy, Debug, Default)]
+pub struct UnwrappingSysRng;
```
josephlr, line 79: Personally, I'm not a fan of introducing another type which is just "`SysRng` but we unwrap all errors".

Currently, there are two ways to already do this with `rand_core` (which is entirely exported by this crate):
  - ```rust
    use getrandom::{SysRng, rand_core::TryRngCore};
    
    let mut rng = SysRng.unwrap_err();
    ```
  - ```rust
    use getrandom::{SysRng, rand_core::UnwrapErr};
    
    let mut rng = UnwrapErr(SysRng);
    ```

I don't think we need to add a third way, especially if (as you say) we recommend using `SysRng` instead of `UnwrappingSysRng`. Also, the two existing methods produce the same type (`UnwrapErr<SysRng>`), while this approach would add a new type.

I think we should instead update the docstring for `SysRng` to:
  - Have the initial usage examples use [`TryRngCore`] without `unwrap`ing the error (like you already have)
  - Include an additional example section on how someone might use this with `RngCore`. Would look something like:

````md
## Use with [`RngCore`]

Unlike the use of [`TryRngCore`] above, the [`RngCore`] trait does not allow returning
an error type. While it is possible for [`SysRng`]/[`fill`] to fail, this is usually
uncommon on most platforms (see [link to main crate docs]). For this reason, users may
want to wrap [`SysRng`] with [`rand_core::UnwrapErr`], to automatically unwrap the unlikely
errors.

```rust
use getrandom::{SysRng, rand_core::{RngCore, UnwrapErr}};

let x: u32 = UnwrapErr(SysRng).next_u32();
let y: u64 = UnwrapErr(SysRng).next_u64();
```

This can also be done via the [`TryRngCore::unwrap_err()`] method as
`let mut rng = SysRng.unwrap_err();`
````

File: src/sys_rng.rs [hunk 2]
```diff
@@ -52,3 +50,49 @@ impl TryRngCore for SysRng {
...
+///
+/// If possible, we recommend to use [`SysRng`] instead and to properly handle potential errors.
+///
+/// This struct is also available as [`rand::rngs::UnwrappingSysRng`] when using [`rand`].
+///
+/// # Usage example
+///
+/// `UnwrappingSysRng` implements [`RngCore`]:
+/// ```
+/// use getrandom::{rand_core::RngCore, UnwrappingSysRng};
```
josephlr, line 66: Nit: rustdoc formats this as:
```suggestion
/// use getrandom::{UnwrappingSysRng, rand_core::RngCore};
```

src/sys_rng.rs
newpavlov: I wouldn't say we have "two ways" right now. They are doing the same thing, but expressed differently. Following your classification we would have a bunch of other "different" ways such as `UnwrapErr::<SysRng>::default()`.

The main rationale for the additional type is that in practice the unwrapped variant probably will be used much more often than `SysRng`. While I think we should nudge users towards `SysRng`, I don't think we should unnecessarily annoy them. 

The `UnwrapErr(SysRng)` is quite inconvenient to use because you have to import two separate items and it looks worse in code because it can contain both `UnwrapErr(SysRng)` and `UnwrapErr<SysRng>`. Granted, it's a small papercut, but a papercut nevertheless.

I don't think `UnwrappingSysRng` will cause any issues or confusion in practice. Users are likely to use it without even being aware about `UnwrapErr` existence and it would "just work". Scenarios in which type conflicts between `UnwrappingSysRng` and `UnwrapErr(SysRng)` could happen are highly unlikely in my opinion.

dhardy: > The main rationale for the additional type is that in practice the unwrapped variant probably will be used much more often than `SysRng`.

By users of `rand`, probably yes. But when it comes to direct users of `getrandom` I'm much less sure.

But even for `rand`, I don't know if we want this. Using `UnwrapErr<SysRng>` is not *that* unergonomic and I suspect will not be *that* common.

But I may be wrong. Which in this case is fine since there is no reason this could not be added in the future.

newpavlov: >I suspect will not be that common.

It looks like `OsRng` has [plenty of uses](https://github.com/search?q=OsRng+lang%3Arust&type=code) in practice and on the first glance most of them use it in the panicking mode.

>Which in this case is fine since there is no reason this could not be added in the future.

It's likely to result in an annoying mix of `UnwrapErr<SysRng>` and `UnwrappingSysRng` in downstream crates. I think we either should introduce `UnwrappingSysRng` right away or stick with `UnwrapErr<SysRng>`.

dhardy: Some of those uses are ancient, as evidenced by usage of `OsRng::new()`.

Also, basically all of these uses are just constructing an instance of `OsRng` for local usage, so this would be replaced with `SysRng.unwrap_err()`, not `UnwrapErr<SysRng>`.

I wonder if we should make `unwrap_err` an inherent fn?

newpavlov:
Closing since no one else supports this addition.

PR #755, opened 2025-11-24
Address MSRV TODOs

tamird:
- **Address TODO(MSRV 1.43)**
- **Address TODO(MSRV 1.64)**
- **Address TODO(MSRV 1.65)**
- **Address TODO(MSRV 1.69)**
- **Remove MSRV documentation**
- **Address TODO(MSRV 1.76)**
- **Address TODO(MSRV 1.78)**
- **Address TODO(MSRV 1.79)**
- **Address TODO(MSRV 1.82)**

### DIFF
--- a/src/backends/linux_raw.rs
+++ b/src/backends/linux_raw.rs
@@ -16,6 +16,7 @@ unsafe fn getrandom_syscall(buf: *mut u8, buflen: usize, flags: u32) -> isize {
             target_arch = "arm",
             any(target_abi = "eabi", target_abi = "eabihf"),
         ))] {
+            const __NR_getrandom: u32 = 384;
             // In thumb-mode, r7 is the frame pointer and is not permitted to be used in
             // an inline asm operand, so we have to use a different register and copy it
             // into r7 inside the inline asm.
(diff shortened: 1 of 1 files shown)


tests/mod.rs
```diff
@@ -55,8 +49,8 @@ fn test_diff() {
     fill(&mut v1).unwrap();
     fill(&mut v2).unwrap();
 
-    let mut t1 = uninit_vec(N);
-    let mut t2 = uninit_vec(N);
+    let mut t1 = [const { MaybeUninit::uninit() }; N];
```
newpavlov: I think `const` is not necessary here and we can write `[MaybeUninit::uninit(); N];`.

src/backends/rdrand.rs
```diff
@@ -51,7 +51,7 @@ compile_error!(
 #[target_feature(enable = "rdrand")]
 unsafe fn self_test() -> bool {
     // On AMD, RDRAND returns 0xFF...FF on failure, count it as a collision.
-    let mut prev = !0; // TODO(MSRV 1.43): Move to usize::MAX
+    let mut prev = usize::MAX;
```
newpavlov: ~~I think it's fine to keep `!0` here to not deal with switching between `u32`/`u64`.~~

UPD: You should use `Word::MAX` here.

tests/mod.rs
tamird: Done.

src/backends/rdrand.rs
tamird: Done.

PR #758, opened 2025-11-24
linux_android_with_fallback: detect getrandom

tamird:
Avoid dlsym when statically linking crt.

Relative to https://github.com/rust-random/getrandom/pull/598 this PR keeps dlsym in place when linking dynamically and avoids the dependency on a C compiler. The benefits are as before: avoiding an indirect call when linking statically.

This also contains minor cleanup in `src/backends/netbsd.rs`.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -6,4 +6,61 @@ fn main() {
     if sanitizers.contains("memory") {
         println!("cargo:rustc-cfg=getrandom_msan");
     }
+
+    if cfg!(target_feature = "crt-static") {
+        match std::process::Command::new(std::env::var_os("RUSTC").unwrap())
+            .arg("--target")
+            .arg(std::env::var("TARGET").unwrap())
+            .arg("--out-dir")
+            .arg(std::env::var("OUT_DIR").unwrap())
+            .args(["--crate-type=bin", "-"])
+            .stdin(std::process::Stdio::piped())
+            .stdout(std::process::Stdio::null())
+            .stderr(std::process::Stdio::piped())
+            .spawn()
+        {
+            Err(err) => {
+                println!("cargo:warning=failed to spawn compiler: {}", err);
+            }
+            Ok(mut child) => {
+                use std::io::{BufRead as _, Write as _};
+
+                let std::process::Child { stdin, stderr, .. } = &mut child;
+                let mut stdin = stdin.take().unwrap();
+                stdin
+                    .write_all(
+                        r#"
+    use std::ffi::{c_uint, c_void};
+    extern "C" {
+        fn getrandom(buf: *mut c_void, buflen: usize, flags: c_uint) -> isize;
+    }
+    fn main() -> std::io::Result<()> {
+        use std::convert::TryFrom as _;
+        let mut buf = [0; 1];
+        let _: usize = usize::try_from(unsafe { getrandom(buf.as_mut_ptr().cast(), buf.len(), 0) })
+            .map_err(|std::num::TryFromIntError { .. }| std::io::Error::last_os_error())?;
+        Ok(())
+    }
+    "#
+                        .as_bytes(),
+                    )
+                    .unwrap();
+
+                std::mem::drop(stdin); // Send EOF.
+
+                // Trampoline stdout to cargo warnings.
+                let stderr = stderr.take().unwrap();
+                let stderr = std::io::BufReader::new(stderr);
+                for line in stderr.lines() {
+                    let line = line.unwrap();
+                    println!("cargo:warning={line}");
+                }
+
+                let status = child.wait().unwrap();
+                if status.code() == Some(0) {
+                    println!("cargo:rustc-cfg=has_libc_getrandom");
+                }
+            }
+        }
+    }
 }
--- a/src/backends/netbsd.rs
+++ b/src/backends/netbsd.rs
@@ -9,11 +9,13 @@ use core::{
     ffi::c_void,
     mem::{self, MaybeUninit},
     ptr,
-    sync::atomic::{AtomicPtr, Ordering},
 };
 
 pub use crate::util::{inner_u32, inner_u64};
 
+#[expect(dead_code, reason = "LazyBool is not used")]
+#[path = "../lazy.rs"]
+mod lazy;
 #[path = "../util_libc.rs"]
 mod util_libc;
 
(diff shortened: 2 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -2,9 +2,7 @@ name: Build
 
 on:
   push:
-    branches: master
```
newpavlov, line 5: These changes are not relevant.

newpavlov (commented):
I don't think I like the proposed approach. Not only it's significantly more complex, but I am also not sure about its universality. IIUC it checks that `getrandom` is available on local machine, but compiled binary may be executed in a completely different environment!

Until the minimum supported Linux kernel version is bumped by Rust, I don't think we should statically link to the `getrandom` symbol.

newpavlov:
Also, it's better to move the NetBSD cleanup into a separate PR.

newpavlov:
And don't forget that we have `getrandom_backend="linux_getrandom"`, if you are fine with bumping minimum supported Linux kernel version for your application.

.github/workflows/build.yml
tamird: Sure, they're in a separate commit with reasoning given (it's a bummer you guys insist on squashing on merge).

tamird:
> I don't think I like the proposed approach. Not only it's significantly more complex, but I am also not sure about its universality. IIUC it checks that `getrandom` is available on local machine, but compiled binary may be executed in a completely different environment!
> 
> Until the minimum supported Linux kernel version is bumped by Rust, I don't think we should statically link to the `getrandom` symbol.

Hm. I don't think I follow you here - we're checking if getrandom is available *statically* on the target, i.e. if we're statically linking against MUSL then getrandom is available statically. But we still check at runtime that the target kernel allows it.

newpavlov:
In my understanding, you check that you can link statically on _your host_, so your build script will happily link to glibc's getrandom (to which your application links implicitly when you compile for GNU Linux targets) and will erroneously  report that static linking can be used just fine, while the resulting application could be then deployed on a system with old glibc and Linux kernel.

Even if I misunderstand something and it works correctly in all corner cases, I don't think the additional complexity is justified by the arguably very minor benefit.

src/backends/netbsd.rs
```diff
@@ -42,36 +44,24 @@ unsafe extern "C" fn polyfill_using_kern_arand(
...
-    // probably unnecessary.
-    let mut fptr = GETRANDOM.load(Ordering::Acquire);
-    if fptr.is_null() {
-        fptr = init();
-    }
-    let fptr = unsafe { mem::transmute::<*mut c_void, GetRandomFn>(fptr) };
+    static GETRANDOM_FN: lazy::LazyUsize = lazy::LazyUsize::new();
+
+    let fptr = GETRANDOM_FN.unsync_init(init);
+    let fptr = unsafe { mem::transmute::<usize, GetRandomFn>(fptr) };
```
newpavlov, line 64: Note that this is probably incorrect from the strict pointer provenance point of view.

tamird:
> In my understanding, you check that you can link statically on _your host_, so your build script will happily link to glibc's getrandom (to which your application links implicitly when you compile for GNU Linux targets) and will erroneously report that static linking can be used just fine, while the resulting application could be then deployed on a system with old glibc and Linux kernel.

I don't understand this. If static linking is used, what does it matter what libc is on the target that is deployed to? Yes, the kernel might not support getrandom, but that is why we still check it at runtime using `is_getrandom_good`.

> Even if I misunderstand something and it works correctly in all corner cases, I don't think the additional complexity is justified by the arguably very minor benefit.

Well, that's insurmountable. If you are not open to persuasion, then there's nothing to discuss.

src/backends/netbsd.rs
tamird: I don't think so. The provenance of the original pointer is the same as the provenance of the pointer constructed here.

tamird:
> Also, it's better to move the NetBSD cleanup into a separate PR.

Moved to https://github.com/rust-random/getrandom/pull/759.

.github/workflows/build.yml
josephlr: @tamird I think this is a reasonable thing to do, but changing on which branches CI runs is tricky from a security perspective, and it deserves a dedicated PR.

josephlr (changes requested):
Can you remove the CI changes the changes in #759 to make it easier to actually see what is changing with `linux_android_with_fallback`?

newpavlov:
I missed the `crt-static` part. In the build script you run the linking check only on targets which link to the C runtime statically. In that case this _could_ probably work, but I still do not like the heaps of additional complexity, especially considering that we already [statically link](https://github.com/rust-random/getrandom/blob/master/src/backends/linux_android_with_fallback.rs#L32-L36) to `getrandom` on MUSL targets. I think the `linux_getrandom` opt-in backend is sufficient for users who want to enforce `getrandom` linking on non-MUSL statically linked targets (niche of the niche by itself).

So unless @josephlr has a different opinion, I think we should not use this approach.

build.rs
```diff
@@ -6,4 +6,61 @@ fn main() {
     if sanitizers.contains("memory") {
         println!("cargo:rustc-cfg=getrandom_msan");
     }
+
+    if cfg!(target_feature = "crt-static") {
+        match std::process::Command::new(std::env::var_os("RUSTC").unwrap())
```
briansmith, line 11: I don't think this kind of autoconf-like detection is good to introduce as autoconf-like configuration is antithetical to how Rust projects' build systems work.

tamird: Sounds dogmatic. The real thing we want is a weak symbol but https://github.com/rust-lang/rust/issues/29603.

briansmith:
Is static linkking to libc on Android supported by Android itself? It seems like it was decided to not try to support it in Cargo: https://github.com/rust-lang/cargo/issues/10765. Go project has a similar concern: https://github.com/golang/go/issues/59942#issuecomment-1542691676. It seems like maybe it can be done and is needed for some internal stuff in Android but I am not sure that an APK-packaged app can do it.

I remember we had a discussion about whether we could issue syscalls directly on Android, and we decided to defer them to bionic because Android prefers that. But if we're linking bionic statically then we'd be effectively back to issuing syscalls directly.

tamird:
This PR doesn't change behavior for android, it changes behavior for linux musl.

build.rs
```diff
@@ -6,4 +6,61 @@ fn main() {
     if sanitizers.contains("memory") {
         println!("cargo:rustc-cfg=getrandom_msan");
     }
+
+    if cfg!(target_feature = "crt-static") {
```
briansmith, line 10: I don't think it is good to have such auto-detection at all, but especially we shouldn't do it based on only whether `crt-static` is enabled. If the goal is to do something specific for MUSL on Linux. then we should restrict it to that. And, we should add a comment about why we are doing this.

Though, again, I don't think we should.

briansmith: Further, if the goal is to just verify that `getrandom()` is available and linking succeeds on statically-linked MUSL on Linux, then it is already guaranteed to succeed because MUSL 1.2.3 is the minimum MUSL version since MSRV 1.71 for Linux targets.

briansmith:
> This PR doesn't change behavior for android, it changes behavior for linux musl.

The PR description says "Avoid dlsym when statically linking crt." But, we have:
```
    #[cfg(target_env = "musl")]
    let raw_ptr = {
        let fptr: GetRandomFn = libc::getrandom;
        unsafe { transmute::<GetRandomFn, *mut c_void>(fptr) }
    };
```
I.e. on Linux MUSL we already avoid DLSYM.

build.rs
tamird: The goal is to avoid drive this based on the user's request to statically link crt rather than e.g. hardcoding musl.

briansmith: > The goal is to avoid drive this based on the user's request to statically link crt rather than e.g. hardcoding musl.

Why would it make sense to dynamically set `rustc-cfg=has_libc_getrandom` on Windows (or any other non-Linux target)?

In general, the end-user will know whether to set `has_libc_getrandom` or not, and they can do it in their build system. And they could implement the dynamic detection in their build system, if they are into such things.

Anyway, I think we can close this PR now.

tamird: The point is that the user may just say "i want static linking" and expect appropriate choices to be made by her dependencies.

newpavlov:
Closing per the comments above.

briansmith:
> The PR description says "Avoid dlsym when statically linking crt." But, we have:
> ```
> >  #[cfg(target_env = "musl")]
> ```

Possibly the intent is to support static linking of a libc that isn't musl, bionic, glibc, or uclibc, namely LLVM-libc or whatever it is called. I will suggest a solution in a new issue.

briansmith:
> Possibly the intent is to support static linking of a libc that isn't musl, bionic, glibc, or uclibc, namely LLVM-libc or whatever it is called. I will suggest a solution in a new issue.

Now it's issue #766.

PR #759, opened 2025-11-24
Simplify atomics using `lazy`

tamird:


src/lib.rs
```diff
@@ -33,8 +33,14 @@ extern crate cfg_if;
 use core::mem::MaybeUninit;
 
 mod backends;
+#[allow(dead_code, reason = "not used in all backends")]
 mod error;
+#[allow(dead_code, reason = "not used in all backends")]
+mod lazy;
```
josephlr, line 39: We can't unconditionally include `lazy`. It will cause breakage on platforms without `AtomicUsize`.

src/lib.rs
```diff
@@ -33,8 +33,14 @@ extern crate cfg_if;
...
+#[allow(dead_code, reason = "not used in all backends")]
 mod util;
+#[allow(dead_code, reason = "not used in all backends")]
+mod util_libc;
```
josephlr: We can't unconditionally include `util_libc` as it depends on `libc` which isn't used on all platforms.

src/lib.rs
```diff
@@ -33,8 +33,14 @@ extern crate cfg_if;
 use core::mem::MaybeUninit;
 
 mod backends;
+#[allow(dead_code, reason = "not used in all backends")]
```
josephlr, line 36: Remove this, we only want `#[allow(dead_code)]` on certain methods in the `error` module.

src/lib.rs
```diff
@@ -33,8 +33,42 @@ extern crate cfg_if;
...
+    ),
+    target_os = "netbsd",
+    target_os = "solaris",
+    target_os = "haiku",
+    target_os = "redox",
+    target_os = "nto",
+    target_os = "aix",
+    target_os = "vxworks",
+))]
+mod util_libc;
```
josephlr: Let's not do this, such a long `cfg` is a maintenance headache. It's much easier to just include `util_libc` when we need it.

src/backends/netbsd.rs
```diff
@@ -42,36 +38,24 @@ unsafe extern "C" fn polyfill_using_kern_arand(
...
-    // probably unnecessary.
-    let mut fptr = GETRANDOM.load(Ordering::Acquire);
-    if fptr.is_null() {
-        fptr = init();
-    }
-    let fptr = unsafe { mem::transmute::<*mut c_void, GetRandomFn>(fptr) };
+    static GETRANDOM_FN: lazy::LazyUsize = lazy::LazyUsize::new();
+
+    let fptr = GETRANDOM_FN.unsync_init(init);
+    let fptr = unsafe { mem::transmute::<usize, GetRandomFn>(fptr) };
```
josephlr: As @newpavlov mentioned in https://github.com/rust-random/getrandom/pull/758#discussion_r2557703093 this might be wrong w.r.t. pointer-provenance. See https://doc.rust-lang.org/std/ptr/index.html#provenance

That document explicitly recommends using `AtomicPtr` over `AtomicUsize` for the sorts of manipulations here. We used to have code like what you are suggesting here, but we changed it to its current version to:
  - Align with what's in the standard library (see [`DlsymWeak` in `std`](https://github.com/rust-lang/rust/blob/b64df9d1012f2482b54a4d959548cf8fc67e820c/library/std/src/sys/pal/unix/weak/dlsym.rs#L28))
  - Ensure correct pointer provenance

tamird: Good chance I'm being dense, but what is the provenance problem here? The point of provenance is to say that a point is not just a number, it also carries temporal and mutability "metadata" (conceptually, not physically). The mere act of casting to an integer does not violate provenance.

The vast majority of the discussion in that document (including the suggestion to use AtomicPtr) is focused on mutability, which doesn't apply here because we do not write through this pointer. We also do not offset it or otherwise derive a new address from it on which we need to retain the same provenance.

src/lib.rs
tamird: There are only two `pub (crate)` items on the error module and prior to this change they are both marked `allow(dead_code)`. Which methods do you *not* want this on?

src/lib.rs
tamird: Which platorms, please? Are they not included in CI?

src/lib.rs
tamird: Yeah, this is unfortunate. I'm looking for a nicer solution, if I can't find one I'll undo this bit.

src/lib.rs
josephlr: Interesting, well it used to be the `riscv32imc` platforms (which do not have atomics). However, for some reason `lazy.rs` is building, even on a target like `riscv32imc-unknown-none-elf` which definitely doesn't have atomics. I'll investigate furthur.

src/backends/netbsd.rs
newpavlov: You lose pointer provenance in naive pointer-integer-pointer casts. While it works fine in practice, it's still recommended to avoid it whenever possible.

newpavlov (commented):
I am not convinced that we should "simplify" the NetBSD backend by reusing the `lazy` module. The current code is fairly straightforward and self-contained, I don't see significant issues with it.

This PR also introduces a lot of other IMO unnecessary changes out of scope for the purpose declared in the PR title.

tamird:
Sure, I can break it up into more PRs if you folks prefer. I was just looking for ways to make things a bit more neat. Anyway, I used a macro to introduce `LazyPtr` which I think works nicely now.

briansmith:
> I am not convinced that we should "simplify" the NetBSD backend by reusing the lazy module. The current code is fairly straightforward and self-contained, I don't see significant issues with it.

At least, I don't see how this change makes things easier to understand. The additional abstraction with the macros makes things harder to understand.

>> This PR also introduces a lot of other IMO unnecessary changes out of scope for the purpose declared in the PR title.
> Sure, I can break it up into more PRs if you folks prefer. 

+1.

tamird:
> > I am not convinced that we should "simplify" the NetBSD backend by reusing the lazy module. The current code is fairly straightforward and self-contained, I don't see significant issues with it.
> 
> At least, I don't see how this change makes things easier to understand. The additional abstraction with the macros makes things harder to understand.
> 
> > > This PR also introduces a lot of other IMO unnecessary changes out of scope for the purpose declared in the PR title.
> > > Sure, I can break it up into more PRs if you folks prefer.
> 
> +1.

I'll mark this as draft - said other changes were in #760 (now #763)

newpavlov:
Closing per my [comment above](https://github.com/rust-random/getrandom/pull/759#pullrequestreview-3502673183).

PR #760, opened 2025-11-25
Avoid conditional module inclusion

tamird:
This requires slightly more logic in util_libc, but produces a more
typical setup for the crate.

Extracted from #759.


newpavlov:
Personally, I think the proposed approach is slightly more complex and less clear. But it gave me an idea which I would like to try a bit later.

newpavlov:
Closing in favor of #763.

PR #761, opened 2025-11-26
rdrand: mark rdrand safe

tamird:
In edition 2024 functions annotated with `target_feature(enable = "..")`
are unsafe to call from contexts not so annotated, and otherwise safe.

This was used in https://github.com/rust-lang/stdarch/commit/59864cd to
mark RDRAND safe, and this change is expected to be in 1.93.0 and is
already in nightly. Since uefi targets are only tested on nightly, we
only saw this lint on uefi, resulting in a misattribution of the
behavior and an incorrect comment.

Thus acknowledge that the intrinsics are safe and mark `rdrand` itself
safe (when called from an annotated context) and remove all the newly
unused unsafe blocks.

Link: https://doc.rust-lang.org/reference/attributes/codegen.html#r-attributes.codegen.target_feature.safety-restrictions.

### DIFF
--- a/src/backends/rdrand.rs
+++ b/src/backends/rdrand.rs
@@ -28,10 +28,19 @@ static RDRAND_GOOD: lazy::LazyBool = lazy::LazyBool::new();
 const RETRY_LIMIT: usize = 10;
 
 #[target_feature(enable = "rdrand")]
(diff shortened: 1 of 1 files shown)


tamird:
Oops, this is incorrect, it has to do with https://github.com/rust-lang/stdarch/pull/1908. I'll update this PR in a moment.

EDIT: done.

tamird:
@dhardy could you please take a look?

PR #763, opened 2025-12-03
Replace `util_libc` and `sanitizer` modules with `impl_utils` macro

newpavlov:
Based on #760


src/impl_utils.rs
```diff
@@ -0,0 +1,151 @@
...
+                }
+            }
+            Ok(())
+        }
+    };
+
+    (get_errno) => {
+        cfg_if! {
+            if #[cfg(any(target_os = "netbsd", target_os = "openbsd", target_os = "android", target_os = "cygwin"))] {
+                use libc::__errno as errno_location;
```
newpavlov, line 58: Technically, we could move these imports into backends, but I think it will result in a more convoluted and harder to review code.

src/impl_utils.rs
```diff
@@ -0,0 +1,151 @@
...
+            let errno: i32 = unsafe { get_errno() };
+
+            if errno > 0 {
+                let code = errno
+                    .checked_neg()
+                    .expect("Positive number can be always negated");
+                Error::from_neg_error_code(code)
+            } else {
+                Error::ERRNO_NOT_POSITIVE
+            }
```
tamird, line 17: not new in this PR but this code is weird, checking for positive and then using `.expect` can just be

`errno.checked_neg().map_or(Error::ERRNO_NOT_POSITIVE, Error::from_neg_error_code)`

src/impl_utils.rs
tamird: personal opinion: that would be better

tamird (commented):
:+1:

src/impl_utils.rs
newpavlov: You forget that `errno` could be theoretically equal to 0. This code is intentionally written in such way so it would be compiled into branchless code. See: https://rust.godbolt.org/z/f646Gczfh

tamird: > You forget that `errno` could be theoretically equal to 0. This code is intentionally written in such way so it would be compiled into branchless code. See: https://rust.godbolt.org/z/f646Gczfh

Ah, thanks. But then why not write

```rust
    if errno > 0 {
        Error::from_neg_error_code(-errno)
    } else {
        Error::ERRNO_NOT_POSITIVE
    }
```

newpavlov: To have a clear assertion that overflow on negation can not happen.

briansmith:
What is the purpose of this change?

newpavlov:
It's my take on #760 which removes reliance on the uncommon `#[path = "..."]` attribute. It's debatable whether the reliance on macros is better than the attribute. Personally, I am fine with either.

Though one change which I think we should certainly do is to move the `sanitizer` module out of the backends folder since it is not a "backend".

briansmith:
> Though one change which I think we should certainly do is to move the sanitizer module out of the backends folder since it is not a "backend".

It is an implementation detail of some backends so I think it is fine there. IMO not worth the churn. But, IDC much.

briansmith:
> It's my take on https://github.com/rust-random/getrandom/pull/760 which removes reliance on the uncommon #[path = "..."] attribute.

It is an ugly thing but I don't think it's worth introducing additional complexity. In other projects, I tend to use `mod X_Y; use X_Y as X` instead of `#[path = "X_Y.rs"] mod X` to get the same effect.

I don't think the macro is an improvement.

tamird:
> > It's my take on #760 which removes reliance on the uncommon #[path = "..."] attribute.
> 
> It is an ugly thing but I don't think it's worth introducing additional complexity. In other projects, I tend to use `mod X_Y; use X_Y as X` instead of `#[path = "X_Y.rs"] mod X` to get the same effect.
> 
> I don't think the macro is an improvement.

That pattern wouldn't work here because the module compiles only on certain targets. This is why I think https://github.com/rust-random/getrandom/pull/763#discussion_r2586006152 is an improvement; it would make that more clear.

briansmith:
> That pattern wouldn't work here because the module compiles only on certain targets. This is why I think [#763 (comment)](https://github.com/rust-random/getrandom/pull/763#discussion_r2586006152) is an improvement; it would make that more clear.

It does work just fine; just the `mod X_Y; use X_Y as X` has to be inside the conditional compilation logic.

newpavlov:
Closing in favor of #765.

PR #765, opened 2025-12-10
Refactor `util_libc`

newpavlov:
This PR refactors `util_libc`, `sanitizer`, and `lazy` modules. The modules are moved into the `utils` directory and "mounted" with `#[path = ".."]`. With this change all uses of `#[path = ".."]` point towards the `utils` directory.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,16 +3,39 @@
 version = 4
 
 [[package]]
+name = "async-trait"
+version = "0.1.89"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "autocfg"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
+
+[[package]]
 name = "bumpalo"
 version = "3.19.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5dd9dc738b7a8311c7ade152424974d8115f2cdad61e8dab8dac9f2362298510"
 
 [[package]]
-name = "cc"
-version = "1.2.50"
+name = "cast"
+version = "0.3.0"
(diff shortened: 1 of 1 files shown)


src/utils/sys_fill_exact.rs
```diff
@@ -0,0 +1,36 @@
...
+    mut buf: &mut [MaybeUninit<u8>],
+    sys_fill: impl Fn(&mut [MaybeUninit<u8>]) -> libc::ssize_t,
+) -> Result<(), Error> {
+    while !buf.is_empty() {
+        let res = sys_fill(buf);
+        match res {
+            res if res > 0 => {
+                let len = usize::try_from(res).map_err(|_| Error::UNEXPECTED)?;
+                let (l, r) = buf.split_at_mut_checked(len).ok_or(Error::UNEXPECTED)?;
+                unsafe { sanitizer::unpoison(l) };
```
newpavlov, line 24: Note that it applies `unpoison` to all users of `sys_fill_exact`. It's redundant in some cases (e.g. when we read from `/dev/urandom`), but AFAIK it has no negative consequences (outside of outlandish scenarios of misbehaving `libc`), while allowing to simplify the code a fair bit.

src/backends/getentropy.rs
```diff
@@ -12,15 +12,16 @@ use core::{ffi::c_void, mem::MaybeUninit};
 
 pub use crate::util::{inner_u32, inner_u64};
 
-#[path = "../util_libc.rs"]
-mod util_libc;
+#[path = "../utils/get_errno.rs"]
+mod utils;
```
newpavlov, line 16: We could move imports of `errno_location` for 4 targets which use this backends here, but I don't think it's worth the trouble since all 4 targets use different names.

src/utils/sys_fill_exact.rs
```diff
@@ -0,0 +1,36 @@
+use crate::Error;
+use core::mem::MaybeUninit;
+
+mod get_errno;
+mod sanitizer;
+
+pub(crate) use get_errno::get_errno;
+
+pub(crate) fn sys_fill_exact(
```
dhardy, line 14: This fn lost its doc which should still be valid and has some (internal) use.

src/backends/linux_raw.rs
dhardy, line 167: Why doesn't this use `sys_fill_exact`?

newpavlov: Because `sys_fill_exact` uses errno, while `asm!` returns error code directly without modifying errno.

PR #769, opened 2025-12-27
Prepare v0.4.0-rc.0

dhardy:
We would like a pre-release for testing purposes. (Due to the `rand_core` pre-release dependency we cannot yet release v0.4, though I believe otherwise the repo is ready for v0.4.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,7 +32,7 @@ checksum = "3a3076410a55c90011c298b04d0cfa770b00fa04e1e3c97d3f6c9de105a03844"
 
 [[package]]
 name = "getrandom"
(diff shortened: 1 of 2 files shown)


dhardy:
Failures are "unnecessary use of `unsafe`", which shouldn't block a pre-release (though CI pass would be preferred) and should be handled in a different PR.

tarcieri:
Opened #770 to fix `unused_unsafe`

dhardy:
Published: https://crates.io/crates/getrandom/0.4.0-rc.0

(The release automation needed re-configuring. So much for automation saving work!)

PR #771, opened 2025-12-27
Remove all uses of `getrandom_backend = "wasm_js"`

newpavlov:
Enabling the `wasm_js` crate feature is sufficient.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ std = []
 # This flag enables the wasm_js backend and uses it by default on wasm32 where
 # the target_os is unknown. The getrandom_backend cfg may override this.
 #
-# WARNING: We recommend strongly against enabling this feature in libraries (except for tests)
+# WARNING: We strongly recommend against enabling this feature in libraries (except for tests)
 # since it is known to break non-Web WASM builds and further since the usage of `wasm-bindgen`
 # causes significant bloat to `Cargo.lock` (on all targets).
 #
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ We do not include support for this target in the default configuration because o
 To enable `getrandom`'s functionality on `wasm32-unknown-unknown` using
 [`Crypto.getRandomValues`] via [`wasm-bindgen`], enable the `wasm_js` crate feature.
 
-WARNING: We recommend strongly against enabling this feature in libraries (except for tests)
+WARNING: We strongly recommend against enabling this feature in libraries (except for tests)
 since it is known to break non-Web WASM builds and further since the usage of `wasm-bindgen`
 causes significant bloat to `Cargo.lock` (on all targets).
 


.github/workflows/tests.yml
```diff
@@ -275,14 +275,13 @@ jobs:
           - {
               description: Web,
               version: stable,
-              flags: '-Dwarnings --cfg getrandom_backend="wasm_js"',
```
dhardy, line 278: We still want `-Dwarnings` (unless substituted below)

Cargo.toml
```diff
@@ -20,10 +20,16 @@ features = ["std", "sys_rng"]
...
 
 # Optional backend: wasm_js
+#
 # This flag enables the wasm_js backend and uses it by default on wasm32 where
 # the target_os is unknown. The getrandom_backend cfg may override this.
-# WARNING: It is highly recommended to enable this feature only for binary crates and tests,
-# i.e. avoid unconditionally enabling it in library crates.
+#
+# WARNING: This feature SHOULD be enabled ONLY for binary crates and tests. In other words,
+# avoid enabling it in library crates whether unconditionally or gated on a crate feature.
```
dhardy: >  or gated on a crate feature

I see no hard reason not to have a "forwarding" feature flag (e.g. `wasm_js = ["getrandom/wasm_js"]`).

I know you don't like this pattern, but I don't think directing users like this will work particularly well.

I suggest we keep our advice direct and minimally opinionated:

> We recommend against enabling this feature in libraries (except for tests) since it is known to break some non-Web WASM builds and further since the usage of `wasm-bindgen` causes significant bloat to `Cargo.lock` (on all targets).

josephlr: I think having a default guidance against forwarding crate features is a reasonable. It's mostly that you run into semver and stability issues when trying to forward the features of other crates. It's possible to do, but really only works if:
  - you pay extra close attention to the issues
  - you work closely between the forwarding crate and `getrandom` (like `rand` does).

Cargo.toml
```diff
@@ -20,10 +20,16 @@ features = ["std", "sys_rng"]
 std = []
 
 # Optional backend: wasm_js
+#
 # This flag enables the wasm_js backend and uses it by default on wasm32 where
 # the target_os is unknown. The getrandom_backend cfg may override this.
-# WARNING: It is highly recommended to enable this feature only for binary crates and tests,
-# i.e. avoid unconditionally enabling it in library crates.
+#
+# WARNING: We recommend strongly against enabling this feature in libraries (except for tests)
```
dhardy:
```suggestion
# WARNING: We strongly recommend against enabling this feature in libraries (except for tests)
```

dhardy (approved):
Then I approve.

newpavlov:
>It might be worth mentioning (somewhere) that we used to allow --cfg getrandom_backend="wasm_js"

This change does not "forbid" `--cfg getrandom_backend="wasm_js"` in any way. It's simply gets ignored and the only visible consequence is lack of the compilation error which warns users to enable the `wasm_js` feature (instead users will get a different compilation error).

So, arguably, this change is not a breaking one and arguably we should've done it as part of the v0.3.4 release.

PR #774, opened 2025-12-31
Release 0.2.17

josephlr:
Release date set for Jan 2nd so we can do this in the new year if folks are taking time off.

I'm not sure how this will interact with Trusted Publishing, but happy to try things out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.2.17] - 2026-01-02
+## [0.2.17] - 2026-01-12
 ### Fixed
 - Use `doc_cfg` instead of `doc_auto_cfg` (partial backport of [#732]) [#768]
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -461,6 +468,7 @@ Publish initial implementation.
 ## [0.0.0] - 2019-01-19
 Publish an empty template library.
 
+[0.2.17]: https://github.com/rust-random/getrandom/compare/v0.2.15...v0.2.17
```
```suggestion
[0.2.17]: https://github.com/rust-random/getrandom/compare/v0.2.16...v0.2.17
```

newpavlov (approved):

dhardy:
> I'm not sure how this will interact with Trusted Publishing, but happy to try things out.

You should just have to push the tag then approve running the release action on GH. In theory. Last time I had to disable a requirement for CI tests to pass which didn't seem to work.

newpavlov:
I think it's worth to also fix #775 as part of this release. We probably can ignore #777 since v0.2 supports only WASIp1 and we did not bother with backporting WASIp2 support for it.

newpavlov:
>I'm not sure how this will interact with Trusted Publishing, but happy to try things out.

We do not have the publish CI job in the v0.2 branch, so I think it simply will not work. We probably have to publish it the old fashioned way.

PR #790, opened 2026-01-26
Bump `wasip3` to v0.4

newpavlov:


.cargo/config.toml
```diff
@@ -5,7 +5,12 @@ runner = 'wasm-bindgen-test-runner'
...
 runner = 'wasmtime'
+[target.wasm32-wasip3]
+runner = 'wasmtime'
 
 # Just run on node by default (that's where emscripten is tested)
 [target.'cfg(target_os = "emscripten")']
 runner = 'node'
+
+[resolver]
+incompatible-rust-versions = "allow"
```
newpavlov, line 16: `wasip3` v0.4 requires Rust 1.87. Technically, we should bump `rust-version` for `getrandom` as well, but I guess it's fine to skip it in this case.

Unfortunately, this setting may complicate MSRV testing since we could pull dependencies into Cargo.lock which require a higher Rust version. I think the right way to handle this would be to execute `cargo update` in each MSRV job.

dhardy: Is it not sufficient to use `cargo update --ignore-rust-version` only in the relevant tests?

newpavlov: Yes, it could work as well. But I think we should commit latest dependency versions into Cargo.lock and test with them by default, not MSRV-compatible ones. Plus, we have more non-MSRV jobs.

newpavlov: See https://github.com/rust-random/getrandom/blob/b229df2fe441b42af616116249209cbc29d77cdb/.github/workflows/tests.yml#L373-L376

Cargo.toml
```diff
@@ -81,7 +81,7 @@ wasip2 = { version = "1", default-features = false }
 
 # wasi_p2_3
 [target.'cfg(all(target_arch = "wasm32", target_os = "wasi", target_env = "p3"))'.dependencies]
-wasip3 = { version = "0.3", default-features = false }
+wasip3 = "0.4"
```
briansmith, line 84: I wonder if the MSRV issue could be worked around by depending on `wasip3 = ">= 0.3, <= 0.4"` (or whatever) so that 0.3 could be used to satisfy the MSRV constraint?

.github/workflows/tests.yml
```diff
@@ -370,6 +370,10 @@ jobs:
           wget -O - $URL | tar -xJ --strip-components=1 -C ~/.cargo/bin
           wasmtime --version
       - uses: Swatinem/rust-cache@v2
+      - name: Generate MSRV-compatible Cargo.lock
+        env:
+          CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS: "fallback"
+        run: cargo update
```
briansmith, line 376: IMO it would be better to restrict this to `-p wasip2 -p wasip3` or whatever, instead of updating all the dependencies.

Issue #796, opened 2026-01-27
Tracking issue for migration to `extern_item_impls`


newpavlov:
[`extern_item_impls`](https://github.com/rust-lang/rust/issues/125418) provides language support for ergonomically adding `getrandom` support for targets which do not have a default impl. It also allows to override default `getrandom` implementations if necessary.

On stabilization of the feature we probably should fully migrate to it by replacing the opt-in backends and the `wasm_js` crate feature with a number of plug-in implementation crates.

Open questions:
- [ ] Should we use [externally implementable traits](https://github.com/rust-lang/rfcs/pull/3645) (EITs) instead of externally implementable functions (EIFs)? IIUC EIFs are likely to get stabilized earlier than EITs. Additionally, it looks like the current design of EITs is not compatible with the `TryRng` trait.
- [ ] Security considerations for silently overwriting the default `getrandom` impls. Ideally, it would be nice if it was impossible to silently override the default impl by just adding a dependency, but the current design of EIFs does not provide a way to protect against it. ([previous discussion](https://github.com/rust-random/getrandom/pull/786))
- [ ] Should we keep plug-in crates in this repository or move it into a new one?

bushrat011899:
> Should we use [externally implementable traits](https://github.com/rust-lang/rfcs/pull/3645) (EITs) instead of externally implementable functions (EIFs)? IIUC EIFs are likely to get stabilised earlier than EITs. Additionally, it looks like the current design of EITs is not compatible with the `TryRng` trait.

I think a trait is a better match conceptually, it would be nice if you could, for example, use an associated type for `getrandom::Error`.  But the current nightly feature [isn't really traits](https://github.com/rust-lang/rfcs/pull/3645#issuecomment-2131304319). I think for the current design of `getrandom`, EIFs are close enough to provide a meaningful ergonomic improvement. Since the whole crate is currently built around providing a single `fn(&mut [MaybeUninit<u8>]) -> Result<(), Error>`, I think the migration to EIFs would be quite straightforward.

> Security considerations for silently overwriting the default `getrandom` impls. Ideally, it would be nice if it was impossible to silently override the default impl by just adding a dependency [...]

I see 3 cases here that need to be designed around:

1. Relying on a default backend from `getrandom`
2. Relying on a 3rd party backend from a dependency
3. Providing your own implementation

[...]

PR #798, opened 2026-02-01
Prepare v0.4.0 release

dhardy:
I believe this is ready for release (see #722)?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.4.0] - 2026-02-02
 
 ### Added
 - `RawOsError` type alias [#739]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -104,7 +104,7 @@ dependencies = [
 
 [[package]]
 name = "getrandom"
-version = "0.4.0-rc.1"
+version = "0.4.0"
 dependencies = [
  "cfg-if",
  "js-sys",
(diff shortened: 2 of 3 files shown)


newpavlov:
#794 should be merged first.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.4.0] - 2026-02-01
```
newpavlov: Update the release date.

newpavlov:
Note that in the custom backend we still use `__getrandom_v03_custom`. I think it's worth to keep it as-is since the symbol is unchanged and it will be more convenient for cases when both v0.3 and v0.4 are present in a project's dependency tree. In v0.5 (or v1.0) we are likely to remove the custom backend completely in favor of #796.

dhardy:
Published. @newpavlov will you make the GH release?

newpavlov:
Since we switched to Trusted Publishing you just need to push new version tag and with the current configuration the workflow needs your approval, so it's probably better for you to push the tag.

UPD: Ah, I see. The environment settings have changed and the list of reviewers now includes me and Joe. Either way, you already pushed the tag.

newpavlov:
@dhardy 
Maybe we should change the environment setting to prevent self-reviews since we now have several eligible reviewers? Though, since the setting can be changed, it's still basically a honor system. Additionally,  it would complicate setting correct release date in the changelog if reviewers are not responsive enough...

dhardy:
I was referring to [GitHub releases](https://github.com/rust-random/getrandom/releases) which aren't automated... or used at all in this case. Never mind.

You can change the requirements to prevent self-review if you like, but yes it is essentially an honour system. The more interesting part would be if we can require a release version to be published to either master or a backports branch, but I'm not sure exactly how the checks work. We could try that.

dhardy:
(I assumed that since this PR was approved that it was fine to also push the tag.)

newpavlov:
>I was referring to [GitHub releases](https://github.com/rust-random/getrandom/releases) which aren't automated... or used at all in this case.

As I wrote in one of issues, personally I don't think we should bother with GH releases, regardless whether they are created manually or using a workflow.

dhardy:
The only use I see for them is that people can choose to "watch releases" on GitHub which I do for a few things, though I don't find it massively useful.

tarcieri:
It's nice to have a page with the release notes for just one particular release, but I can always link to the changelog

PR #801, opened 2026-02-02
Fix documentation on docs.rs

a1phyr:
Cf logs here: https://docs.rs/crate/getrandom/latest/builds/2867462


newpavlov:
Well, this is unfortunate... I published a test crate to see whether the fix works.

PR #804, opened 2026-02-03
Unify lazy atomics in entropy backends

tamird:
Replace ad-hoc atomic lazy caches with shared lazy helpers in the Linux/Android fallback, NetBSD, RDRAND, and RNDR backends. Add `LazyPtr` alongside `LazyUsize` and `LazyBool` so pointer and boolean caches use the same initialization contract.

This reduces duplicated cache logic and keeps backend probing/fallback semantics aligned while preserving the existing retry-until-cached behavior.

### DIFF
--- a/src/utils/lazy_ptr.rs
+++ b/src/utils/lazy_ptr.rs
@@ -24,7 +24,7 @@ impl<T> LazyPtr<T> {
         Self(AtomicPtr::new(ptr::null_mut()))
     }
 
-    /// Call the `init` closure and return the result after caching it in the case of succeess.
+    /// Call the `init` closure and return the result after caching it in the case of success.
     #[cold]
     fn cold_init<E>(&self, init: impl FnOnce() -> Result<NonNull<T>, E>) -> Result<NonNull<T>, E> {
         let val = init()?;
@@ -33,7 +33,7 @@ impl<T> LazyPtr<T> {
     }
 
     /// Retrieve the cached value if it was already initialized or call the potentially fallible
-    /// `init` closure and return the result after caching it in the case of succeess.
+    /// `init` closure and return the result after caching it in the case of success.
     #[inline]
     pub fn try_unsync_init<E>(
         &self,
@@ -49,6 +49,7 @@ impl<T> LazyPtr<T> {
     /// Retrieve the cached value if it was already initialized or call the `init` closure
     /// and return the result after caching it.
     #[inline]
+    #[allow(dead_code, reason = "Some modules use only `try_unsync_init`")]
     pub fn unsync_init(&self, init: impl FnOnce() -> NonNull<T>) -> NonNull<T> {
         let Ok(p): Result<_, Infallible> = self.try_unsync_init(|| Ok(init()));
         p


src/utils/lazy.rs
```diff
@@ -1,63 +1,107 @@
...
+    };
 }
 
-// Identical to LazyUsize except with bool instead of usize.
+lazy_atomic!(LazyUsize, AtomicUsize, usize, usize::MAX);
+lazy_atomic!(LazyPtr<T>, AtomicPtr<T>, *mut T, ptr::dangling_mut());
+
+/// Lazily initializes a cached bool; reuses `LazyUsize` to avoid sentinel
+/// issues with `AtomicBool`.
+#[allow(dead_code)]
```
newpavlov: Use `#![allow(dead_code)]` for the whole module.

src/utils/lazy.rs
```diff
@@ -1,63 +1,107 @@
...
+                }
+            }
         }
-    }
+    };
 }
 
-// Identical to LazyUsize except with bool instead of usize.
+lazy_atomic!(LazyUsize, AtomicUsize, usize, usize::MAX);
+lazy_atomic!(LazyPtr<T>, AtomicPtr<T>, *mut T, ptr::dangling_mut());
```
newpavlov: Why use dangligng pointer instead of null?

src/utils/lazy.rs
```diff
@@ -1,63 +1,107 @@
...
-//          let v = b();
-//          c();
-//          v
-//      }
-// the effects of c() or writes to shared memory will not necessarily be
-// observed and additional synchronization methods may be needed.
-struct LazyUsize(AtomicUsize);
+macro_rules! lazy_atomic {
+    ($name:ident $(<$($gen:ident),+>)?, $atomic:ty, $value:ty, $uninit:expr) => {
+        // This structure represents a lazily initialized static usize value. Useful
```
newpavlov: This doc is not correct for `LazyPtr`. Probably better to just move the description to the module-level docs.

tamird: Done, thanks.

src/utils/lazy.rs
tamird: This is removed now.

src/utils/lazy.rs
tamird: Done.

src/backends/linux_android_with_fallback.rs
```diff
@@ -15,20 +14,14 @@ type GetRandomFn = unsafe extern "C" fn(*mut c_void, libc::size_t, libc::c_uint)
 
 /// Sentinel value which indicates that `libc::getrandom` either not available,
 /// or not supported by kernel.
-const NOT_AVAILABLE: NonNull<c_void> = unsafe { NonNull::new_unchecked(usize::MAX as *mut c_void) };
-
-static GETRANDOM_FN: AtomicPtr<c_void> = AtomicPtr::new(core::ptr::null_mut());
+const NOT_AVAILABLE: NonNull<c_void> = NonNull::dangling();
```
newpavlov: I don't think this is correct. Quoting the method docs:
>Note that the address of the returned pointer may potentially be that of a valid pointer, which means this must not be used as a “not yet initialized” sentinel value.

Meanwhile with the `usize::MAX` variant we can argue that it fundamentally can not be a valid function pointer.

src/backends/linux_android_with_fallback.rs
```diff
@@ -37,10 +30,9 @@ fn init() -> NonNull<c_void> {
 
     let res_ptr = match NonNull::new(raw_ptr) {
         Some(fptr) => {
-            let getrandom_fn = unsafe { transmute::<NonNull<c_void>, GetRandomFn>(fptr) };
-            let dangling_ptr = NonNull::dangling().as_ptr();
+            let getrandom_fn = unsafe { transmute::<*mut c_void, GetRandomFn>(fptr.as_ptr()) };
```
newpavlov, line 33: Why are you casting from `NonNull<c_void>` to `*mut c_void`?

src/backends/linux_android_with_fallback.rs
```diff
@@ -77,23 +68,17 @@ fn use_file_fallback(dest: &mut [MaybeUninit<u8>]) -> Result<(), Error> {
...
+
+    static GETRANDOM_FN: lazy::LazyNonNull<c_void> = lazy::LazyNonNull::new();
+    let fptr = GETRANDOM_FN.unsync_init(init);
 
     if fptr == NOT_AVAILABLE {
         use_file_fallback(dest)
     } else {
         // note: `transmute` is currently the only way to convert a pointer into a function reference
-        let getrandom_fn = unsafe { transmute::<NonNull<c_void>, GetRandomFn>(fptr) };
+        let getrandom_fn = unsafe { transmute::<*mut c_void, GetRandomFn>(fptr.as_ptr()) };
```
newpavlov, line 81: Same.

src/backends/netbsd.rs
```diff
@@ -42,36 +41,28 @@ unsafe extern "C" fn polyfill_using_kern_arand(
...
-        fptr = init();
-    }
-    let fptr = unsafe { mem::transmute::<*mut c_void, GetRandomFn>(fptr) };
+    #[path = "../utils/lazy.rs"]
+    mod lazy;
+
+    static GETRANDOM_FN: lazy::LazyNonNull<c_void> = lazy::LazyNonNull::new();
+
+    let fptr = GETRANDOM_FN.unsync_init(init);
+    let fptr = unsafe { mem::transmute::<*mut c_void, GetRandomFn>(fptr.as_ptr()) };
```
newpavlov, line 66: Same.

src/utils/lazy.rs
```diff
@@ -1,64 +1,91 @@
...
+        match NonNull::new(self.0.load(Ordering::Relaxed)) {
+            Some(val) => Ok(val),
+            None => self.try_do_init(init),
         }
     }
 }
 
-// Identical to LazyUsize except with bool instead of usize.
-pub(crate) struct LazyBool(LazyUsize);
+pub(crate) struct LazyBool(AtomicUsize);
```
newpavlov: We can use `AtomicU8` here.

newpavlov (commented):
I think it's worth to split the `lazy` module to `lazy_bool` and `lazy_ptr` modules. It would allow us to remove the `#![allow(dead_code)]` part.

I also think it's fine to name `LazyNonNull` simply as `LazyPtr` and write proper docs for the structs since you no longer use macro to generate them.

src/utils/lazy.rs
tamird: Done.

src/backends/netbsd.rs
tamird: Same.

src/backends/linux_android_with_fallback.rs
tamird: Same.

src/backends/linux_android_with_fallback.rs
tamird: That's not what is happening here. We are using `as_ptr()` to get the `*mut void` and casting it to `GetRandomFn`. This is better than casting the `NonNull` directly because it doesn't rely on `NonNull` being `repr(transparent)`.

src/backends/linux_android_with_fallback.rs
tamird: Good point. Done.

src/backends/linux_android_with_fallback.rs
newpavlov: By "casting" I meant the `fptr.as_ptr()` part.

>This is better than casting the NonNull directly because it doesn't rely on NonNull being repr(transparent).

I don't think it's "better". `repr(transparent)` is a promise made by `std` which for `NonNull` was introduced in Rust 1.29. We can rely on this property no less than  `*mut c_void` being castable to a function pointer. For example, see https://github.com/rust-lang/rust/issues/49220.

newpavlov: Hm, looking into it a bit more carefully it looks like it's not as clear cut as I would've liked. I asked about it [here](https://github.com/rust-lang/rust/issues/145989#issuecomment-3873687655).

tamird: Seems better to sidestep this potential problem by just casting the raw pointer. Yes?

newpavlov: Yeah, it looks like this aspect is surprisingly (in a bad way) under-specified...

tamird:
@newpavlov I believe I've addressed all your comments. Does this LGTY?

newpavlov:
Yes, I think it more or less ready for merge. I want to give it one last look and push some doc-related changes in the following days.

newpavlov:
@tamird
I pushed minor changes to the lazy modules. Please take a look. If you don't have anything to add on top or tweak something in my changes, I think we are good to merge.

src/utils/lazy_ptr.rs
```diff
@@ -1,65 +1,57 @@
...
+            None => self.cold_init(init),
         }
     }
+
+    /// Retrieve the cached value if it was already initialized or call the `init` closure
+    /// and return the result after caching it.
+    #[inline]
+    #[allow(dead_code, reason = "Some modules use only `try_unsync_init`")]
+    pub fn unsync_init(&self, init: impl FnOnce() -> NonNull<T>) -> NonNull<T> {
+        let Ok(p): Result<_, Infallible> = self.try_unsync_init(|| Ok(init()));
```
tamird, line 54: Nice! I forgot about this irrefutable pattern trick.

PR #830, opened 2026-03-12
Use manual bindings for WASIp2/3

newpavlov:
Closes #827

### DIFF
--- a/src/backends/wasi_p2_3.rs
+++ b/src/backends/wasi_p2_3.rs
@@ -2,20 +2,31 @@
 use crate::Error;
 use core::{mem::MaybeUninit, ptr::copy_nonoverlapping};
 
-#[cfg(target_env = "p2")]
-use wasip2 as wasi;
-
-// Workaround to silence `unexpected_cfgs` warning
-// on Rust version between 1.85 and 1.91
-#[cfg(not(target_env = "p2"))]
-#[cfg(target_env = "p3")]
-use wasip3 as wasi;
-
 #[cfg(not(target_env = "p2"))]
 #[cfg(not(target_env = "p3"))]
 compile_error!("Unknown version of WASI (only previews 1, 2 and 3 are supported)");
 
-use wasi::random::random::get_random_u64;
+// Use manual bindings instead of `wasip2/3` crates to prevent Cargo.lock bloat.
+//
+// Technically, such manual bindings may not work since we also have to provide
+// type information which is expected by `wasm-component-ld`, but we assume that
+// in practice the relevant type information will be provided by linking to `std`
+// and/or `wasip2/3` crate.
+//
+// See https://github.com/rust-random/getrandom/pull/828 for more information.
+#[cfg_attr(
+    target_env = "p2",
+    link(wasm_import_module = "wasi:random/random@0.2.0")
+)]
+// TODO(MSRV-1.92): Use `target_env = "p3"`
+#[cfg_attr(
+    not(target_env = "p2"),
+    link(wasm_import_module = "wasi:random/random@0.3.0")
+)]
+unsafe extern "C" {
+    #[link_name = "get-random-u64"]
+    safe fn get_random_u64() -> u64;
+}
 
 #[inline]
 pub fn inner_u32() -> Result<u32, Error> {
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -370,15 +370,23 @@ jobs:
         # Firefox doesn't support module service workers and therefor can't import scripts
         run: wasm-pack test --headless --chrome -- --target ${{ matrix.rust.target }} ${{ matrix.rust.args }}
 
-  wasi_p1:
-    name: WASIp1
+  wasi:
+    name: WASI
     runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        target: [
+          wasm32-wasip1,
+          wasm32-wasip2,
+          # TODO: enable after pre-built std will be provided by Rust 
+          # wasm32-wasip3,
+        ]
     steps:
       - uses: actions/checkout@v6
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: 1.85
-          targets: wasm32-wasip1
+          targets: ${{ matrix.target }}
       - name: Install Wasmtime
         run: |
           VERSION=v40.0.0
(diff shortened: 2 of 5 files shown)


src/backends/wasi_p2_3.rs
```diff
@@ -2,20 +2,25 @@
...
+// in practice the relevant type information will be provided by linking to `std`
+// and/or `wasip2/3` crate.
+//
+// See https://github.com/rust-random/getrandom/pull/828 for more information.
+#[cfg_attr(target_env = "p2", link(wasm_import_module = "wasi:random/random@0.2.0"))]
+// TODO: drop the RC part on full WASIp3 release
+#[cfg_attr(target_env = "p3", link(wasm_import_module = "wasi:random/random@0.3.0-rc-2026-01-06"))]
+unsafe extern "C" {
+    #[link_name = "get-random-u64"]
+    safe fn get_random_u64() -> u64;
```
newpavlov, line 28: @alexcrichton 
Is it fine to use `u64` here instead of `i64` used by the auto-generated bindings? Otherwise it looks somewhat weird to get `i64` from a function named `get_random_u64`.

newpavlov: Also, would `random@0.3.0-rc-2026-01-06` continue to work after WASI v0.3.0 is released in the same way as was explained by you [here](https://github.com/rust-random/getrandom/pull/828#issuecomment-4035043874)? (obviously assuming that `get-random-u64` is not changed)

alexcrichton: > Is it fine to use u64 here instead of i64 used by the auto-generated bindings?

yep!

> Also, would random@0.3.0-rc-2026-01-06 continue to work after WASI v0.3.0 is released in the same way as was explained by you here?

Alas, no, for all future rcs it'll require an update here to change the version string. Once WASIp3 is officially released as 0.3.0 this can be updated to 0.3.0 and that'll be it.

briansmith: Under some circumstances it is UB to declare an extern function with different types than another `extern` declaration for the same function. I think we ran into this with CFI related stuff on Windows. I think the more conservative approach of doing what the generator does is better.

src/backends/wasi_p2_3.rs
```diff
@@ -2,20 +2,26 @@
...
+// Technically, such manual bindings may not work since we also have to provide
+// type information which is expected by `wasm-component-ld`, but we assume that
+// in practice the relevant type information will be provided by linking to `std`
+// and/or `wasip2/3` crate.
+//
+// See https://github.com/rust-random/getrandom/pull/828 for more information.
+#[cfg_attr(target_env = "p2", link(wasm_import_module = "wasi:random/random@0.2.0"))]
+// TODO: drop the RC part on full WASIp3 release
+// TODO(MSRV-1.92): Use `target_env = "p3"`
+#[cfg_attr(not(target_env = "p2"), link(wasm_import_module = "wasi:random/random@0.3.0-rc-2026-01-06"))]
```
briansmith: For WASI p2, I think this makes sense because WASI 0.2 it stable (right?). But, I am concerned that we're omitting the additional info that the WASI p2 linker needs. IMO, it would be better to just take the output of wit-bindgen as-is for p2.

For WASI p3, IMO it would be better to depend on the `wasip3` crate with a `">= 0.3"` or whatever, so that we don't have to release `getrandom` over and over just to deal with the evolution of WASI p3. Users shouldn't need to update `getrandom` when they update the WASI p3 edition they use, unless/until WASI p3 actually changes the type signature of its random API.

src/backends/wasi_p2_3.rs
newpavlov: Yes, this is why I asked the question above. IIUC on the fundamental level WASM does not have a separate notion of "unsigned" integers, so for WASM targets `u64` is guaranteed to be indistinguishable from `i64` on the ABI level.

src/backends/wasi_p2_3.rs
newpavlov: I _really_ do not want to include the obscure type information static and `#[used]` hacks in `getrandom`. As I commented [here](https://github.com/rust-random/getrandom/pull/828#pullrequestreview-3919955685), I would prefer to keep the `wasip*` dependency over that.

I agree with your concern about p3. Unfortunately, IIUC depending on `wasip3` would keep the bloated Cargo.lock, so it does not make much sense to migrate only p2. I guess we could postpone migration to manual bindings until p3 is properly stabilized.

src/backends/wasi_p2_3.rs
briansmith: > Yes, this is why I asked the question above. IIUC on the fundamental level WASM does not have a separate notion of "unsigned" integers, so u64 is indistinguishable from i64 on the ABI level for WASM targets.

I understand that it is something like that. But, still, I don't think the risk is worth the benefit.

src/backends/wasi_p2_3.rs
briansmith: The main motivation for wasi p2 is to keep the MSRV stable and lower than the wasip2 crate's MSRV, to reduce build bloat and build times for WASI p2, at the cost of having to maintain the binding. As we expect WASI p2 to never need to be touched again, I think this is the right trade-off for p2.

But, the fact that WASI p3 is a moving target totally shifts the balance the other way for p3. It is too much of a burden for users to have to upgrade getrandom every time they want to move forward tracking WASI p3. More importantly, for people NOT doing WASI p3, they shouldn't have to deal with new getrandom releases that would be necessary to track p3.

The Cargo.lock issue should be dealt with by improving Cargo, IMO.

newpavlov: But the MSRV issue is caused by `wasip3`, not `wasip2`, no? So migrating just p2 would not help here.

>The Cargo.lock issue should be dealt with by improving Cargo, IMO.

Ideally, yes. But with the current Cargo it's still a concern and it's unlikely to change in the near future.

briansmith: > I _really_ do not want to include the obscure type information static and `#[used]` hacks in `getrandom`. As I commented [here](https://github.com/rust-random/getrandom/pull/828#pullrequestreview-3919955685), I would prefer to keep the `wasip*` dependency over that.

How can we be confident that it is OK to omit it? We shouldn't let an aesthetic concern override correctness. Based on what Alex said, the additional info seems to be necessary. OTOH, why does this PR pass CI? I guess because something else used in the test harness and/or stdlib is bringing in the extra type info somehow?

briansmith: > But the MSRV issue is caused by `wasip3`, not `wasip2`, no? So migrating just p2 would not help here.

Oh, yeah, right.

IMO, there's not a big gain to be had by using the manual bindings then. Seems like the cure is worse than the problem it is trying to solve.

alexcrichton: W.r.t. wasip3, I would personally say that no one should expect this crate to work reliably until some point after WASIp3 is officially released. There's simply no good answer for you as maintainers of this crate other than publishing lots of versions to keep up to date, and I think it's completely reasonable to not want to do that. Put another way, I would personally have an expectation that y'all would be receptive to PRs to update the wasip3 version string as necessary, but you would not make releases exclusively for updating the wasip3 version string. That way users could use `[patch]` and such to get something working, but there's no need to place the burden of wasip3's instability on y'all.

> why does this PR pass CI?

Right yeah it's other stuff pulling in type information. For example the standard library depends on the `wasip2` crate which pulls in type information. The standard library also depends on wasi-libc which itself pulls in type information. This'll be true for the `wasip3` target as well in the future where the standard library will largely pull in type information.

You can see an example [here](https://github.com/rust-random/getrandom/pull/828#issuecomment-4033162473) of something that doesn't compile with this PR, and how it requires being relatively niche.

newpavlov: >OTOH, why does this PR pass CI? I guess because something else used in the test harness and/or stdlib is bringing in the extra type info somehow?

The test harness links to `std` which provides the necessary type information. Arguably, it's highly unlikely to have in practice a WASI project which does not link to either `std` or `wasip*` crate and you have to go pretty far out of your way to trigger the potential build failure as demonstrated [here](https://github.com/rust-random/getrandom/pull/828#issuecomment-4033162473).

IMO the linker should just include type information for the base WASI syscalls as part of the target spec.

newpavlov:
@alexcrichton 
Do you have an approximate timeline for a full (i.e. non-RC) wasip3 release?

alexcrichton:
I'd say 3-6 months-ish would be my best guess

newpavlov:
@dhardy @josephlr 
Non-RC version of WASIp3 was recently released, so I believe this PR is ready for merge.

src/backends/wasi_p2_3.rs
```diff
@@ -2,20 +2,32 @@
...
+// in practice the relevant type information will be provided by linking to `std`
+// and/or `wasip2/3` crate.
+//
+// See https://github.com/rust-random/getrandom/pull/828 for more information.
+#[cfg_attr(
+    target_env = "p2",
+    link(wasm_import_module = "wasi:random/random@0.2.0")
+)]
+// TODO: drop the RC part on full WASIp3 release
+// TODO(MSRV-1.92): Use `target_env = "p3"`
```
flisky, line 21: dead comments?

dhardy (approved):
I'm far from an expert in WASI, but from reading the discussions in #828/... I'm happy to stamp this as "approved".

IIUC from reading the discussion, link-time failure with contrived no-std examples is possible, but this will rarely be observed in practice; further, incorrect linking (due to the missing type information) should not happen.

@alexcrichton any further comments?

newpavlov:
IIUC if we want to be absolutely sure that the linking problem can not happen we could use `extern crate std;` for this backend, though I would prefer not to.

alexcrichton:
No concerns/comments from me, I think this is reasonable to land 👍

PR #861, opened 2026-08-11
Add Apache NuttX support

kay-lambdadelta:
This PR adds the ability to use getrandom on NuttX.

Supporting this target is fairly uninteresting as it has support for a linux-style [getrandom](https://github.com/apache/nuttx/blob/2567b729cf83f46f3ca1807fa09323dd49908e15/include/sys/random.h#L77) which is supported [via `libc`](https://github.com/rust-lang/libc/commit/c2e03d56aef34249bfb5010a366c26b0173f04e4)

This does however, require a bump for the version of `libc` this crate uses (to [0.2.176 for errno support](https://github.com/rust-lang/libc/commit/a831211d3fd4a7d914c38ac64f4c8547b0888dc0))


newpavlov (approved):

newpavlov:
Would it be possible to test the target in out CI using build-only job?

kay-lambdadelta:
> Would it be possible to test the target in out CI using build-only job?

A build-only test shouldn't require anything more than a nightly compiler. I would suggest pinning to a nightly version like `nightly-2026-08-09`, as rust NuttX support is pretty alpha and it breaks sometimes. Right now the riscv and the arm variants of the target should work.

